PR #24, opened 2019-03-18
Add travis integration

isears-cms:
Just adding the default java build for now.


isears-cms:
I think we're good. Especially seeing as a few other cms-gov projects are already on travis-org.

PR #26, opened 2019-03-18
DPC-162: Add patient and provider details to attribution engine

nickrobison-usds:
The PR modifies the roster engine in order to add necessary details for patients and providers.

Per the discussion in DPC-161, we now require additional fields to be present in the FHIR resources. DPC-171 and DPC-172 have been created to track those issues.

The major changes is that we now require submitting actual FHIR resources to the Roster resource. This is partially implemented in that it is possible to query for an attribution relationship using only the provider NPI and beneficiary ID, but that will be resolved in DPC-90.

There are a number of things still left to do, but this provides the foundation for DPC-21, DPC-22 and DPC-90.


dpc-attribution/src/main/java/gov/cms/dpc/attribution/models/PatientEntity.java
```diff
@@ -0,0 +1,114 @@
...
+        this.dob = dob;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        PatientEntity that = (PatientEntity) o;
+        return patientID.equals(that.patientID) &&
+                beneficiaryID.equals(that.beneficiaryID) &&
```
isears-cms: Thinking of a weird case where the same patient is registered under two different beneficiaries: would it be helpful for this equivalence function to identify those two patients as the same?

dpc-attribution/src/main/java/gov/cms/dpc/attribution/models/PatientEntity.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public LocalDate getDob() {
+        return dob;
+    }
+
+    public void setDob(LocalDate dob) {
+        this.dob = dob;
+    }
+
+    @Override
```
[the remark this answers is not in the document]
isears-cms: I don't necessarily agree with the "cognitive complexity" here, but would it be helpful to do unit tests for some of these equivalency or fromFHIR functions?

dpc-attribution/src/main/java/gov/cms/dpc/attribution/models/PatientEntity.java
nickrobison-usds: That's a good point. I think bene ids are unique, but I'll have to check.

nickrobison-usds: Created DPC-175 to track this. But I think we can go ahead and merge it.

PR #27, opened 2019-03-19
DPC-142: Add EOB-fetching functionality in BBclient

isears-cms:
BBclient needs to be able to fetch explanation of benefit data from the bluebutton server.

### DIFF
--- a/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/bbclient/DefaultBlueButtonClient.java
+++ b/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/bbclient/DefaultBlueButtonClient.java
@@ -27,7 +27,7 @@ public class DefaultBlueButtonClient implements BlueButtonClient {
 
     public Bundle requestEOBBundleFromServer(String patientID) throws ResourceNotFoundException {
         // TODO: need to implement some kind of pagination? EOB bundles can be HUGE
-        logger.debug("Attempting to fetch patient ID {} from baseURL: {}", patientID, client.getServerBase());
+        logger.debug("Attempting to fetch EOBs for patient ID {} from baseURL: {}", patientID, client.getServerBase());
         Bundle ret = client.search()
                 .forResource(ExplanationOfBenefit.class)
                 .where(ExplanationOfBenefit.PATIENT.hasId(patientID))
--- a/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/bbclient/DefaultBlueButtonClient.java
+++ b/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/bbclient/DefaultBlueButtonClient.java
@@ -2,6 +2,7 @@ package gov.cms.dpc.aggregation.bbclient;
 
 import ca.uhn.fhir.rest.client.api.IGenericClient;
 import ca.uhn.fhir.rest.server.exceptions.ResourceNotFoundException;
+import org.hl7.fhir.dstu3.model.Bundle;
 import org.hl7.fhir.dstu3.model.ExplanationOfBenefit;
 import org.hl7.fhir.dstu3.model.Patient;
 import org.slf4j.Logger;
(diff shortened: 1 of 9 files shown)


dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/bbclient/DefaultBlueButtonClient.java
```diff
@@ -17,18 +19,26 @@ public DefaultBlueButtonClient(IGenericClient client){
...
 
-        try {
-            logger.debug("Attempting to fetch patient ID {} from baseURL: {}", beneficiaryID, client.getServerBase());
-            return client.read().resource(Patient.class).withUrl(buildSearchUrl(beneficiaryID)).execute();
+    }
 
-        } catch (ResourceNotFoundException ex) {
-            throw new BlueButtonClientException("Could not find beneficiary with ID: " + beneficiaryID, ex);
+    public Bundle requestEOBBundleFromServer(String patientID) throws ResourceNotFoundException {
+        // TODO: need to implement some kind of pagination? EOB bundles can be HUGE
```
isears-cms: @nickrobison-usds do we need to do any kind of special pagination for potentially v. large EOB records returned from bluebutton?

nickrobison-usds: Yes we do.

isears-cms: According to the docs, the apache http client we're using does compression automatically unless we explicitly disable it (ref: http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/client/HttpClientBuilder.html#disableContentCompression()) 

So I think we're good there.

I'll do pagination in a new PR

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/bbclient/DefaultBlueButtonClient.java
```diff
@@ -17,18 +19,26 @@ public DefaultBlueButtonClient(IGenericClient client){
...
+        logger.debug("Attempting to fetch EOBs for patient ID {} from baseURL: {}", patientID, client.getServerBase());
+        Bundle ret = client.search()
+                .forResource(ExplanationOfBenefit.class)
+                .where(ExplanationOfBenefit.PATIENT.hasId(patientID))
+                .returnBundle(Bundle.class)
+                .execute();
+
+        if(ret.getTotal() == 0) {
+            throw new ResourceNotFoundException("Cannot find any EOBs for patient with ID: " + patientID);
```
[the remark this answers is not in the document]
isears-cms: This is a good point. I've updated with some logic to differentiate between those two cases, but the tricky thing is that there are no synthetic patients in the BlueButton sandbox w/0 EOBs ([Ref](https://github.com/CMSgov/bluebutton-site-static/blob/master/synthetic_users_by_claim_count_full.csv)), so I can't test this functionality against the BlueButton API.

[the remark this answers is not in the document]
nickrobison-usds: One concern that I have is that having 0 EOBs is not really an error, the record exists, it's just empty. So I don't think we should throw the same error for both cases.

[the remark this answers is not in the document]
isears-cms: I updated the logic to throw an exception when patient not found and just pass on the empty bundle received from bluebutton when the patient has no EoBs.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/bbclient/DefaultBlueButtonClient.java
```diff
@@ -17,18 +19,54 @@ public DefaultBlueButtonClient(IGenericClient client){
...
+     *  that contain no EoBs.
+     *
+     * @param patientID The requested patient's ID
+     * @return {@link Bundle} Containing a number (possibly 0) of {@link ExplanationOfBenefit} objects
+     * @throws ResourceNotFoundException when the requested patient does not exist
+     */
+    public Bundle requestEOBBundleFromServer(String patientID) throws ResourceNotFoundException {
+        // TODO: need to implement some kind of pagination? EOB bundles can be HUGE
+        logger.debug("Attempting to fetch EOBs for patient ID {} from baseURL: {}", patientID, client.getServerBase());
+        Bundle ret = client.search()
```
nickrobison-usds: Does BB throw a 404 error when requesting an non-existent patient?

PR #29, opened 2019-03-22
DPC-178: Turn off apache httpclient debug logging

isears-cms:


isears-cms:
I'm having trouble making this work within the application.conf. For example, 

```
logging {
    loggers {
      org.apache.http.wire = "WARN"
    }
}
```

Doesn't seem to have any effect. I think there's some issue with reaching into the apache http client set the log level from the application.conf, but I'm blocked trying to figure out where it is.

nickrobison-usds:
@isears-cms You need to put the config settings in the dpc-aggregation namespace, like so:

```
dpc.aggregation {
  logging {
    loggers {
      "gov.cms.dpc" = DEBUG
    }
  }
```

The aggregation service only reads config settings from within the path, everything else is ignored.

isears-cms:
I've tried it both inside and outside the namespace, but I'm still getting DEBUG level logs from the apache http client. I'll show you what I mean when I'm in this afternoon.

PR #30, opened 2019-03-22
DPC-164: Add jacoco plugin for tests

isears-cms:


nickrobison-usds:
Cool. Do we need to wire this into travis, in order to actually generate the coverage reports?

isears-cms:
Yeah, I was kinda hoping codeclimate would magically take care of everything, but it did not, so I'm still working on manually uploading the jacoco tests.

PR #34, opened 2019-03-28
DPC-176: Mock BlueButton API to avoid test network dependency

isears-cms:


dpc-aggregation/pom.xml
```diff
@@ -76,6 +76,24 @@
             <groupId>io.dropwizard</groupId>
             <artifactId>dropwizard-testing</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>2.24.0</version>
```
nickrobison-usds: Both the version and scope are set in the root pom, so we don't need to override them here. That way they'll automatically update along with all the other poms.

nickrobison-usds: Both the version and scope are set in the root pom, so we don't need to override them here. That way they'll automatically update along with all the other poms.

dpc-aggregation/pom.xml
```diff
@@ -76,6 +76,24 @@
...
+            <scope>test</scope>
+        </dependency>
+        <!-- https://mvnrepository.com/artifact/org.mock-server/mockserver-core -->
+        <dependency>
+            <groupId>org.mock-server</groupId>
+            <artifactId>mockserver-netty</artifactId>
+            <version>3.10.8</version>
```
nickrobison-usds: Can we pull this out into a property variable? That way both dependencies will be kept in sync?

local.application.conf
```diff
@@ -0,0 +1,8 @@
+dpc.attribution {
```
nickrobison-usds: This file shouldn't be committed. Looks like the gitignore failed to catch it because it's looking for application.local.conf. That should probably change to be more consistent.

nickrobison-usds: Can we remove this file from the commit set?

PR #35, opened 2019-03-28
DPC-169: Implement typesafe SQL

nickrobison-usds:
Added support for the JOOQ SQL framework, which means we now have a clean, typesafe interface for executing custom SQL queries.

This is primarily used in the `SeedCommand` class, since we can't inject the Hibernate context in a `Command` because it's not configured yet. This lets us remove a bunch of code and much more cleanly express our intent.

The Java classes are generated by JOOQ, based on our JPA entities. Due to the order of compilation, we had to move the entities in the `dpc-common` module, so that they would be available for JOOQ to parse. This add a step to the `generate-sources` phase of maven, but it should be handled automatically. The resulting classes are generated in the `target/` directory, so there's nothing for us to commit to git.

### DIFF
--- a/dpc-common/pom.xml
+++ b/dpc-common/pom.xml
@@ -34,5 +34,10 @@
             <groupId>javax.ws.rs</groupId>
             <artifactId>javax.ws.rs-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate.</groupId>
+            <artifactId>hibernate-core</artifactId>
+            <version>5.2.15.Final</version>
+        </dependency>
     </dependencies>
 </project>
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,10 +1,10 @@
-version: 2
+version: "2"
 checks:
   # IntelliJ generates suspicious equals/hashcode methods.
   # So we bump the complexity metric to 9, which should weed them out
   method-complexity:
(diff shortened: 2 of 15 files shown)


isears-cms:
I think this was an issue introduced in #33 that I didn't catch: when I run tests locally with the docker-compose provided psql container, the db connection fails because the tests are trying to log in with user "travis". We should either make everything work by default (i.e. `docker-compose --detach up db && mvn clean package`) or we should document the fact that specific config options are required in `application.local.conf` in order to start everything up.

dpc-common/pom.xml
```diff
@@ -34,5 +34,10 @@
             <groupId>javax.ws.rs</groupId>
             <artifactId>javax.ws.rs-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate.</groupId>
+            <artifactId>hibernate-core</artifactId>
+            <version>5.2.15.Final</version>
```
isears-cms: What's the rationale behind using 5.2 as opposed to 5.4? FWIW all the tests pass when I change this to `<version>5.4.2.Final</version>`

nickrobison-usds: Pinning it to the version pulled in by Dropwizard, that way we don't get any potential class conflicts in the future.

PR #36, opened 2019-04-03
DPC-22: Add task to automatically prune attribution relationships

nickrobison-usds:
Added a cron job to automatically expire attribution relationships that were created _before_ a given expiration threshold. The job is run and managed by Dropwizard using the Sundial job scheduler. It's currently configured to run everyday at midnight (based on the system clock).

The testing for it is rather simplistic, we essentially seed the database with attributions and manually override the creation time to be before the expiration threshold. DPC-186 has been created to improve the testing process by manually creating attributions using the current date time and verifying that they are not removed.

In the future, we may want to consider moving the JOOQ SQL into a stored procedure and potentially schedule it within the database itself, but we'll cross that bridge when we get to it.

This PR also includes some changes to the SeedCommand, by using JOOQ we've made things significantly simpler and safer.


dpc-attribution/src/test/java/gov/cms/dpc/attribution/AttributionResourceTest.java
```diff
@@ -19,6 +19,7 @@
 import java.io.IOException;
```
isears-cms: `com.fasterxml.jackson.databind.ObjectMapper` and  `org.apache.http.HttpEntity` are unused after this refactor.

dpc-attribution/src/test/java/gov/cms/dpc/attribution/jobs/ExpirationJobTest.java
```diff
@@ -0,0 +1,117 @@
...
+    void startJob(Client client, String jobName) {
+        Response response = client.target(
+                String.format(
+                        "http://localhost:%d/%s/tasks/startjob?JOB_NAME=%s",
+                        APPLICATION.getAdminPort(), APPLICATION.getEnvironment().getAdminContext().getContextPath(), jobName
+                ))
+                .request()
+                .post(Entity.text(""));
+
+        assertEquals(200, response.getStatus(), "Job should have started correctly");
```
isears-cms: We should be consistent about using HttpStatus.OK_200 (ref: line 72) vs just raw int '200' for checking status codes. Personally, I've been using raw ints but I like HttpStatus.OK_200 better.

PR #39, opened 2019-04-05
DPC-80: Proposal for Redis based Job Queue 

nickrobison-usds:
This is a proposed implementation for a Redis based job queue, in order to fully decouple the API service from the Aggregation Service, as proposed in DPC-80. Because this queue lives inside the host services, DPC-154 is completed as well, since it deploys along side the api and aggregation services.

This simplifies both the deployment and operational footprint by relying on services that we already have deployed (or will deploy) in the environments.
Redis serves as the main messaging layer between the two services, with Postgres serving as the persistence layer.

When a job is added to the queue, it's persisted in Postgres and the jobID is written to the Redis queue.
On the consuming side, the aggregation service monitors for jobs in the queue and then fetches the record from Postgres in order to get the job arguments. If the job is still in the correct state (e.g. *QUEUED*) it's moved into the *RUNNING* state and the job is processed.
Once the job is done, it is *completed* by setting the status to either *COMPLETED* or *FAILED*.
The API service can poll the queue for the job status, which is fetched from the Postgres table.

Unit tests have been implemented to check basic functionality and the functional E2E tests are passing as well.

While this queue fulfills the primary design goals, several improvements still need to be made:

1. It's possible for the queue to enter an inconsistent state where the queue and/or the database fail to persist correctly, leaving the job in only one service, but not the other. We either need to implement some locking or improve our rollback logic. We may also want to consider moving the update logic into the database, doing an update, only if the job is in the correct state.

1. The configuration for the redis queue is messy and spread out across the various service files. DPC-94 should help with this, but for now, it's pretty brittle.

1. The database migrate is currently handled by Hibernate itself, which is incredibly simplistic and unreliable, we'll need to implement some more robust schema migration using Liquibase.

1. Given that the proposed architecture uses Redis for caching patient data, we'll need to separate out the instances to ensure that no patient data crosses the data/app boundary.

1. We need to implement cleanup logic to purge completed jobs from the database tables, as well as handle any stale jobs that were not completed correctly.

1. When a job fails, it's not re-added to the queue, that might be a good idea for transient failures (e.g. IO or BB failures).

1. The job failure currently passes no failure messages, just the pass/fail status.


docker-compose.yml
```diff
@@ -12,18 +18,23 @@ services:
...
-      - "8081:8081"
+      - "8088:8081"
     environment:
       - ENV=dev
-
+    depends_on:
+      - db
+      - redis
+    volumes:
+      - export-volume:/app/data
```
ginabeena: What are the space limitations of /app/data?

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/DPCAggregationConfiguration.java
```diff
@@ -1,7 +1,37 @@
...
+import gov.cms.dpc.queue.DPCQueueConfig;
+import io.dropwizard.db.DataSourceFactory;
+import org.redisson.config.Config;
 
-public class DPCAggregationConfiguration extends TypesafeConfiguration {
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+import java.io.IOException;
 
+public class DPCAggregationConfiguration extends TypesafeConfiguration implements IDPCDatabase, DPCQueueConfig {
```
ginabeena: @RickHawesUSDS made a good point that maybe we should consider moving `TypesafeConfiguration` into the project rather than relying on an external oss product that has limited community base. What do you think?

dpc-aggregation/src/main/resources/application.conf
```diff
@@ -1,25 +1,35 @@
...
-      bbclient {
-          keyStore {
-              type = "JKS"
-              defaultPassword = "changeit"
-              location = "../bbcerts/bb.keystore"
-          }
+  bbclient {
+    keyStore {
+      type = "JKS"
+      defaultPassword = "changeit"
```
ginabeena: goteem

dpc-api/src/test/java/gov/cms/dpc/api/FHIRSubmissionTest.java
```diff
@@ -53,7 +53,7 @@ public void resetMocks() {
         // Mock the submission call to verify the job type
         doAnswer((answer -> {
             final JobModel data = answer.getArgument(1);
-            assertEquals(testJobModel, data, "Should have 4 beneies");
+            assertEquals(testJobModel.getPatients().size(), data.getPatients().size(), "Should have 4 beneies");
```
ginabeena: Typo

dpc-attribution/src/main/java/gov/cms/dpc/attribution/cli/SeedCommand.java
```diff
@@ -75,7 +75,7 @@ protected void run(Environment environment, Namespace namespace, DPCAttributionC
         final Timestamp creationTimestamp = generateTimestamp(namespace);
 
         // Read in the seeds file and write things
-        logger.info("Seeding attributions at time {}");
+        logger.info("Seeding attributions at time {}", creationTimestamp.toLocalDateTime());
```
ginabeena: Is this UTC and Unix?

dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedQueue.java
```diff
@@ -0,0 +1,137 @@
...
+        logger.debug("Adding job {} to the queue with data {}.", jobID, data);
+        // Persist the job in postgres
+        final Transaction tx = this.session.beginTransaction();
+        try {
+            this.session.save(data);
+            tx.commit();
+        } catch (Exception e) {
+            logger.error("Cannot add job to database", e);
+            tx.rollback();
+            throw new JobQueueFailure(jobID, e);
```
ginabeena: How do transactions play with different threading models? If you have a transaction and you are running this as a cooperative thread, then would it block until complete?

dpc-queue/src/main/java/gov/cms/dpc/queue/JobQueueModule.java
```diff
@@ -1,38 +1,49 @@
...
+        if (this.inMemory) {
+            binder.bind(JobQueue.class)
+                    .to(MemoryQueue.class)
+                    .in(Scopes.SINGLETON);
+        } else {
+            binder.bind(JobQueue.class)
+                    .to(DistributedQueue.class)
+                    .in(Scopes.SINGLETON);
+        }
     }
```
ginabeena: Is this the instance of the queue or something?... confused what the singleton pattern is here

docker-compose.yml
nickrobison-usds: The size of your disk. This simply maps /tmp into the docker container, which we only do for testing.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/DPCAggregationConfiguration.java
nickrobison-usds: Not a bad idea, currently we're using my fork of the library, which has some fixes. More than happy to vendor it.

dpc-attribution/src/main/java/gov/cms/dpc/attribution/cli/SeedCommand.java
nickrobison-usds: It's Unix, from the TZ of the machine it's running on.

dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedQueue.java
nickrobison-usds: I think we covered this in person, correct?

dpc-queue/src/main/java/gov/cms/dpc/queue/JobQueueModule.java
nickrobison-usds: Yes, it means that there will only be a single queue, which is shared amongst the threads.

PR #41, opened 2019-04-12
DPC-94: Unified Dropwizard configuration

nickrobison-usds:
This PR is mainly around reworking the configuration files so that we have single points of contact for making changes.

Each service now has its options isolated into its own namespace (e.g. `dpc.attribution{}`), which means we can override settings for each service as we need to.

In order to reduce config duplication, most of the common settings (e.g. service ports, database connection properties, logging, etc) have been moved into a `server.conf` file in the `src/main/resources` directory. Each service can then `include "server.conf" within its namespace and inherit the default settings, which it is free to override as it sees fit.
In order for this to work correctly, each service needs to include the `src/main/resources` directory in its build, which is accomplished by adding:

```xml
<resources>
  <resource>
    <directory>${project.basedir}/../src/main/resources</directory>
   </resource>
</resources>
```

under the `<build>` key of the pom.

Likewise, the `dpc-queue` module now ships its own `reference.conf` file which includes its default values as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,9 +26,14 @@ docker-compose up db redis
 By default, the application attempts to connect to the `dpc_atrribution` database on the localhost as the `postgres` user.
 This database needs to be manually created, but table setup and data migration will be handled by the DPC services.
 
-For Redis, we assume the server is running on the localhost, with the default port. 
+For Redis, we assume the server is running on the localhost, with the default port.
 
 The defaults can be overridden in the configuration files.
+Common configuration options (such as database connection strings) are stored in the `server.conf` file within the `src/main/resources` directory.
+These settings are included in the various modules via the `include "server.conf"` attribute in module application config files.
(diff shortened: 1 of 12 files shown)


RickHawesUSDS:
One question. I ran into some configuration problems running the API server tests in IntelliJ over the weekend. When run tests from IntelliJ my local.application.conf in the top-level folder was ignored. I solved my problem by temporarily modifying the API server application.conf. I noticed that the Aggregation server solved this problem by add a test.application.conf to the test resources which seems like a better solution. With this change, what would be the best way to get tests to run under IntelliJ?

nickrobison-usds:
That's a good question. I added a note in the README regarding the IntelliJ quirks. The test.application.conf files are a good idea. The challenge is making sure that we don't commit our environment specific values, which could eventually include passwords and such. That was the goal behind the application.local file. It's probably worth revisiting, since it seems to keep causing issues.

isears-cms:
Ship it!

PR #43, opened 2019-04-19
DPC-23: Add ability to resubmit roster

nickrobison-usds:
**Why**

This PR adds the ability for vendors to add/remove attribution relationships to their existing rosters. Duplicate records are handled using an *upsert* methodology, where new records are inserted as they are, conflicting records are merged using a user defined function.

**What changed**

The external API did not change much at all. No additional endpoints were created. The majority of the changes are internal, with largest being that I replaced the existing Roster logic (which was using Hibernate) to an alternative implementation using JOOQ. This allows for a much cleaner implementation and type-safe SQL, which is generated from the Hibernate entities.
This also adds logic for merging new data into existing records, in order to do this, I've created an `AbstractRecordUpserter` class that allows specializations for specific record types (e.g. Patients, Providers, etc). It mainly allows you to specify which columns to *exclude* for updating.

**Choices Made**
Moved away from Hibernate because it's not really designed to support the more complex querying that we need to do. It was also proving to be really brittle when trying to retrieve related entities from different projections. JOOQ makes it much cleaner and simpler. Hibernate still has value so we won't entirely get rid of it, but it definitely has query limitations.

**Testing Done**
 - Attribution Service integration tests have been expanded to test roster updating
- Manually verified that relationships are indeed updated.
- Updated expiration test to add a new attribution relationship and ensure it gets left around

**Future work**
- Operations are wrapped within atomic transactions but each operation is executed as a separate SQL statement. This may cause an issue with large attribution lists, in the future we may want to look into SQL batch operations to reduce the size of the transaction log.
- We should add the ability to retrieve rosters and individual patient/provider records. At the very least, this will help with debugging/administration but my be useful to expose externally.

**Tickets closed**
- DPC-23: Ticket for this PR
- DPC-170: We now handle duplicate relationships and patient/provider records
- DPC-186: Verified that expiration job only expires job older than the cutoff value.

### DIFF
--- a/dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/RosterUtils.java
+++ b/dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/RosterUtils.java
@@ -10,6 +10,8 @@ import org.hl7.fhir.dstu3.model.Patient;
 import org.hl7.fhir.dstu3.model.Practitioner;
 import org.hl7.fhir.dstu3.model.ResourceType;
 import org.jooq.DSLContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.sql.Timestamp;
 import java.util.UUID;
@@ -18,6 +20,8 @@ import static gov.cms.dpc.attribution.dao.tables.Patients.PATIENTS;
 import static gov.cms.dpc.attribution.dao.tables.Providers.PROVIDERS;
 
 public class RosterUtils {
+
+    private static final Logger logger = LoggerFactory.getLogger(RosterUtils.class);
 
     /**
      * Helper method for adding a roster {@link Bundle} to the RDBMS backend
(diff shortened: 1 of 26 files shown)


dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/RosterUtils.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Helper method for adding a roster {@link Bundle} to the RDBMS backend
+     *
+     * @param attributionBundle - Roster {@link Bundle} to add/update
+     * @param ctx               - {@link DSLContext} DB context to utilize
+     * @param creationTimestamp - {@link Timestamp} of when the attribution relationships were updated
+     */
+    public static void handleAttributionBundle(Bundle attributionBundle, DSLContext ctx, Timestamp creationTimestamp) {
```
RickHawesUSDS: Suggest that this name can have a more descriptive verb than "handle". Isn't this where most of the business logic happens?

dpc-attribution/src/test/java/gov/cms/dpc/attribution/AttributionFHIRTest.java
```diff
@@ -102,26 +98,55 @@ private void submitRoster(Bundle bundle) throws Exception {
...
+                assertEquals(bundle.getEntry().size(), patients.size(), "Should have an additional patient");
+            }
+
+            // Check that a specific patient is attributed
+            final HttpGet updatedAttributed = new HttpGet(String.format("http://localhost:%d/v1/Group/%s/%s", APPLICATION.getLocalPort(), providerID, newPatientID));
+            updatedAttributed.setHeader("Accept", FHIRMediaTypes.FHIR_JSON);
+
+            try (CloseableHttpResponse response = client.execute(updatedAttributed)) {
+                assertEquals(HttpStatus.OK_200, response.getStatusLine().getStatusCode(), "Should be attributed");
+            }
```
RickHawesUSDS: This test is getting very large. Could it be broken up?

dpc-attribution/src/test/java/gov/cms/dpc/attribution/AttributionFHIRTest.java
```diff
@@ -30,6 +32,7 @@
 import java.util.function.Function;
 import java.util.stream.Stream;
 
+import static gov.cms.dpc.attribution.SharedMethods.createAttributionBundle;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class AttributionFHIRTest {
```
RickHawesUSDS: I'm expecting in addition to the submitTests to have removeTests?

nickrobison-usds: That's a good point. I looked into it briefly and there are a couple of issues pertaining to really supporting deletion in a _non-hacky_ way. I've opened DPC-209 to track the issue, which we can land as a separate PR.

PR #44, opened 2019-04-24
DPC-203: Encrypt temp filesystem resources

isears-cms:
**Why**
Patient data needs to be encrypted at rest on the temp filesystem.

**What Changed**
A new `EncryptingAggregationEngine` was created with associated tests.

**Choices Made**

- Vendors supply an RSA public key, which is used to encrypt an AES secret key (generated within DPC), which is used to ultimately encrypt the data. The encrypted data, as well as the encrypted AES secret key, is passed back to the user who must then (1) decrypt the AES secret key with his/her RSA private key and (2) decrypt the data with the AES secret key.
- Use `javax.crypto` libraries, rather than libsodium
- Encryption metadata (e.g. symmetric encryption algorithm, asymmetric encryption algorithm, initialization vector) is written to the temp filesystem alongside the encrypted files in a -metadata.json file.
- AES/GCM is used to provide both confidentiality and integrity guarantees
- AES 128-bit keys are used to reduce encryption overhead

**Issues Fixed**
DPC-203


dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/EncryptingAggregationEngine.java
```diff
@@ -0,0 +1,151 @@
...
+import java.security.GeneralSecurityException;
+import java.security.SecureRandom;
+import java.util.Base64;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public class EncryptingAggregationEngine extends AggregationEngine {
+
+    private static final String SYMMETRIC_CIPHER =  "AES/GCM/NoPadding";
```
nickrobison-usds: Should these values be pulled out into the config?

dpc-queue/src/main/java/gov/cms/dpc/queue/models/JobModel.java
```diff
@@ -57,6 +58,8 @@ public static String outputFileName(UUID jobID, ResourceType resourceType) {
 
     private JobStatus status;
 
+    private RSAPublicKey rsaPublicKey;
```
nickrobison-usds: Do we know how this will be serialized in the database? Would it be better to simply hex encode the key and send it along as a byte array?

isears-cms:
WRT thought 2: Java automatically rejects RSA keys with weak parameters e.g. trying to create an RSA key with exponent less than 3 results in the following exception:

```
java.security.spec.InvalidKeySpecException: java.security.InvalidKeyException: exponent is smaller than 3
```
So I think any tests for that are most applicable to the code that actually takes the vendor's raw input and turns it into an RSA key. That functionality is out of the scope of this PR

PR #45, opened 2019-04-25
DPC 110 - JobCompletion record improvement 

RickHawesUSDS:
**Why**
To match the FHIR bulk export spec as listed in [FHIR Bulk Data Export Implementation Guide](https://github.com/smart-on-fhir/fhir-bulk-data-docs/blob/master/export.md)

**What Changed**
No major changes. Added output records for the JobCompletionModel. JobModel had timestamp fields added to track the time the job processing took. The logs contain info records now with this info. Splunk should be able to pick these records up. 

**Choices Made**
The bulk export job completion record has changed over time. FHIR 3.3 doesn't have it. FHIR4 is not finalized. The specs appears to be converging to the what was described in the implementors guide.

**What's Next**
DPC-205 will fill in the error field. 

**Issues Fixed**
DPC-110


dpc-api/src/main/java/gov/cms/dpc/api/models/JobCompletionModel.java
```diff
@@ -1,25 +1,38 @@
...
+
+    /**
+     * The full request of the original request URL
+     */
     private String request;
     private final boolean requiresAccessToken = false;
-    private List<String> output;
-    // FIXME(nickrobison): Should return errors as well. OperationOutcomes, serialized as NDJSON
+    private List<OutputEntryModel> output;
+    // FIXME(rickhawes): DPC-205 will fill in this array.
```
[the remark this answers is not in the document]
RickHawesUSDS: Will fix this in the next pull request.

dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedQueue.java
```diff
@@ -31,8 +33,9 @@
 
     @Override
     public void submitJob(UUID jobID, JobModel data) {
-        assert(jobID == data.getJobID());
+        assert(jobID == data.getJobID() && data.getStatus() == JobStatus.QUEUED);
```
nickrobison-usds: No assert here.

dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedQueue.java
```diff
@@ -111,14 +120,21 @@ public void submitJob(UUID jobID, JobModel data) {
     @Override
     public void completeJob(UUID jobID, JobStatus status) {
         assert(status == JobStatus.COMPLETED || status == JobStatus.FAILED);
```
nickrobison-usds: No assert.

dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedQueue.java
```diff
@@ -111,14 +120,21 @@ public void submitJob(UUID jobID, JobModel data) {
...
-        logger.debug("Completing job {} with status {}.", jobID, status);
-
         final Transaction tx = this.session.beginTransaction();
         try {
             final JobModel jobModel = this.session.get(JobModel.class, jobID);
+            if (jobModel.getStatus() != JobStatus.RUNNING) {
+                throw new JobQueueFailure(jobID, String.format("Cannot complete job in state: %s", jobModel.getStatus()));
+            }
+            assert(jobModel.getStartTime().isPresent());
```
nickrobison-usds: No assert

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
```diff
@@ -49,10 +53,14 @@ public Response checkJobStatus(@PathParam("jobID") String jobID) {
                     break;
                 }
                 case COMPLETED: {
+                    assert(job.getCompleteTime().isPresent());
+                    final var resourceQueryParam = job.getResourceTypes().stream()
+                            .map(type -> type.toString())
```
nickrobison-usds: This should be a method reference

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
```diff
@@ -49,10 +53,14 @@ public Response checkJobStatus(@PathParam("jobID") String jobID) {
...
+                            .collect(Collectors.joining(GroupResource.LIST_DELIM));
                     final JobCompletionModel completionModel = new JobCompletionModel(
-                            Instant.now().atOffset(ZoneOffset.UTC),
-                            String.format("%s/Job/%s", baseURL, jobID),
-                            outputURLs(jobUUID, job.getResourceTypes()));
+                            job.getCompleteTime().get(),
```
nickrobison-usds: I believe this should be `startTime()` per the FHIR spec:

```
transactionTime - a FHIR instant type that indicates the server's time when the query is run.
The response SHOULD NOT include any resources modified after this instant, and SHALL include any matching resources modified up to (and including) this instant.
Note: to properly meet these constraints, a FHIR Server might need to wait for any pending transactions to resolve in its database, before starting the export process.
```

Is that your interpretation as well?

dpc-api/src/main/java/gov/cms/dpc/api/models/OutputEntryModel.java
```diff
@@ -0,0 +1,27 @@
+package gov.cms.dpc.api.models;
+
+import org.hl7.fhir.dstu3.model.ResourceType;
+
+/**
+ * An entry in the {@link JobCompletionModel} output field.
+ */
+public class OutputEntryModel {
```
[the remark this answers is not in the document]
RickHawesUSDS: Agreed.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
```diff
@@ -49,10 +53,14 @@ public Response checkJobStatus(@PathParam("jobID") String jobID) {
                     break;
                 }
                 case COMPLETED: {
+                    assert(job.getCompleteTime().isPresent());
```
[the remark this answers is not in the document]
RickHawesUSDS: I disagree about that asserts don't add value. They help document the logic of the code and they fire during debugging/unit-tests. It's like a compiled comment. 

True, a db fetch should be checked and an assert doesn't do it. So I'll add the throw, but I like to see asserts in code.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
RickHawesUSDS: Ok.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
RickHawesUSDS: No. I'm worried about 

> The response SHOULD NOT include any resources modified after this instant

My thinking is that BB can update a resource during the aggregation of DPC's response which would be **after** the job start time. True? To make jobStart time not violate this constraint, we would be need the ability to query BB with a time filter.

RickHawesUSDS: After discussing with Nick, this is a tricky problem. Using completeTime introduces an error where some returned records could not be up to date. In the end we decided to use startTime since will probably to lead to a provider requesting updates since this time.

dpc-api/src/test/java/gov/cms/dpc/api/resources/JobResourceTest.java
```diff
@@ -101,10 +102,10 @@ public void testSuccessfulJob() {
 
         // Test the completion model
         final var completion = (JobCompletionModel) response.getEntity();
-        JobModel.validResourceTypes.stream().forEach(resourceType -> {
-            var expectedURL = String.format("%s/Data/%s", TEST_BASEURL, JobModel.outputFileName(jobID, resourceType));
-            assertTrue(completion.getOutput().contains(expectedURL));
-        });
+        for (OutputEntryModel entry: completion.getOutput()) {
+            assertTrue(JobModel.validResourceTypes.contains(entry.type));
```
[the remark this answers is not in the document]
RickHawesUSDS: Agreed.

dpc-queue/src/main/java/gov/cms/dpc/queue/models/JobModel.java
```diff
@@ -105,6 +106,20 @@ public JobModel(UUID jobID, List<ResourceType> resourceTypes, String providerID,
         this.status = JobStatus.QUEUED;
     }
 
+    /**
+     * Is the job model fields consistent. Useful before and after serialization.
+     *
+     * @return True if the fields are consistent with each other
+     */
+    public Boolean isValid() {
```
nickrobison-usds: I like this!

PR #46, opened 2019-04-25
DPC-206: Add demo client and re-enable integration tests

nickrobison-usds:
**Why**

This PR adds a small demo application within the `dpc-api` service that illustrates an example flow through the entire application. This includes submitting a roster, starting an export, polling for the export to complete and returning the download URLs for the files.

**What changed**

The `dpc-api` service now has an additional `demo` command, which runs the test workflow against a fresh instance of dpc-app. This resulted in some refactoring of the database handling as well as some test helper utilities.

The integration tests have been updated to run on each commit, this required some modification of Travis CI setup in order to persist the code coverage numbers over both the unit test and integration test runs.
We also no longer rely on the built-in Postgres and Redis services; instead, we use the ones provided by our docker-compose file.
There is a now a new maven profile `integration-tests` that only runs tests marked with the `@IntegrationTest` annotation. Eventually, this should probably be merged into a larger _CI_ profile, but I wanted to be explicit about it 

**Choices Made**
Probably the most impactful choice is to create an admin _task_ in the attribution service that allows the tests to automatically truncate the attribution database. This is needed in order to allow the integration tests to be run from a clean slate and the admin port will **NOT** be exposed outside the Docker container in production. DPC-210 tracks actually disabling that command in non-dev/test environments.

**Testing Done**
Verified that the entire test suite (including integration) tests run and pass on my local machine and Travis.

**Future work**
DPC-210: Disable DB truncation in production.

**Tickets closed**
DPC-206: Ticket for this PR
DPC-163: Re-enabled integration tests under Travis.
DPC-184: Exclude JOOQ generated files from Jacoco analysis


dpc-aggregation/src/main/resources/dev.application.conf
```diff
@@ -1,13 +1,8 @@
 # Override the keystore location to point to correct location when run within docker "dev" environment
 dpc.aggregation {
   database {
-    driverClass = org.postgresql.Driver
     url = "jdbc:postgresql://db:5432/dpc_attribution"
-    user = postgres
```
RickHawesUSDS: Why remove the user?

dpc-api/src/main/java/gov/cms/dpc/api/cli/DemoCommand.java
```diff
@@ -0,0 +1,122 @@
...
+            // Retry the export request
+            System.out.println("Retrying export request");
+            String exportURL = "";
+
+            try {
+                exportOperation.execute();
+            } catch (NonFhirResponseException e) {
+                final NonFhirResponseException e1 = e;
+                if (e1.getStatusCode() != HttpStatus.NO_CONTENT_204) {
+                    e.printStackTrace();
```
RickHawesUSDS: Perhaps a demo shouldn't print a stack trace?

dpc-api/src/main/java/gov/cms/dpc/api/client/ClientUtils.java
```diff
@@ -0,0 +1,145 @@
...
+        try (CloseableHttpClient client = HttpClients.createDefault()) {
+            final HttpGet jobGet = new HttpGet(jobLocation);
+            boolean done = false;
+
+            while (!done) {
+                Thread.sleep(1000);
+                logger.debug(statusMessage);
+                try (CloseableHttpResponse response = client.execute(jobGet)) {
+                    final int statusCode = response.getStatusLine().getStatusCode();
+                    done = statusCode == HttpStatus.OK_200 || statusCode > 300;
```
RickHawesUSDS: Why is statusCode > 300 treated as the same as OK? Suggest that the demo should error and exit?

dpc-aggregation/src/main/resources/dev.application.conf
nickrobison-usds: It's already specified in the base config. So this file should only list the values which are overriden in the dev environment.

dpc-api/src/main/java/gov/cms/dpc/api/cli/DemoCommand.java
nickrobison-usds: Good call!

dpc-api/src/main/java/gov/cms/dpc/api/client/ClientUtils.java
nickrobison-usds: Correct.

PR #49, opened 2019-05-03
DPC-215: Improve Application Healthchecks

nickrobison-usds:
**Why**
In the AWS environment, the Application Load Balancers require an endpoint that returns the health of the application, in order to make scaling decisions. Dropwizard includes a built-in Healthcheck system, but we need to expand it in order to take into account the various services that we use.

**What Changed**
Each Service has expanded health checks that validate the health of their external dependencies (e.g. the aggregation service checks Blue Button and the Job Queue). These checks can be manually run by sending a `GET` request to: `http://{hostname}:{admin port}/healthcheck`.

In addition, the Attribution Service now has a `_healthy` endpoint, which lets the API service ensure it's reachable.

**Choices Made**
The Job Queue includes separate checks for the Postgres database and the Redis cluster.

The Attribution Service's `_healthy` endpoint always returns a 200 status. The thinking behind this is that the API service really only cares about accessibility an can rely on the Attribution Service to report it's own health status. This should hopefully avoid cascading database checks.

**Tickets closed**
DPC-215


dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedQueue.java
```diff
@@ -130,34 +141,61 @@ public String queueType() {
         return "Redis Queue";
     }
 
+    @Override
+    public void isHealthy() {
```
RickHawesUSDS: I expect that a method named isHealthy would return a Boolean like isEmpty, isBlank, isNull etc do. It's in a quite a number of style guides that I've followed.

dpc-api/src/main/java/gov/cms/dpc/api/health/AttributionHealthCheck.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    @Override
+    protected Result check() throws Exception {
+        if (engine.isHealthy()) {
+            return Result.healthy();
+        }
+
+        return Result.healthy("Cannot reach Attribution Service");
+    }
```
RickHawesUSDS: Shouldn't this be a return of Result.unhealthy when engine.isHealthy is false.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/health/JobQueueHealthCheck.java
```diff
@@ -0,0 +1,27 @@
...
+    public JobQueueHealthCheck(JobQueue queue) {
+        this.queue = queue;
+    }
+
+    @Override
+    protected Result check() throws Exception {
+        this.queue.isHealthy();
+        return Result.healthy();
+    }
+}
```
RickHawesUSDS: As I read DropWizzard's docs, a check function should return Result.unhealthy() when a check is unhealthy.

dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/RosterEngine.java
```diff
@@ -135,4 +135,9 @@ public boolean isAttributed(Practitioner provider, Patient patient) {
                         .and(PROVIDERS.PROVIDER_ID
                                 .eq(FHIRExtractors.getProviderNPI(provider)))));
     }
+
+    @Override
+    public boolean isHealthy() {
+        return true;
```
RickHawesUSDS: Suggest that a better health check would be to try to get a record from the attribution db.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/health/JobQueueHealthCheck.java
nickrobison-usds: The underlying health check library automatically handles catching thrown exceptions and wrapping them in an unhealthy response, with stack trace. See [here](https://metrics.dropwizard.io/3.1.0/manual/healthchecks/).

Should be more explicit about the errors that we're throwing?

dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedQueue.java
nickrobison-usds: Agreed, this is confusing. How does `assertHealthy` sound? It seems easiest to throw an exception when unhealthy, that way we at least preserve the reason why the health check failed.

RickHawesUSDS: Ok. assertHealthy works.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/health/JobQueueHealthCheck.java
RickHawesUSDS: I believe that health checks shouldn't throw exceptions when they detect an unhealthy state. That is being unhealthy is a normal response, and shouldn't be an exception. If an exception happens, logging it is appropriate.

dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/RosterEngine.java
nickrobison-usds: Yup. Good call. Done.

dpc-api/src/main/java/gov/cms/dpc/api/health/AttributionHealthCheck.java
nickrobison-usds: Fixed!

dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedQueue.java
nickrobison-usds: Done!

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/health/JobQueueHealthCheck.java
nickrobison-usds: I believe I addressed this in the latest set of commits.

PR #51, opened 2019-05-09
DPC-229: Fix failing Travis integration tests

nickrobison-usds:
**Why**
Last week, integration tests started in failing on Travis when reaching out to the BB devcloud sandbox. The issue was tracked down to opening an SSL connection to BB, which for some reason was never being closed and hanging the entire application.

This PR makes a couple of changes to the aggregation engine to better handle BB failures and clean up some of the logging. The end result was that things simply started 'working', so it seems like a network issue on the Travis side of things.

**What Changed**
The main changes were refactoring the Queue polling operations in the `Aggregation Engine` from a polling loop, to a polling reactive pipeline. This has a couple of benefits.

1. There's less code for polling/retry as rxJava has some built-in operators for doing so.
1. We have improved error handling that allows us to deal with each patient request individually
1. The network requests are executed on their own thread pool, which gives us better resource utilization and firewalling of the main process. It also means we get parallel processing for 'free'.
1. We're able to easily integrate with the retry handler provided by [resilience4j](http://resilience4j.github.io/resilience4j/#_retry) so within the reactive pipeline we can automatically retry a request failure and be more resilient to failures.

We also added configuration options for setting the network timeouts for the BB HTTP client, these default to 5 seconds, but can be overridden within the config.

This also includes some refactoring of configuration handling, particularly moving the BB client config values into the main aggregation service configuration. This means we now get typesafe operators for getting/setting config values.

**Choices Made**
Moving the queue polling to a reactive pipeline does introduce some drawbacks.

1. It's a new programming paradigm that has its own edge cases and limitations.
1. It makes stacktraces less readable 
1. It's another set of dependencies for us to bundle.

We should decide whether or not the tradeoffs are worthwhile. Forgoing the reactive implementation means we need to handle individual export failures and retry operations ourselves, which is definitely possible.

**Tickets closed**
DPC-229: Original integration test failure ticket
DPC-146: Add retry logic to BB client.

**Future work**
DPC-145:  Improve thread pool handling of network requests and encryption/serialization.
DPC-230: Cleanup configuration settings for EncryptingAggregationEngine


dpc-aggregation/pom.xml
```diff
@@ -81,6 +82,21 @@
...
+            <groupId>io.reactivex.rxjava2</groupId>
+            <artifactId>rxjava</artifactId>
+            <version>2.2.8</version>
+        </dependency>
+        <dependency>
+            <groupId>io.github.resilience4j</groupId>
+            <artifactId>resilience4j-retry</artifactId>
+            <version>${resilience4j.version}</version>
+        </dependency>
+        <dependency>
```
RickHawesUSDS: We should discuss where we are going with microservice control/resilience. I would argue for something like Netflix Hysterix, but I would argue that we need to do it along with BB and BCDA.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
```diff
@@ -128,38 +124,96 @@ public void completeJob(JobModel job) {
...
                         writer.write(DELIM);
                     } catch (IOException e) {
                         throw new JobQueueFailure(job.getJobID(), e);
                     }
                 });
     }
+
+    private void pollQueue() {
+        subscribe = Observable.fromCallable(this.queue::workJob)
+                .doOnNext(job -> logger.trace("Awaiting job from queue"))
```
RickHawesUSDS: If I read the code, "Awaiting job from queue" is logged after the call to workJob completes. The comment seems miss leading.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
```diff
@@ -128,38 +124,96 @@ public void completeJob(JobModel job) {
...
+                .filter(Optional::isPresent)
+                .repeatWhen(completed -> {
+                    logger.debug("No job, retrying in 2 seconds");
+                    return completed.delay(2, TimeUnit.SECONDS);
+                })
+                .map(Optional::get)
```
RickHawesUSDS: Minor quible. Suggest that the filter and map operators should be right after each other, since they deal with the onNext event, while the onError and onComplete handlers should be at the end of the chain.

dpc-aggregation/pom.xml
nickrobison-usds: That sounds like a great idea. I looked into Hystrix, but I think it's in [maintenance mode](https://github.com/Netflix/Hystrix) now?

I would definitely lean on your recommendation.

PR #55, opened 2019-05-15
DPC-242: Update capabilities statement

nickrobison-usds:
**Why**
FHIR clients use the capabilities statement to determine which operations they can perform against our system. Previously, the statement was really bare bones and hadn't been updated with the new Roster, Job and Data endpoints. That's now been fixed.

**What Changed**
Updated the Capabilities Statement to include the new endpoints.
Added a generic ability to inject Maven variables (build time values) into the Java runtime, so we can now automatically use the build timestamp and version number in our capabilities statement.

**Choices Made**
I'm intentionally keeping things simple for now. The statements are really verbose which means the best option going forward is to find a way to autogenerate values using something like ByteBuddy.

**Tickets closed**
DPC-242

**Future Work**
DPC-125: Expand the amount of information available in the capabilities statement
DPC-66: We should automatically generate the capabilities statement from our annotated FHIR resources.

### DIFF
--- a/dpc-api/src/main/java/gov/cms/dpc/api/core/Capabilities.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/core/Capabilities.java
@@ -62,9 +62,25 @@ public class Capabilities {
         final CapabilityStatementRestOperationComponent metadataResource = new CapabilityStatementRestOperationComponent(new StringType("Metadata"), new Reference(String.format("%s/metadata", baseURI)));
         final CapabilityStatementRestOperationComponent versionResource = new CapabilityStatementRestOperationComponent(new StringType("Version"), new Reference(String.format("%s/_version", baseURI)));
 
-        // Add the Group export paths
-        final CapabilityStatementRestOperationComponent providerExport = new CapabilityStatement.CapabilityStatementRestOperationComponent(new StringType("Provider export"), new Reference(String.format("%s/Group/providerID/$export", baseURI)));
-        serverComponent.setOperation(Arrays.asList(providerExport, metadataResource, versionResource));
+        //  Group resources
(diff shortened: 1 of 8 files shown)


dpc-api/src/main/java/gov/cms/dpc/api/core/Capabilities.java
```diff
@@ -1,32 +1,35 @@
...
 
+import java.time.format.DateTimeFormatter;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
 import static org.hl7.fhir.dstu3.model.CapabilityStatement.*;
 
 public class Capabilities {
+    private static final DateTimeFormatter FHIR_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss");
```
RickHawesUSDS: As I read the FHIR spec, Instant and DateTime fields should have a time zone. If you look at OffsetDateTimeToStringConverter you see. 

`fhirFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSxxx")`

We should have this magic formatter string defined in one place. I suggest that we move this pattern and formatter to the common module.

nickrobison-usds: @RickHawesUSDS FHIR DateTime does not support offset seconds, which means we can't use the same instant formatter:

```
A date, date-time or partial date (e.g. just year or year + month) as used in human communication. The format is YYYY, YYYY-MM, YYYY-MM-DD or YYYY-MM-DDThh:mm:ss+zz:zz, e.g. 2018, 1973-06, 1905-08-23, 2015-02-07T13:28:17-05:00 or 2017-01-01T00:00:00.000Z. If hours and minutes are specified, a time zone SHALL be populated. Seconds must be provided due to schema type constraints but may be zero-filled and may be ignored at receiver discretion. 
```

I've moved both formatters into a helper class and added java docs on what they're used for.

RickHawesUSDS:
Look's good, Nick.

PR #58, opened 2019-05-30
DPC-12: Export specific patients

nickrobison-usds:
**Why**
One of our use cases is to support exporting a subset of patients associated to a given provider. For example, a hospital might want to perform an export operation on all of the patients that are currently scheduled for an appointment in the coming week. Given that the bulk spec currently only only allows exporting the entire dataset for a given group, we need a way to indicate which subset of patients should be exported.

**What Changed**
This PR adds an additional method to the `Group/{:group id of provider or organization}/$export` endpoint.
Users can now `POST` an ad-hoc `Group` resource to the endpoint and kick off an export request that will only return data for the patients included in the group.

Each group member MUST include an `Identifier` which references the patient MBI. An example is given below:

```javascript
{
	"resourceType": "Group",
	"id": "temporary-group-id",
	"type": "person",
	"actual": true,
	"member": [
		{
			"entity": {
				"identifier": {
					"system": "http://bluebutton.cms.hhs.gov/identifier#bene_id",
					"value": "patient-one-bene-id"
				}
			}
		},
		{
			"entity": {
				"identifier": {
					"system": "http://bluebutton.cms.hhs.gov/identifier#bene_id",
					"value": "patient-two-bene-id"
				}
			}
		},
		{
			"entity": {
				"identifier": {
					"system": "http://bluebutton.cms.hhs.gov/identifier#bene_id",
					"value": "patient-three-bene-id"
				}
			}
		}
	]
}
```

In the event that a Group is submitted which includes any patient IDs for which the provider or organization is not authorized the export job is rejected and an `OperationOutcome` is returned which indicates which patients may not be included. The client then has the option to either add the patients to their roster or remove them from the ad-hoc group.
This process happens synchronously, before the job is submitted for execution. 

**Choices Made**
The export request is set to fail quickly if _any_ of the requested patents are not attributed to the provider. The result is an `OperationOutcome` that indicates which patients are unattributed. The rest of the roster and export flow matches the normal bulk export flow.

**Tickets closed**
DPC-12: Implement ad-hoc group export

**Future Work**
DPC-25: The attribution logic does not infer any relationships from the claim history, simply performs a lookup in the database and returns a true/false result. This will be added in a future ticket.
DPC-32: The `OperationOutcome` should give the user a method for programmatically determining which patients are _not_ attributed to the provider. Currently, we return the data in a text field, but we'll need to develop a cleaner way of communicating the result.
DPC-261: Implement StructureDefinition for ensuring required data elements.


jmandel:
Thanks @nickrobison-usds! The basic functionality here makes sense, but it seems to be overloading existing export operation. Operations in FHIR can already be called with POST (see http://build.fhir.org/operations.html#executing), so there's no way to add new semantics to `POST /Group/:if/$export` by passing in an entirely new body in this way. If you really wanted to do something like this, you could define a new parameter like `limit-to-subset` and then invoke the entire operation via POST with a Parameters body including existing params like `_type` as well as the new optional param. This would be a bit surprising for anyone calling the API, because people are not very familiar with the Parameters resource -- but at least would fit within the constraints of FHIR today.

nickrobison-usds:
@jmandel Thanks for the feedback! I didn't realize that POST was already defined, so we definitely don't want to overload that.

What you're saying makes sense, seems like the Parameters approach would be a bit clunky to work with so we should probably steer clear. That seems to leave us with the two-step approach of defining a custom group and then invoking export.

I appreciate your help and patience with this!

PR #60, opened 2019-06-04
DPC-13: Organization Registration

nickrobison-usds:
**Why**
We need the ability to register an organization, in order to given them access to the API

**What Changed**
DPC now has a process for manually registering partner organizations. This includes the FHIR endpoints and the Dropwizard command that's been added to support registration.

We've added a `organization.tmpl.json` file in the `src/main/resources` directory. This can be used as a basis for creating and registering new organizations and endpoints.

The basic structure of the resource is a `Bundle` with a least 2 resources. The first is a FHIR `Organization` that includes an organization name, address, NPI and at least one contact with a name, address and contact point (email or phone).

We also require at least one registered endpoint, and as we don't currently have a way to require that it be a FHIR endpoint, we'll rely on soft power to make it happen.

This PR also adds some conversion classes translating from FHIR resources and elements into our Hibernate entities. It's pretty manual at this point, but DPC-276 should help improve things. 

**Choices Made**
The `Organization` and `Endpoint` endpoints are built into the Attribution service and thus only exposed to the API service and internal CMS net systems. This avoids the need to immediately require an authorization scheme in order to avoid bogus organizations being created, but does mean that we'll have to come up with a solution in the future, if we want our organizations to be able to update their own records.

There isn't much in the way of resource validation, mostly because we're waiting on DPC-255 to implement true FHIR validation using structure definitions and such, but also because the manual registration process means we can just enforce the constraints ourselves.

The overall testing story is not great and largely waiting on DPC-{278,279} which will give us the ability to read resources back out of the database and verify that they're correct. I've manually verified that the entries are created correctly in the database.

**Tickets closed**
DPC-13: Closes this registration story
DPC-201: Added org registration command

**Future Work**
DPC-276: Refactor FHIR conversion utilities into single converter class.
DPC-277: Catch and report Hibernate errors when using FHIR error handler
DPC-255: Implement validations for Organization and Endpoint resources
DPC-278: Add READ operation to Organization and Endpoints endpoints
DPC-279: Add SEARCH operation to Organization and Endpoints endpoints


dpc-common/src/main/java/gov/cms/dpc/common/entities/OrganizationEntity.java
```diff
@@ -0,0 +1,186 @@
...
+
+        // Find the first Organization ID that we can use
+        final Optional<Identifier> identifier = resource
+                .getIdentifier()
+                .stream()
+                // Don't support UNKNOWN systems for now, only things we can use
+                .filter(resourceID -> {
+                    final String system = resourceID.getSystem();
+                    // If it's null, we can use it
+                    try {
```
RickHawesUSDS: I'm not sure what is going on here with "null" values being valid. Is there a better way to express the logic?

dpc-common/src/main/java/gov/cms/dpc/common/entities/OrganizationEntity.java
```diff
@@ -0,0 +1,186 @@
...
+        // Add the entity reference
+        contactEntities.forEach(contact -> contact.setOrganization(entity));
+        entity.setContacts(contactEntities);
+
+        return entity;
+    }
+
+    @Override
+    public Organization toFHIR() {
+        throw new UnsupportedOperationException("Not implemented yet");
```
RickHawesUSDS: perhaps a todo comment?

dpc-common/src/main/java/gov/cms/dpc/common/entities/OrganizationEntity.java
```diff
@@ -0,0 +1,186 @@
...
+        final OrganizationEntity entity = new OrganizationEntity();
+
+        // Find the first Organization ID that we can use
+        final Optional<Identifier> identifier = resource
+                .getIdentifier()
+                .stream()
+                // Don't support UNKNOWN systems for now, only things we can use
+                .filter(resourceID -> {
+                    final String system = resourceID.getSystem();
+                    // If it's null, we can use it
```
RickHawesUSDS: Not sure what is meant by this comment.

dpc-common/src/main/java/gov/cms/dpc/common/entities/OrganizationEntity.java
nickrobison-usds: Done!

dpc-common/src/main/java/gov/cms/dpc/common/entities/OrganizationEntity.java
nickrobison-usds: It doesn't mean anything, it's a leftover from a previous implementation. I've removed it. Good catch!

PR #65, opened 2019-06-12
DPC-17: Create API token for organizations

nickrobison-usds:
**Why**

This PR adds the token generation required in order to successfully secure the API endpoints. It currently only contains the generation changes, not the actual security components.

**What Changed**

Added two token endpoints to the `Organization` resource. `token/create` and `token/verify`. Which do pretty much what you'd expect. When creating a token, we check to see if one already exists and if so, we throw an error. The user can force creation of a new token by passing the `refresh` query parameter. Token verification simply returns a true/false as to whether or not the token is valid.

We've also added a `dpc-macaroons` module which abstracts most of the token generation logic. It's patterned off of the excellent [macaroon-bakery](https://github.com/go-macaroon-bakery/macaroon-bakery) library. At this point, it's pretty incomplete but the goal is to eventually expand its capabilities and split it off into its own, more general purpose, package. For now, we're trying to keep things 'simple'.

When constructing the bakery, we give it a default set of caveats that always get applied and then can provide some additional caveats as we go along.

**Choices Made**

Probably the most important choice is the use of Macaroons as the underlying tokens. We're currently using them as opaque tokens, which no special features, but eventually we'll expand their usage.

**Tickets closed**:
DPC-17

**Future Work**
DPC-30: Implement endpoint token validation
DPC-284: Make root key expiration configurable
DPC-285: Improve error messaging in verification
DPC-286: Add support for third-party macaroon caveats


nickrobison-usds:
Merging this even though the PR tests are failing. This issue is due to some poor DB connection handling, which is fixed as part of #77.

PR #73, opened 2019-06-25
DPC-289: Address default application startup issues

nickrobison-usds:
**Why**

Per initial reports from users, the documentation is woefully out of date as to how to actually start and test the application. This PR not only addresses the documentation issues, but also fixes a reported bug where the `seed` command was not working properly.

**What Changed**

Made a number of tweaks to the `README.md` file to hopefully make it a bit clearer as to how to actually get started.

Added a new `application.local.conf.sample` file, which includes all the necessary overrides for running the application on a developer machine. The users can now copy this file and everything should work correctly. Instructions have been added to the README.

Fixed an issue where the `seed` command was not working correctly when called from within the UberJAR due to the `InputStream` closing prematurely. A couple of small tweaks to the way we pass the resource references between the `SeedCommand` and the `SeedProcessor` seems to have resolved the issue.

**Choices Made**

Public methods on the `SeedProcessor` class are now static which requires us to directly provide an input stream (pointing to the attributions csv) on each invocation. This avoids issues where the calling class might accidentally close the InputStream before we're done with it. This was a simple tweak, given that we don't actually need any state.

**Tickets closed**:

DPC-289: Documentation and default settings cleanup.
DPC-287: Fixed closed InputStream when running seed command from JAR

**Future Work**

DPC-291: Improve handling of local configuration overrides.


dpc-attribution/src/main/java/gov/cms/dpc/attribution/cli/SeedCommand.java
```diff
@@ -76,11 +65,18 @@ protected void run(Environment environment, Namespace namespace, DPCAttributionC
             context.truncate(Providers.PROVIDERS).cascade().execute();
             context.truncate(Organizations.ORGANIZATIONS).cascade().execute();
 
-            this.seedProcessor
-                    .extractProviderMap()
+
+            // Get the test seeds
+            final InputStream resource = SeedCommand.class.getClassLoader().getResourceAsStream(CSV);
```
RickHawesUSDS: Shouldn't this go in a try() block to close the input stream.

nickrobison-usds:
Merging this, even though the tests are failing. It seems like Travis is failing to install Java for some reason, but tests are passing on my machine.

PR #74, opened 2019-06-26
create api documentation route

jaredcunha-usds:
**Why**

Create a route for API documentation and page in `views/public`. http://localhost:3000/docs should now resolve.

**Choices Made**

Used `match` method since that's what we were using for the landing page. 

**Future Work**

- Breaking up some of the partials and layout files to better improve the design of this site.  
- Don't quite know what will be included in the documentation. Will it be a single page or include multiple sections? I think we are going to follow Bue Button 2.0 for now, which is a single page.

### DIFF
--- a/dpc-web/app/views/public/docs.html.erb
+++ b/dpc-web/app/views/public/docs.html.erb
@@ -1,0 +1,2 @@
+
+<h1>API Documentation</h1>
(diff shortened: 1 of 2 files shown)


nickrobison-usds:
@jaredcunha-usds Thanks for working on this! A couple of question thoughts.

1. Can we get some JIRA tickets created to track these items? I think we have a `Website` component in JIRA which we can use.
1. I'm not sure if you've had a chance to take a look at our 'developer documentation': https://confluenceent.cms.gov/display/DAPC/Developer+Workflow. Primarily, we're trying to organization branches by username/jira-ticket-description.

Let me know if you have any other questions!

PR #79, opened 2019-07-01
Create API documents controller

jaredcunha-usds:
**Why**

Need to set up controller for API documentation

**What Changed**

- Created a new controller the API Documentation
- Organized layout files to support appropriate content
- Set up a better way of handling page titles.

**Choices Made**

- New controller for API documentation would the section up better for future needs
- Need page help to organize page titles outside of main content, also in the `<title>` tag.

**Tickets closed**:

https://jiraent.cms.gov/browse/DPC-299


jaredcunha-usds:
@nickrobison-usds I haven't gotten to that point just yet, but that's what I was thinking. The landing might have to be HTML since it's going to be need to be organized in a grid for presentation, but I think the API docs would be a heck of a lot of easier if we used markdown. I installed the `rouge` gem already, but haven't look too deeply into markdown in rails (still a little rusty since I haven't done much rails work since 2016..

PR #81, opened 2019-07-02
DPC-30: Add token authentication and validation

nickrobison-usds:
**Why**
This PR adds the fundamental components for enabling token base authentication against the API. While its doesn't actually require authentication against endpoints, yet, it puts the major pieces in place.

Due to some refactoring, this PR got a little out of hand, but also encompasses changes to the DB seeding process and adds a public facing `Organization` endpoint.

We'll need to spend a fair amount of time going over the auth changes, to ensure that they're actually correct.

**What Changed**

The smaller changes are pretty easy to describe:

Added a public facing `Organization` endpoint that allows an Organization to retrieve the data that we have about them. This is _read only_ and doesn't support extra features such as search, as it's tied directly to the organization's security token. We use this endpoint to test the basic authentication functionality.

With DPC-309 I made a number of changes to the DB seed command. The major change is that we're now loading synthetic data generated by [synthea](https://synthea.mitre.org). I've added some basic details to a README file in the `src/main/resources` directory. Basically, I manually ran some scripts to pull down fake data and then enriched it to ensure that it meets our standards. We then load the FHIR bundles as part of the Seed command.

Now the main event. Authentication.

By taking advantage of the built-in `dropwizard-auth` support, we now have the ability to annotate which endpoints should be secured and which permissions are required for access. This is done through the annotations in the `cms.gov.dpc.api.auth.annotations` package. The `@PathAuthorizer` annotation is the primary method for providing endpoint security, it allows specifying a specific path component to then verify against the auth backend. We're currently using it to verify that a token is linked to the organization being requested. It's fairly brittle at this point, but DPC-371 should take care of that.

We also have a `@Public` annotation, which we can use to specify that an endpoint has no security requirements. This gives us the ability to write a unit test that ensures all endpoints have an explicit security decision and that we haven't neglected to secure anything.

The authorization logic is handled by the `MacaroonsDynamicFeature` which automatically loads the auth handlers for the required endpoints.  The `MacaroonsAuthFilter` class does the initial check to see if the token is present and valid for the given Organization. If so, it passes the token do the `MacaroonsAuthenticator` class, which does the actual permissions validation. We also have a `MacaroonsAuthorizer` class, which is wired up, but disabled as we're not doing fine-grained permissions verification, yet.

Authentication is enabled by default, but can be disabled by passing `authenticationDisabled=true` as a configuration option.

**Choices Made**

The main choice was to use Macaroons for the opaque auth token, but seems to be ok so far. I also decided not to roll out auth yet, mostly due to PR size, but also because of the ongoing API churn. We also need to make sure that we can disable auth for testing and local development purposes.

**Tickets closed**:

DPC-30: This ticket
DPC-309: Refactor resource seeding
DPC-305: Add Organization resource

**Future Work**
DPC-308: Add endpoint resource
DPC-306: Add new endpoints to capabilities statement
DPC-313: Enable authentication on all endpoints
DPC-370: Move permissions logic to Macaroons
DPC-371: Make PathAuthorizer more generic
DPC-372: Ensure we can't disable authentication in production.


dpc-api/src/main/java/gov/cms/dpc/api/auth/annotations/PathAuthorizer.java
```diff
@@ -0,0 +1,23 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * {@link Security} annotation which indicates that a given path component must be used to match security against the provided token
+ */
+@Documented
+@Retention(RUNTIME)
+@Target({TYPE, METHOD})
+@Security
+public @interface PathAuthorizer {
```
RickHawesUSDS: Is the idea of a PathAuthorizer the same as "Scope" in the OAuth world?

dpc-api/src/main/java/gov/cms/dpc/api/DPCAPIConfiguration.java
```diff
@@ -65,6 +67,14 @@ public void setExportPath(String exportPath) {
...
     }
 
+    public boolean isEncryptionDisabled() {
+        return encryptionDisabled;
+    }
+
+    public void setEncryptionDisabled(boolean encryptionDisabled) {
+        this.encryptionDisabled = encryptionDisabled;
+    }
+
```
RickHawesUSDS: Suggest that "encryption" is not the right name for this configuration. If I understand it, it controls authorization validation. Right?

dpc-api/src/main/java/gov/cms/dpc/api/auth/MacaroonsAuthFilter.java
```diff
@@ -0,0 +1,92 @@
...
+
+        if (macaroon == null) {
+            macaroon = requestContext.getUriInfo().getQueryParameters().getFirst(TOKEN_URI_PARAM);
+        }
+
+        // We need to verify that the Macaroon is valid
+        final String pathValue = requestContext.getUriInfo().getPathParameters().getFirst(this.pa.pathParam());
+        if (pathValue == null) {
+            throw new WebApplicationException(unauthorizedHandler.buildResponse(BEARER_PREFIX, realm));
+        }
```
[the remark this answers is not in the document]
nickrobison-usds: The unauthorized handler adds the appropriate error codes.

dpc-api/src/main/java/gov/cms/dpc/api/auth/annotations/PathAuthorizer.java
nickrobison-usds: It's similar, this annotation just allows us to pull a parameter from the request path. That way we can verify the token is valid for the given org. Does that answer your question?

dpc-api/src/main/java/gov/cms/dpc/api/auth/MacaroonsAuthFilter.java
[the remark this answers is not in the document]
RickHawesUSDS: Didn't know. Thanks.

dpc-api/src/main/java/gov/cms/dpc/api/auth/annotations/PathAuthorizer.java
RickHawesUSDS: Yes. Thanks.

PR #82, opened 2019-07-03
DPC-257: Add Prefer Header to $export operation

RickHawesUSDS:
**Why**

Tighten up the header and parameter checking in the $export operation to be in compliance with the [FHIR Bulk Data Export Implementation Guide](https://github.com/HL7/bulk-data/blob/master/spec/export/index.md) and what BCDA does. 

**What Changed**

- Required a `Accept: application/fhir+json` header
- Required a `Prefer: respond-async` header
- Check that the `_since` parameter is not specified
- Check that the `_outputFormat` parameter is not specified or `application/fhir+ndjson`
- Returned an `OperationOutcome` JSON structure when bad request parameters where found

**Choices Made**

- Choose to return an OperationOutcome on errors to be like BCDA
- For the same reason, choose not to use Jersey/DropWizard parameter validators because they do not return OperationOutcome compatible bodies.  
- Created a separate @FHIRAsync annotation to compliment the @FHIR annotation. 

**Tickets closed**

DPC-257: Ensure export requests have async header set

**Future Work**


dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
```diff
@@ -114,18 +138,105 @@ public Patient marshalTest(Group group) {
...
+     * @param headers to check
+     * @param resourceTypes list to check
+     * @param since list to check (must be null)
+     * @return a optional {@link Response} with a BAD_REQUEST status
+     */
+    public static Optional<OperationOutcome> checkExportRequest(MultivaluedMap<String, String> headers, String resourceTypes, String outputFormat, String since) {
+        var issues = new ArrayList<OperationOutcome.OperationOutcomeIssueComponent>();
+
+        // Accept
+        final List<String> accepts = headers.get(HttpHeaders.ACCEPT);
```
nickrobison-usds: The `@FHIR` annotation should handle this check before calling the method. `FHIRRequestFilter` has the appropriate logic.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
```diff
@@ -114,18 +138,105 @@ public Patient marshalTest(Group group) {
...
+            issues.add(formNotSupportedIssue("'_since' query parameter is not implemented"));
+        }
+
+        // _outputFormat
+        if (StringUtil.isNotEmpty(outputFormat) && !FHIR_NDJSON.equals(outputFormat)) {
+            issues.add(formNotSupportedIssue("'_outputFormat' query parameter must be 'application/fhir+ndjson'"));
+        }
+
+        // _type
+        if (StringUtils.isNotEmpty(resourceTypes)) {
```
nickrobison-usds: Would it be possible to use the built-in Dropwizard validators? Like [so](https://www.dropwizard.io/1.3.12/docs/manual/core.html#parameters)?

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
RickHawesUSDS: OK.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
RickHawesUSDS: Good idea, but I decided not to go this way. Reason: These validators run before the FHIR filters that turn exceptions into Operation Outcomes. Other validations return a nice operational outcome like BCDA does. If we use Jersey / Dropwizard validators then we wouldn't get the FHIR operation outcome message.

PR #87, opened 2019-07-10
DPC-139: Implement FHIR validation support

nickrobison-usds:
**Why**
This PR adds the basic scaffolding for adding [profiling](http://www.hl7.org/fhir/profiling.html) support to DPC. This allows us to be explicit about the data content and layout that we require and communicate that requirement to clients.

**What Changed**
This adds the base definition for our Patient requirements and integrates our custom profiles with both the FHIR validator and Dropwizard's built-in Hibernate validation framework.

By adding the `@Profiled` annotation to a given resource, we can specify a profile URL that the resource must comply with and Dropwizard takes it from there.

Getting the integration to work was pretty tricky. Dropwizard doesn't support injecting resources into Validators, which means, by default, we can't reference the FHIR Validator in our custom classes. To fix this, we had to override the default Validation Factories and wrap them up in a way that Guice is able to identify and inject into. It ended up being a lot of boilerplate, but the end result seems pretty good.

This also adds a `StructureDefinition` endpoint, so that clients can fetch what they need from us.

**Choices Made**

This PR also includes a fair amount of refactoring in order to get all the injection and configuration working correctly. I think it will be worth in in the long run, because it allows us to create FHIR resources as just another type of Object, from Dropwizard's perspective. 

The actual validation hasn't been wired up yet, we'll address that in future PRs, in order to avoid this getting entirely out of control.

**Tickets closed**:
DPC-139: Partially completed in this PR.
DPC-405: Added structure definition endpoints
DPC-258: Created matcheable patient definition.
DPC-404: Authorization tests are randomly failing.

**Future Work**
DPC-403: Migrate manual validator binding to Guice
DPC-406: Enable debug validation of generated resources


nickrobison-usds:
@RickHawesUSDS This is ready for review, even though the tests are failing due to DPC-405

PR #88, opened 2019-07-10
Jared/user guide dpc 401

jaredcunha-usds:
**Why**

Let's set up the draft markdown page. The content will still need some work. 

**Choices Made**

Not setting up the page layout just yet in case we want to change how the content might be organized.

**Tickets closed**:

https://jiraent.cms.gov/browse/DPC-401

### DIFF
--- a/dpc-web/app/views/api_documents/user_guide.html.erb
+++ b/dpc-web/app/views/api_documents/user_guide.html.erb
@@ -1,0 +1,5 @@
+<% title "User guide" %>
+
+<p>Content for User Guide</p>
+
+<%= raw(@html_content) %>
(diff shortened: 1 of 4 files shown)


dpc-web/app/views/api_documents/user_guide.html.erb
```diff
@@ -0,0 +1,5 @@
+<% title "User guide" %>
+
+<p>Content for User Guide</p>
+
+<%= raw(@html_content) %>
```
nickrobison-usds: Dumb question, is this handling the Markdown rendering?

jaredcunha-usds: Yes. This is what @mphprogrammer had set up to include the markdown content on the page. 

Basically, `<%= raw(@html_content) %>` on page with a file name of `foo.html.erb` is going to pull in content from a file named `foo.md`. The file names will need to remain the same for each page.

nickrobison-usds: Great. Thanks!

PR #90, opened 2019-07-12
Update headers needed for manual testing in README

anniesullie-gov:
**Why**

Easier setup for manual testing

**What Changed**

Added documentation on a header which should be set.


anniesullie-gov:
> Awesome! One request. Can you add yourself to the developers list in the pom.xml file?

Done.

PR #92, opened 2019-07-14
DPC-320: Reduce default threadpool size

nickrobison-usds:
**Why**

During integration testing, we're opening a ton of database connections (nearly 100) since the connection pool will automatically scale to 100. Since Postgres only allows 100 connections by default, this means we quickly run into errors.

**What Changed**

Reduced the default thread pool size, and the maximum value as well. We now start 5 DB connections and can flex up to 10. This prevents Postgres from throwing exceptions when too many clients are open. Such as when rapidly running integration tests where the connection pool doesn't get cleaned up in time.

This can (and will be) be overridden in higher environments.

**Choices Made**
Small for now, with the theory that we'll always override in production, plus our unit of scaling isn't making the individual services bigger, it's adding more services.

**Tickets closed**:
DPC-320

**Future Work**
None

### DIFF
--- a/src/main/resources/server.conf
+++ b/src/main/resources/server.conf
@@ -5,6 +5,9 @@ database {
   url = "jdbc:postgresql://localhost:5432/dpc_attribution"
   user = postgres
   password = dpc-safe
+  initialSize = 5
+  minSize = 5
+  maxSize = 10
 }
 
 server {


ronaldheft-usds:
It sounds idle connection eviction may not be working. I agree with the smaller pool size, but we should check when doing load testing that the connection pool will drop idle connections after scaling up. The settings to configure this are listed here: https://www.dropwizard.io/0.7.1/docs/manual/jdbi.html

(Would log a ticket but don't have access yet.)

nickrobison-usds:
@ronaldheft-usds Good call. I created DPC-418 to track this issue. Thanks!

PR #93, opened 2019-07-15
DPC-303: Link providers to organizations

nickrobison-usds:
**Why**
Now that we've added authentication and authorization, we need a way to link Practitioners to an Organization, so that way can keep separate between entities.

This PR adds that support as well as makes the necessary changes for threading the organization context through the request chain.

**What Changed**

Added a `PractitionerRole` endpoint to the attribution service, which is how we link a provider to a given organization. When creating a practitioner, they're not actually useable until they're assigned. The API service does the assignment automatically, by first creating the practitioner and then issuing a separate request to create the role. This should eventually be cleaned up.

We now have the ability to 'disable' authentication on the API server. Since we need to know which organization is actually making the request, we can't disable it entirely. Instead, we switch to a 'static' authenticator, which always returns a test organization. You can override the default organization by setting the `Organization` header on a given request.

Registering an organization now happens through the `$submit` operation on the `Organization` endpoint. Yes, I know we hate dollar sign operators, but it seems like the simplest way to getting the necessary context into the creation request.

The `Practitioner` endpoint in the Attribution service now request that each request specify which Organization is performing the operation. This is done via the `_tag` parameter in the request.

**Choices Made**
Using the `_tag` parameter for authentication is definitely not a great idea, but the implementation was much quicker than defining a custom extension.

**Tickets closed**:
DPC-303
DPC-93: Fixed test configuration overrides
DPC-371: PathAuthorizer is more generic.
DPC-415: Authentication is now partially disable-able.

**Future Work**

List any additional tickets that have either been created due to work in this PR, or existing tickets that expand upon the feature or provide additonal fixes.


nickrobison-usds:
@RickHawesUSDS @ronaldheft-usds This PR is ready for review, the failing PR test will be fixed as part of #101.

dpc-attribution/src/main/resources/migrations.xml
```diff
@@ -245,4 +245,39 @@
...
+                referencedColumnNames="id"/>
+        <addForeignKeyConstraint
+                baseTableName="PROVIDER_ROLES"
+                baseColumnNames="provider_id"
+                constraintName="fk_provider_roles_providers"
+                referencedTableName="PROVIDERS"
+                referencedColumnNames="id"/>
+    </changeSet>
+
+    <changeSet id="remove provider constraint" author="nickrobison-usds">
```
ronaldheft-usds: The id on this change set using spaces instead of hyphens and all our other change sets are using hyphens.

dpc-common/src/main/java/gov/cms/dpc/common/entities/ProviderRoleEntity.java
```diff
@@ -0,0 +1,110 @@
...
+        uniqueConstraints = {@UniqueConstraint(columnNames = {"organization_id", "provider_id"})})
+public class ProviderRoleEntity implements Serializable {
+
+    public static final long serialVersionUID = 42L;
+
+    @Id
+    @Column(name = "id", updatable = false, nullable = false, columnDefinition = "uuid")
+    private UUID roleID;
+
+    @OneToOne(fetch = FetchType.LAZY)
```
ronaldheft-usds: I don't think this is a problem with how everything is setup now, but I have had major issues in the past trying to annotate @OneToOne and @ManyToOne annotations with lazy fetching. Hibernate has trouble determining if it needs to proxy the method or just return null. I've seen results where either the lazy fetch doesn't work and it's still fetched eagerly, or in some cases exceptions when Hibernate does proxy the method, but has trouble fetching the object when requested.

This tends to be problem on the side of the relationship where the id isn't in the table and Hibernate can't determine if it's null or not, and it looks like everywhere we've done this so far we're safe. Just something to be aware of, as I've felt the pain in the past. 😬

For more info: https://discourse.hibernate.org/t/hibernate-5-1-10-lazy-loading-doesnt-work/401

nickrobison-usds: This is good to know. Thanks!

PR #94, opened 2019-07-15
DPC-328: Docker container for website

ronaldheft-usds:
**Why**

The DPC hosting infrastructure will involve hosting the website in a Docker container on ECS. To prepare going live with the website, the static website should be dockerized for development testing in preparation for standing up staging and production environments.

**What Changed**

- A Docker file and docker-compose.yml file has been added to dpc-web to facilitate building the docker container.
- The dpc-web readme documentation has been updated with instructions for building the docker container.
- The dpc-web Ruby configuration has been tweaked to allow for environment variable overrides needed by the Docker container.

**Choices Made**

The base Ruby docker image was based off `ruby:2.6.2-alpine`. This is an official Docker image provided by Ruby. The Alpine Linux variant was chosen as it's smaller and simpler. Due to this, some additional dependencies had to be installed in the image to allow for building of the ruby gems.

Instead of creating a separate environment configuration for Docker, the existing Ruby development environment was made more configurable by allowing overriding of database variables using environment variables.

**Tickets closed**:

DPC-328

**Future Work**

Additional environment configuration will be needed to use this docker image in staging and production, but none of it is docker specific. The Ruby credentials file will be updated with database credentials and host information, and then Docker container will be started with `RUBY_ENV` and `RAILS_MASTER_KEY` environment variables to use the correct environment and decrypt the credentials file.

### DIFF
--- a/dpc-web/Dockerfile
+++ b/dpc-web/Dockerfile
@@ -1,0 +1,23 @@
+FROM ruby:2.6.2-alpine
+
+# Install build dependencies
+RUN apk add --no-cache postgresql-dev alpine-sdk npm tzdata
+
+# Set the working directory
+RUN mkdir /dpc-web
+WORKDIR /dpc-web
+
+# Copy over the files needed to fetch dependencies
+COPY Gemfile /dpc-web/Gemfile
+COPY Gemfile.lock /dpc-web/Gemfile.lock
+COPY package.json /dpc-web/package.json
+COPY package-lock.json /dpc-web/package-lock.json
+
+# Install the website dependencies
+RUN gem install bundler --no-document && bundle install && npm install
+
+# Copy the website
+COPY . /dpc-web
+
+# Start the rails server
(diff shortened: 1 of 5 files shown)


dpc-web/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM ruby:2.6.2-alpine
+
+# Install build dependencies
+RUN apk add --no-cache postgresql-dev alpine-sdk npm tzdata
```
nickrobison-usds: Does this update the apk registry each time?

ronaldheft-usds: @nickrobison-usds Yes. The Alpine Linux base Docker container does not ship with a package registry, so it needs to be fetched anyway. They provide a `--no-cache` option to delete the registry after to make the Docker image smaller. For more information: https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md

PR #96, opened 2019-07-16
DPC-331: Integrate Swagger

nickrobison-usds:
**Why**
Application developers are used to consuming API documentation via Swagger (OpenAPI) documents. Even though most of our endpoints are FHIR resources (and thus expect their own Implementation Guide), this is still really helpful and with doing.

**What Changed**
Added the `dropwizard-swagger` bundle to both the API and Attribution services. This allows us to annotate our public facing resources with the appropriate datums and have the `swagger.json` auto generate for us.

In order to ensure we don't miss anything, added simple unit tests to the API and Attribution services which verifies that all Jersey routes have at least an `@ApiOperation` annotation.

Also updated the PR template to include documentation tasks, we'll need to ensure that we keep things up to date.

As part of the documentation, any method that returns a resource was modified to correctly return a 201 status, rather than 200. This means we're more compliant with the FHIR spec (and HAPI).

**Choices Made**

Short descriptions for now, will be updated going forward.

**Tickets closed**:
DPC-331: This ticket

**Future Work**
DPC-418: Update example requests and responses
DPC-419: Add links to external documentation
DPC-420: Align Swagger UI styling with DPC website


dpc-api/src/test/java/gov/cms/dpc/api/APIResourceAnnotationTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+        methods = methods.stream()
+                .filter(method -> !Modifier.isAbstract(method.getModifiers()))
+                .collect(Collectors.toUnmodifiableSet());
+
+        assertFalse(methods.isEmpty(), "Should have annotated methods");
+    }
+
+    @Test
+    void allResourcesHaveSwaggerAnnotations() {
```
ronaldheft-usds: Slick test for checking all the methods are annotated!

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/DefinitionResource.java
```diff
@@ -39,6 +45,8 @@ public Bundle getStructureDefinitions() {
     @Override
     @GET
     @Path("/{definitionID}")
+    @FHIR
+    @ApiOperation(value = "Fetch specific structure definition", notes = "FHIR endpoitn to fetch a specific structure definition from server.", response = StructureDefinition.class)
```
RickHawesUSDS: typo in "endpoitn"

RickHawesUSDS: also, missing a "the" at the end.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
```diff
@@ -62,9 +63,20 @@ public GroupResource(JobQueue queue, AttributionEngine client, @APIV1 String bas
     @Timed
     @ExceptionMetered
     @FHIRAsync
-    public Response export(@PathParam("providerID") String providerID,
+    @ApiOperation(value = "Begin export request", tags = {"Group", "Bulk Data"},
+            notes = "FHIR $export operation which initiates a bulk data export for the given Provider")
```
RickHawesUSDS: perhaps "$" is not needed or appropriate in the description

RickHawesUSDS: A question, the `Accept` header must be `application/fhir+ndjson` is that implied or should it be added as a param.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
```diff
@@ -46,6 +51,16 @@ public JobResource(JobQueue queue, @APIV1 String baseURL) {
     @GET
     @Timed
     @ExceptionMetered
+    @ApiOperation(value = "Check export job status",
+            notes = "This endpoint is used to query the status of a given Export operation. " +
+                    "When the job is in progress, the API returns a 204 status." +
```
RickHawesUSDS: Should be API returns a 202 ACCEPTED status

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
```diff
@@ -46,6 +51,16 @@ public JobResource(JobQueue queue, @APIV1 String baseURL) {
...
+                        "When completed, an output response is returned, which contains the necessary metadata for retrieving any output files.")
+    @ApiResponses({
+            @ApiResponse(code = 202, message = "Export job is in progress"),
+            @ApiResponse(code = 404, message = "Export job cannot be found"),
+            @ApiResponse(code = 500, message = "Export job has failed with no results"),
+            @ApiResponse(code = 200, message = "Export job has completed with any failures listed in the response body", response = JobCompletionModel.class)
```
RickHawesUSDS: Should read "without" any failures.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
nickrobison-usds: Good catch. That should be added to all the FHIR endpoints.

PR #97, opened 2019-07-16
Migrate EndToEndRequestTest to postman

anniesullie-gov:
**Why**

We think using postman will make it easier to understand what the tests are doing. Developers can load them into the UI, and inspect all requests/responses.

**What Changed**

1. I added a `src/test/EndToEndRequestTest.postman_collection.json` file, exported from postman GUI, which does the same actions as `EndToEndRequestTest.java`, but at a smaller scale, since postman couldn't handle the large volume of response data from 100 patients.
2. I added a `src/main/resources/attribution_bundle.json` file, which represents the format in which a sample roster is sent over the wire, and is used in the test.
3. I updated `.travis.yml` to automatically run the postman tests.
4. I removed the old java `EndToEndRequestTest`.


src/test/EndToEndRequestTest.postman_collection.json
```diff
@@ -0,0 +1,364 @@
+{
+	"info": {
+		"_postman_id": "46f1266f-6a54-4a33-9d31-3ea944875a35",
+		"name": "EndToEndRequestTest",
+		"description": "This automates EndToEndRequestTest in postman.",
```
anniesullie-gov: Do you think the description field for the test and description fields in the steps are good enough places to put documentation on how the test works? Feedback on improving documentation welcome!

src/test/EndToEndRequestTest.postman_collection.json
```diff
@@ -0,0 +1,364 @@
...
+		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
+	},
+	"item": [
```
anniesullie-gov: Each item represents a request. I added separate "items" for the first export attempt that 404s and the export attempt that works correctly after adding the roster, because although it's possible to add logic in the `exec` field, I thought it would be confusing to combine them.

src/test/EndToEndRequestTest.postman_collection.json
```diff
@@ -0,0 +1,364 @@
...
+							"    pm.response.to.have.status(404);",
+							"});",
+							"// Response body should include:",
+							"// The resourceType \"OperationOutcome\"",
+							"// A single issue, severity fatal",
+							"pm.test(\"Response body has correct error information\", function() {",
+							"    pm.expect(pm.response.json().resourceType).to.equal(\"OperationOutcome\");",
+							"    pm.expect(pm.response.json().issue).to.have.lengthOf(1);",
+							"    pm.expect(pm.response.json().issue[0].severity).to.equal(\"fatal\");",
+							"});"
```
anniesullie-gov: The `exec` fields in each step are where the response is verified.

They also contain some logic, the most complex of which is in the Job Response step (to retry on 202, and parse on 200)

src/test/EndToEndRequestTest.postman_collection.json
```diff
@@ -0,0 +1,364 @@
...
+						"type": "text"
+					},
+					{
+						"key": "Prefer",
+						"value": "respond-async",
+						"type": "text"
+					}
+				],
+				"url": {
+					"raw": "http://localhost:3002/v1/Group/{{PROVIDER_ID}}/$export",
```
anniesullie-gov: This `PROVIDER_ID` variable is set wayyyy at the bottom in the `event` field.

src/test/EndToEndRequestTest.postman_collection.json
```diff
@@ -0,0 +1,364 @@
+{
+	"info": {
+		"_postman_id": "46f1266f-6a54-4a33-9d31-3ea944875a35",
+		"name": "EndToEndRequestTest",
+		"description": "This automates EndToEndRequestTest in postman.",
+		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
+	},
+	"item": [
+		{
+			"name": "First export operation",
```
anniesullie-gov: Would love feedback on good names/descriptions for these, to make it more self documenting.

src/test/EndToEndRequestTest.postman_collection.json
```diff
@@ -0,0 +1,364 @@
...
+						"key": "Prefer",
+						"value": "respond-async",
+						"type": "text"
+					}
+				],
+				"url": {
+					"raw": "http://localhost:3002/v1/Group/{{PROVIDER_ID}}/$export",
+					"protocol": "http",
+					"host": [
+						"localhost"
```
anniesullie-gov: We'd probably want variables for host and port eventually?

src/test/EndToEndRequestTest.postman_collection.json
```diff
@@ -0,0 +1,364 @@
...
+						"key": "Content-Type",
+						"name": "Content-Type",
+						"value": "application/fhir+json",
+						"type": "text"
+					}
+				],
+				"body": {
+					"mode": "file",
+					"file": {
+						"src": "../main/resources/attribution_bundle.json"
```
anniesullie-gov: **PROBLEM**: This path is relative to where the test is run from. So right now this test only works if you run it from `src/test`. Any ideas on a good approach to making it more robust?

src/test/EndToEndRequestTest.postman_collection.json
```diff
@@ -0,0 +1,364 @@
...
+		{
+			"name": "Job response",
+			"event": [
+				{
+					"listen": "prerequest",
+					"script": {
+						"id": "28fe7a36-50ab-4e7d-bf21-312b5f5777f3",
+						"exec": [
+							"// Wait between pings",
+							"setTimeout(function() {}, 60000);"
```
anniesullie-gov: The request has to be repeated until we get a 200 response, so this inserts a pause before making each request in this step. 60s worked on my macbook air to send few enough requests so that the server doesn't OOM, but I can imagine it could be too long to use as a default across test environments.

src/test/EndToEndRequestTest.postman_collection.json
```diff
@@ -0,0 +1,364 @@
...
+							"        pm.expect(pm.response.json().output[2].count).to.equal(400);",
+							"        pm.environment.set(\"coverage_url\", pm.response.json().output[2].url);",
+							"    })",
+							"} else {",
+							"    // If response code is not 200, it should be 202. Assert that, and retry.",
+							"    pm.test(\"Status code is 202\", function () {",
+							"        pm.response.to.have.status(202);",
+							"    });",
+							"    postman.setNextRequest(\"Job response\");",
+							"}"
```
anniesullie-gov: This logic both does some testing of the response, and decides whether to repeat the request or parse the response.

src/test/EndToEndRequestTest.postman_collection.json
```diff
@@ -0,0 +1,364 @@
...
+					"raw": "{{patient_url}}",
+					"host": [
+						"{{patient_url}}"
+					]
+				}
+			},
+			"response": []
+		},
+		{
+			"name": "Explanation of Benefit data",
```
anniesullie-gov: **PROBLEM**: receiving the 3152 item explanation of benefits response, even if I remove any logic to validate it in the exec field below, causes postman (both the gui and the command line) to hang until timeout/crash on my MacBook Air. The java version of the test appears to validate this response and runs just fine on the same machine, so I think this is a performance issue in postman.

I'm not sure how to address beyond just skipping making this request in the test.

src/test/EndToEndRequestTest.postman_collection.json
nickrobison-usds: I think that would make sense, is there a way to specify the relative path in the GUI itself?

I'm thinking of the use case where someone just opens our collection and wants to run the tests immediately.

src/test/EndToEndRequestTest.postman_collection.json
nickrobison-usds: Yup. The issue you're facing should be addressed with #101. So maybe something like 5-10 seconds would work.

src/test/EndToEndRequestTest.postman_collection.json
nickrobison-usds: Ok. This may have been addressed as well. We'll have to come back to this if it causes an issue on CI.

src/test/EndToEndRequestTest.postman_collection.json
anniesullie-gov: Collection->Edit->Pre-request scripts.

(There is also a "Variables" field in the collection, but it had some unexpected behavior which I couldn't find documented. It seems to sync the expected value of variables with a server, which is weird because we wouldn't want our CI tests interacting with a postman server for this. And it kept resetting the value to empty. Setting it in code was fully deterministic, so I did that instead).

src/test/EndToEndRequestTest.postman_collection.json
anniesullie-gov: Willdo.

src/test/EndToEndRequestTest.postman_collection.json
anniesullie-gov: I'm wondering if maybe it would be simpler to:
1. Significantly shorten the length of the roster
2. Inline it here instead of putting in a separate file

This would make the test easier to read, the requests/responses shorter so developers might be more encouraged to read them, solve the postman performance issue with large responses, and sidestep the problem here with relative paths.

The downside is that we wouldn't be solving this problem, just sidestepping it. I think deciding whether this is a good idea depends on what we intend to use postman for. This will work for small, self-documenting tests. It will not work for load tests. But after seeing the performance issues inside postman, I worry that it's not a good solution for load tests anyway.

src/test/EndToEndRequestTest.postman_collection.json
anniesullie-gov: willdo!

src/test/EndToEndRequestTest.postman_collection.json
```diff
@@ -0,0 +1,347 @@
...
...lue\": \"8D80925A-027E-43DD-8AED-9A501CC4CD91\"\n          }\n        ],\n        \"name\": [\n          {\n            \"family\": \"Provider\",\n            \"given\": [\n              \"Test\"\n            ]\n          }\n        ]\n      }\n    },\n    {\n      \"fullUrl\": \"http://something.gov/20000000001809\",\n      \"resource\": {\n        \"resourceType\": \"Patient\",\n        \"identifier\": [\n          {\n            \"value\": \"20000000001809\"\n          }\n        ],\n        \"name\": [\n          {\n            \"family\": \"Patient\",\n            \"given\": [\n              \"Tester -624899019\"\n            ]\n          }\n        ],\n        \"birthDate\": \"2019-03-01\"\n      }\n    },\n    {\n      \"fullUrl\": \"http://something.gov/20000000001810\",\n      \"resource\": {\n        \"resourceType\": \"Patient\",\n        \"identifier\": [\n          {\n            \"value\": \"20000000001810\"\n          }\n        ],\n        \"name\": [\n          {\n            \"family\": \"Patient\",\n            \"given\": [\n              \"Tester -1092912540\"\n            ]\n          }\n        ],\n        \"birthDate\": \"2019-03-01\"\n      }\n    }\n  ]\n}"
```
anniesullie-gov: This is the update where I shorten/inline the roster. It's kind of a bummer that the JSON formatting for postman doesn't really allow this to be multiple lines, so you can't easily read this file and understand the format. However when you import this file into the postman UI, you can see the formatted JSON, and everything runs correctly from commandline/ui with no issues around paths. What do you think?

anniesullie-gov:
Update: I added the test to the .travis.yml file, and it is passing in the last run (the failure is unrelated to my changes).

src/test/EndToEndRequestTest.postman_collection.json
nickrobison-usds: Makes sense to me!

PR #98, opened 2019-07-17
DPC-374: Initial commit of FHIR IG

nickrobison-usds:
**Why**

Since we're a FHIR project, we need to have a fancy FHIR implementation guide (IG) that serves the same purpose.

This PR adds the basic skeleton and build process, which was kind of tricky to figure out, as there's little to no documentation. Or good examples.

**What Changed**

Added the initial set of files required to build and publish the IG.

Created a Makefile with an `ig/publish` command that downloads and runs the publisher. We can eventually expand this file to help automate more of the build process itself.

**Choices Made**

Very little documentation at this point, we still need to figure out how to get the static assets into the web app and where we want the documentation source of truth.

As we finalize DPC-139, we'll keep adding new resources to the IG.

**Tickets closed**:

DPC-374

**Future Work**

We'll need to theme the IG to more closely follow our web design. We also need to link it into the public documentation pages.


ig/source/resources/implementationguide/ig-new.xml
```diff
@@ -0,0 +1,95 @@
...
+  <url value="https://dpc.cms.gov/ImplementationGuide/cms-dpc-ig"/>
+  <name value="DPCImplementationGuide"/>
+  <status value="draft"/>
+  <experimental value="false"/>
+  <publisher value="Centers for Medicare and Medicaid Services"/>
+  <description value="Draft implementation guide for DPC Application"/>
+  <packageId value="cms-dpc-ig"/>
+  <license value="CC0-1.0"/>
+  <fhirVersion value="3.0.1"/>
+  <fhirVersion value="3.0.1"/>
```
RickHawesUSDS: Dup value.

PR #106, opened 2019-07-18
Moved Travis Testing to enable Local Testing

ronakdev:
**Why**

The `travis.yml` includes some neat tests that would be useful to run locally before running the PR. For this reason, we moved much of the script section to a new file, `dpc-test.sh`, which can be run using the new `Makefile` target, `travis`.

**What Changed**

- Added new file `dpc-test.sh`
- Modified `travis.yml` to run `dpc-test.sh` instead of having a very long script section
- Added new `travis` target to the Makefile

The codeclimate section of `dpc-test` can be run by setting the environment variable `REPORT_COVERAGE`

**Choices Made**

**Tickets closed**:
Assigned by @nickrobison-usds 

**Future Work**

Compartmentalize the `dpc-test` script into functions so we can test specific parts using the `Makefile`


Makefile
```diff
@@ -7,4 +7,8 @@ ${IG_PUBLISHER}:
 .PHONY: ig/publish
 ig/publish: ${IG_PUBLISHER}
 	@echo "making IG"
-	@java -jar ${IG_PUBLISHER} -ig ig/ig.json
\ No newline at end of file
+	@java -jar ${IG_PUBLISHER} -ig ig/ig.json
+
+unexport REPORT_COVERAGE
```
nickrobison-usds: Can you explain the reason for using unexport, rather than the default variable syntax?

e.g.:
`REPORT_COVERAGE ?= true`

ronakdev: I used `unexport` to make sure that it doesn't run the CC test (the CC test only runs if REPORT_COVERAGE is set).

ronakdev: Switched to using `REPORT_COVERAGE ?= false`

PR #111, opened 2019-07-22
Add auth annotation tests

ronakdev:
**Why**

A large amount of the `dpc-api` methods did not have the appropriate annotations specifying the level of authentication provided.

**What Changed**
- Added `allResourcesHaveSecurityAnnotations()` test to `APIResourceAnnotationTest`
- Attempted to fix all the `v1` endpoints to make the test pass -- this needs review, I'm not sure if I added the correct annotations for everything, specifically the `PathAuthorizer` tests

**Choices Made**
**Tickets closed**
**Future Work**
**Checklist**

### DIFF
--- a/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/BaseResource.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/BaseResource.java
@@ -48,6 +48,7 @@ public class BaseResource extends AbstractBaseResource {
     }
 
     @Override
+    @Public
     @ApiOperation(value = "Get FHIR Metadata", notes = "Returns the FHIR Capabilities statement for the application", response = CapabilityStatement.class)
     public CapabilityStatement metadata() {
         return Capabilities.buildCapabilities();
--- a/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/BaseResource.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/BaseResource.java
@@ -1,5 +1,6 @@
 package gov.cms.dpc.api.resources.v1;
 
+import gov.cms.dpc.api.auth.annotations.Public;
 import gov.cms.dpc.api.core.Capabilities;
 import gov.cms.dpc.api.resources.*;
 import io.swagger.annotations.Api;
@@ -40,6 +41,7 @@ public class BaseResource extends AbstractBaseResource {
     }
 
     @Override
+    @Public
     @ApiOperation(value = "Return the software version", hidden = true)
     public String version() {
         return "Version 1";
--- a/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/BaseResource.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/BaseResource.java
@@ -8,6 +8,7 @@ import io.swagger.annotations.ApiOperation;
 import org.hl7.fhir.dstu3.model.CapabilityStatement;
 
 import javax.inject.Inject;
+import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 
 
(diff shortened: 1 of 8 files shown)


dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/BaseResource.java
```diff
@@ -40,6 +41,7 @@ public BaseResource(GroupResource gr,
     }
 
     @Override
+    @Public
```
nickrobison-usds: The `metadata()` method should require a Public annotation as well.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/DefinitionResource.java
```diff
@@ -45,6 +47,7 @@ public Bundle getStructureDefinitions() {
     @Override
     @GET
     @Path("/{definitionID}")
+    @PathAuthorizer(type = ResourceType.PractitionerRole, pathParam = "definitionId")
```
nickrobison-usds: This endpoint should be public

dpc-api/src/test/java/gov/cms/dpc/api/APIResourceAnnotationTest.java
```diff
@@ -54,6 +59,26 @@ void allResourcesHaveSwaggerAnnotations() {
         methods
                 .forEach(method ->
                         assertTrue(method.isAnnotationPresent(ApiOperation.class), String.format("Method: %s in Class: %s must have Swagger annotation", method.getName(), method.getDeclaringClass())));
-    }
+    }// iterate over parameters (at least one should have Auth)
+
+    @Test
+    void allResourcesHaveSecurityAnnotations() {
```
nickrobison-usds: Let's extract this into its own private, static method.

dpc-api/src/test/java/gov/cms/dpc/api/APIResourceAnnotationTest.java
```diff
@@ -54,6 +59,26 @@ void allResourcesHaveSwaggerAnnotations() {
...
+        methods.forEach(method -> {
+            boolean validParameters = false;
+            final Annotation[][] paramAnnotations = method.getParameterAnnotations();
+            for (Annotation[] annotations : paramAnnotations) {
+                for (Annotation an : annotations) {
+                    if(an.annotationType().equals(Auth.class)) {
+                        validParameters = true;
```
nickrobison-usds: One optimization would be to `break` as soon as you find a matching `@Auth` annotated parameter.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/DataResource.java
```diff
@@ -36,6 +38,7 @@ public DataResource(@ExportPath String exportPath) {
 
     @Override
     @Path("/{fileID}/")
+    @PathAuthorizer(type = ResourceType.PractitionerRole, pathParam = "fileID")
```
nickrobison-usds: Using `PractitionerRole` is fine for now, because we don't have our authorization wired up with the rest of the system, yet.

We'll need to make some changes to this endpoint in the future, but this is fine for now.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/BaseResource.java
ronakdev: Fixed this issue, also added all methods annotated with `ApiOperation` to the testing method — not sure if there are any other annotations that need to be checked for auth.

ronakdev: Made a mistake — ApiOperation is just Swagger stuff, instead realized that `metadata()` was just not annotated properly. Added `@GET` annotation to it

PR #112, opened 2019-07-22
DPC-428: Add Patient Resource Endpoint

nickrobison-usds:
**Why**

In order to support our new attribution logic, we need the ability to add `Patients` to an organization, in order to be able to attribute them to individual providers.

This also fixes DPC-429 in provider records weren't being reconciled correctly and causing duplicated attribution relationship.s

**What Changed**

Added a `Patient` endpoint, which allows organizations to create/update/delete patients from their organizations. These individual records are _duplicated_ in the sense that a given beneficiary might be associated to multiple organizations, each each org will have a record.

All resources are now tied to Organizations, so we need to ensure that at least one Organization is present in the database before running any integration tests. All of the tests have been updated to ensure that's the case, but something to be aware of going forward.

Removed the `PractitionerRole` resource because it's no longer needed, `Practitioner` resources now have a link to their parent organization directly in the DB table. In order for this to work, we need to add a `Tag` to the `Practitioner` `Meta` block, which contains the organization resource ID.

**Choices Made**

FHIR documentation will come in another PR, since this one is already ginormous.

**Tickets closed**:

DPC-428: This ticket
DPC-429: Fix duplicate attribution tests

**Future Work**

DPC-433: Improve Path Authorizer to account for missing resources
DPC-432: Improve ID parsing logic.

**Checklist**

- [X] Swagger documentation has been updated
- [X] FHIR documentation has been updated


dpc-web/.rakeTasks
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ronaldheft-usds: This is a generated IntelliJ file. .gitignore?

nickrobison-usds: Gah! Missed it. thanks!

PR #123, opened 2019-07-24
Modified Jobs to Require The Right Organization

ronakdev:
**Why**

Currently, anyone can get the job `export` data if they have the job id. This means we aren't authenticating them on checking the export data. So, we changed it so that we check the organization the request is coming from to see if there's a match

**What Changed**

_Core Changes_
- Modified `gov.cms.dpc.queue.models.JobModel.java` to include a new property `orgID`
- Modified `AbstractGroupResource.java` and `GroupResource` 's `export` method to include a required `OrganizationPrincipal` param
- Modified `AbstractJobResource.java` and `JobResource` 's `checkJobStatus` method to include a required `OrganizationPrincipal` param
- Added a new column `organization_id` to `queue_migrations.xml`

_Test Changes_
- Fixed Aggregation Tests to work with the new `JobModel` (uses `UUID.randomUUID` for orgID)
- Modified `QueueTest` and `JobModelTest` to work with the new `JobModel` (uses `UUID.randomUUID` for orgID)
- Added a new static method `makeOrganizationPrincipal` to `APITestHelpers` for use in `dpc-api` tests
- Fixed `FHIRSubmissionTest` and `JobResourceTest` and used `makeOrganizationPrincipal` for orgID
- @nickrobison-usds changed the tests to use a local configuration thing to in `FHIRSubmissionTest` to make it so that we use the default organization when we do our tests

**Choices Made**
- Decided to use `UUID.randomUUID` sometimes instead of making a full proper mock `OrganizationPrincipal` to avoid importing `dpc-api` into `dpc-aggregation`

**Tickets closed**:
Assigned by @nickrobison-usds 

**Future Work**

More robust tests regarding the newly added check for organization. I'm thinking we could create some jobs with different organizations and try to get data. We could also use the same provider (but with different organizations) and make sure that it only works with the right organization.

**Checklist**


dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
```diff
@@ -63,8 +65,9 @@ public JobResource(JobQueue queue, @APIV1 String baseURL) {
             @ApiResponse(code = 500, message = "Export job has failed with no results"),
             @ApiResponse(code = 200, message = "Export job has completed. Any failures are listed in the response body", response = JobCompletionModel.class)
```
nickrobison-usds: Let's add a 404 response to our Swagger docs.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
```diff
@@ -73,6 +76,9 @@ public Response checkJobStatus(@PathParam("jobID") String jobID) {
             if (!job.isValid()) {
                 throw new JobQueueFailure(jobUUID, "Fetched an invalid job model");
             }
+            if (!job.getOrgID().equals(orgUUID)) {
```
nickrobison-usds: Nitpicky tweak, let's move this above the isValid() check. Just to save a couple of cycles.

dpc-attribution/src/main/resources/queue_migrations.xml
```diff
@@ -36,4 +36,9 @@
                                  referencedColumnNames="jobid"/>
     </changeSet>
 
+    <changeSet id="add-job-id" author="ronakdev">
```
nickrobison-usds: Let's change this to `add-organization-id`

ronakdev:
> One question. Is there anywhere in the test suite that we're checking if the 401 is actually thrown?
> 
> It looks like we need to update the tests to cover failure cases as well.

Added to [`testWrongOrgJobAccess()` in `JobResourceTest`](https://github.com/CMSgov/dpc-app/pull/123/files#diff-b73e85529d866d1d5015b9fa33599aa4R195)

PR #125, opened 2019-07-25
DPC-173: Added Practitioner Profile and updated FHIR IG

nickrobison-usds:
**Why**

Given that we're now requiring Organizations to submit Practitioners as part of their rostering process, we need to ensure that they're actually given the required information.

This not only adds the `Practitioner` StructureDefinition, but also updates the others and added some basic content to the FHIR IG.

**What Changed**

All of our major resource types are now profiled:

- Organization
- Endpoint
- Patient
- Practitioner

They're loaded by default into our custom Verification module and we've added some basic unit tests to ensure the profiles are correct and valid.

Also added the basic data requirements to the FHIR IG and cleaned up a couple of rough edges. The README has been updated with more details as to how to add custom resource types to the IG, it's not particularly clear or easy.

**Choices Made**

The IG builds, but it's not wired into the  website build yet, that will come in a later PR.

**Tickets closed**:
DPC-174
DPC-173
DPC-255

**Future Work**
DPC-446: Integrate FHIR IG with DPC Website
DPC-448: Add updated timestamp to Providers table, in order to populate meta fields.
More style cleanups, now that we actually have content.
Wording review.

**Checklist**


dpc-common/src/main/java/gov/cms/dpc/fhir/converters/entities/ProviderEntityConverter.java
```diff
@@ -23,7 +24,9 @@ public static Practitioner convert(ProviderEntity entity) {
 
         practitioner.addIdentifier().setValue(entity.getProviderNPI());
         final Meta meta = new Meta();
+        // TODO: This is incorrect fix it.
```
ronaldheft-usds: Do we have an issue logged for this? Is the last updated field indicating when the spec for the meta was last updated or any of the meta data?

nickrobison-usds: Good catch. DPC-448 has been created to track it.

PR #128, opened 2019-07-25
MERGE 7/30 ONLY!!!! allow site to be indexed by search engines

jaredcunha-usds:
This will remove the meta tag that prevents search engines from indexing the site.


ronaldheft-usds:
Approved by @nickrobison-usds for merging.

PR #131, opened 2019-07-25
Jaysheth/attribution size

jaysheth1998:
**Why**

There needed to be way to add an attribution size limit.

**What Changed**
Added in a statement that compared the attribution IDs to the provider IDs. An exception will be thrown if the limit is exceeded.

**Choices Made**

**Tickets closed**:

**Future Work**
The function needs to be made configurable so that the limit can be modified accordingly.

**Checklist**


dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/RosterUtils.java
```diff
@@ -66,10 +68,20 @@ public static void submitAttributionBundle(Bundle attributionBundle, DSLContext
                 .peek(patient -> patient.setManagingOrganization(new Reference("Organization/" + organizationID.toString())))
                 .map(PatientEntity::fromFHIR)
                 .forEach(patientEntity -> RosterUtils.createUpdateAttributionRelationship(ctx, patientEntity, providerRecord, creationTimestamp));
+
+        int attributionListSize = ctx
+                .selectCount()
+                .from(Attributions.ATTRIBUTIONS)
+                .where(Attributions.ATTRIBUTIONS.PROVIDER_ID.eq(providerRecord.getId()))
```
jeskew-gov: This is a postgres noob question: do we need to add a separate index on the provider ID field to support this query?

PR #147, opened 2019-07-27
Setup Matomo instance

ronakdev:
**Why**

Adds analytics to `dpc-web`

**What Changed**
- Added `matomo_app` and `matomo_db` services to `dpc-web`'s `docker-compose.yml`
- Added the Matomo tracking tag to `views/layouts/application.html.erb`

**Choices Made**
- chose not to use the [Official Gem](https://rubygems.org/gems/piwik_analytics) since it hasn't been updated since 2012
- Currently using port 80 for Matomo to run --> I need to modify matomo's settings to make this work with a different port, but wasn't sure how we're planning on running this matomo instance (if we're doing it on a completely different server there's no need to change the port we're using)
- Passwords are all set to be `password` in plaintext --> I need to change this but I'm not sure how we're doing passwords
- Using a brand new database instead of using the existing one so we can run matomo on a seperate server with no problems

**Tickets closed**:
Assigned by @nickrobison-usds 

**Future Work**
- Fixing password management
- Reconfiguring where Matomo is run (different port or different server?)

**Checklist**


ronakdev:
Shelving this in favor of Google Analytics, see #151

PR #150, opened 2019-07-29
Configurations for test and prod-sbx

RickHawesUSDS:
**Why**

Add the configurations for prod-sbx and test. These are needed to run in these CCS VPCs. 

**What Changed**

- New configs

**Choices Made**

**Tickets closed**:

**Future Work**

**Checklist**


nickrobison-usds:
@RickHawesUSDS One more change. Addressing the root url issue, adding the following to the `server` block of the API configuration, sets up the base context path, which should be all we need.

`applicationContextPath: "/fhir"`. I'd like to use the `/fhir` path if possible, rather than API.

RickHawesUSDS:
Merging now to get things deployed

PR #154, opened 2019-07-29
DPC-469: Pass auth token through Demo command

nickrobison-usds:
**Why**

We need the newly generated auth token, and we need to pass it along through the demo command.

**What Changed**

Updated the `exportClient` commands to take the auth token as a parameter.

**Choices Made**

**Tickets closed**:
DPC-469

**Future Work**

**Checklist**


nickrobison-usds:
Done as part of #186

PR #167, opened 2019-07-30
Potentially Fixed IG Analytics [Needs Proper Testing]

ronakdev:
**Why**

Analytics were breaking IG 
**What Changed**

Describe the changes made to the code base including new features, refactoring of existing services and user visible API changes.

- moved analytics code to where the other javascript was being imported
- used older version of google analytics' tracking code that doesn't rely on an async external script import

**Choices Made**

**Tickets closed**:

**Future Work**
- I'm unfamilar with ig and I wasn't able to figure out where the generated html was being put — so I'm not 100% sure this will work
- I'm not sure what domain this will be on; if it is a different domain than `dpc.cms.gov`, we'll need to create a new thing in analytics (i can do this)

**Checklist**

### DIFF
--- a/ig/source/resources/implementationguide/ig-new.json
+++ b/ig/source/resources/implementationguide/ig-new.json
@@ -7,23 +7,23 @@
   },
   "extension" : [{
     "url" : "http://hl7.org/fhir/StructureDefinition/igpublisher-folder-resource",
-    "valueString" : "/Users/jaredcunha/work/dpc-app/ig/source/resources"
+    "valueString" : "/Users/ronakshah/Developer/HHS/dpc-app/ig/source/resources"
   },
   {
     "url" : "http://hl7.org/fhir/StructureDefinition/igpublisher-folder-resource",
-    "valueString" : "/Users/jaredcunha/work/dpc-app/dpc-common/src/main/resources/validations"
+    "valueString" : "/Users/ronakshah/Developer/HHS/dpc-app/dpc-common/src/main/resources/validations"
   },
   {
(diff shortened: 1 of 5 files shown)


ig/source/resources/implementationguide/ig-new.json
```diff
@@ -7,23 +7,23 @@
   },
   "extension" : [{
     "url" : "http://hl7.org/fhir/StructureDefinition/igpublisher-folder-resource",
-    "valueString" : "/Users/jaredcunha/work/dpc-app/ig/source/resources"
+    "valueString" : "/Users/ronakshah/Developer/HHS/dpc-app/ig/source/resources"
```
ronaldheft-usds: @ronakdev @nickrobison-usds Are these files that should be added to .gitignore? They look like they have generated information in them?

ronakdev: Good catch — maybe are those added when you run the `make` command?

nickrobison-usds: They should be. I thought I git rm-ed them, but maybe not. @ronakdev Can you nuke these files before we merge? Thanks!

ronakdev: Nuke the resources folder? Or remove the `valueString` property from `ig-new.json`?

ronaldheft-usds: @ronakdev I assume remove the `ig-new.*` files from the Git history and add them to .gitignore.

PR #168, opened 2019-07-30
update privacy info on faq page

jaredcunha-usds:
**Why**

Text changes from OCR


jaredcunha-usds:
This is just text, so I'm going to merge it in.

PR #173, opened 2019-08-01
Add number of providers to Users: Addresses DPC-495

mphprogrammer:
**Why**
It would be a benefit to know company sizes for registrations.

**What Changed**
Added migration to add num_providers to the User model. Whitelisted num_providers, added specs, and inserted column into views.

### DIFF
--- a/dpc-web/app/views/devise/registrations/new.html.erb
+++ b/dpc-web/app/views/devise/registrations/new.html.erb
@@ -54,6 +54,11 @@
         </div>
 
         <div class="field">
+          <%= f.label :num_providers, "Number of Providers at Your Organization", class: "ds-c-label" %>
+          <%= f.number_field :num_providers, value: 0, min: 0, class: "ds-c-field" %>
+        </div>
+
+        <div class="field">
           <%= f.label :address_1, class: "ds-c-label" %>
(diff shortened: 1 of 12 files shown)


dpc-web/app/views/devise/registrations/new.html.erb
```diff
@@ -53,6 +53,11 @@
           %>
         </div>
 
+        <div class="field">
+          <%= f.label :num_providers, "Number of Providers at Your Organization", class: "ds-c-label" %>
+          <%= f.number_field :num_providers, value: 0, min: 0, class: "ds-c-field" %>
```
nickrobison-usds: Can we just leave this blank, if they haven't entered anything yet?

jaredcunha-usds:
To resolve some of the potential confusion on behalf of Vendors, I've made the question about the number of providers to serve as a follow-up depending on the response to the organization type question. This is done on `registrations#new` and `registrations#edit`.

Screen Recording 2019-08-05 at 01 36 PM

Now, on the profile page, the number of providers data is hidden if the user's organization is a Health IT Vendor.

jaredcunha-usds:
@nickrobison-usds @mphprogrammer I think this is ready to go on my end.

PR #180, opened 2019-08-05
DPC-293: Refactor attribution

nickrobison-usds:
**Why**

Our current implementation of the attribution logic was really hacky and gross. We essentially submitted a Bundle to the Group endpoint (which is not allowed) and required that the Bundle be sliced a certain way. This PR refactors everything to use the `Group` based attribution logic described [here](https://github.com/smart-on-fhir/smart-on-fhir.github.io/wiki/Bulk-data:-thoughts-on-attribution-lists-and-groups). This also greatly simplifies our implementation and testing logic as we're now native FHIR resources instead of creating our own to achieve the same results.

**What Changed**

Patient and Practitioner endpoints now have `$submit` operations, which let us post Bundles of resources to the related endpoints. This should dramatically simplify on boarding processes, as we can now do things in one fell swoop.

The Group endpoint now has a `POST` command, which takes a Group with references to the newly created Patients and Providers. We then create a `RosterEntity` which holds all the attributions and allows us to cascade deletions and updates.

The test setup and seed commands have been simplified to simply work with resources, instead of just bundles. There should be a lot of reduction in the testing space and helper methods as a bunch of shenanigans are no longer required. We may even be able to get rid of Jooq now, which would improve our handling of DB resources.

**Choices Made**

Testing the Group functionality for the API service is done in the Postman tests, @ronakdev is working on a PR to get Code Coverage numbers from our Docker instances.

**Tickets closed**:

DPC-293: Refactor attribution logic

**Future Work**

DPC-490: Tune DB queries.
DPC-261: Add Attribution Group profile
DPC-532: Re-enable group profile validation

**Checklist**


nickrobison-usds:
@ronaldheft-usds @RickHawesUSDS This PR is ready for review, I'm still finalizing the documentation pieces, but the core logic should be working.

dpc-api/src/main/java/gov/cms/dpc/api/DPCAPIModule.java
```diff
@@ -113,6 +92,7 @@ public String provideV1URL(@ServiceBaseURL String baseURL) {
     @Provides
     @Singleton
     public IGenericClient provideFHIRClient(FhirContext ctx) {
+        logger.debug("Connecting to attribution server at {}.", getConfiguration().getAttributionURL());
```
RickHawesUSDS: I like this practice. I suggest that it could be an INFO level event.

PR #182, opened 2019-08-06
Integrated Jacoco into Docker

ronakdev:
**Why**

We miss all of the coverage numbers when run the End to End tests under Docker. Thus, we need to have Jacoco run with docker at runtime

**What Changed**

*Biggest Change*
In order to include the jacoco jar and the docker entrypoint scripts, I upgraded our version of JIB from `1.0.0` to `1.4.0`. I've run all the local demo commands and it seems to be working, but that's something to check more to make sure this upgrade didn't break anything.

*Other Changes*
- made `api` and `aggregation` use `entrypoint` scripts (make sure they are working the way they need to
- Added a `jacocoReport/` volume to docker-compose that stores all api/agg/att 's `jacoco.exec` outputs

**Choices Made**

- Upgraded to Jib 1.4, not sure if we were using any 1.0 only features (i think it should be fine, no syntax errors occurred)

**Tickets closed**:
**Future Work**
- Compile the html report instead
  - @nickrobison-usds I think it might be best to talk in person about using the maven `jacoco:report-aggregate` goal with Docker, not sure how to do this
- Integrate with CodeClimate

**Checklist**

- [X] Demo and Seed commands are working
- [X] Swagger documentation has been updated
- [X] FHIR documentation has been updated


ronakdev:
Additionally, I have not committed the `jacocoReport/` folder to git, not sure if this should just be in the `.gitignore` or if we should be committing it for codecoverage

dpc-aggregation/docker/entrypoint.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -e
+
+JACOCO="-javaagent:/org.jacoco.agent-runtime.jar=destfile=/jacoco-report/jacoco-it.exec"
+
+CMDLINE="java ${JACOCO} -cp /app/resources:/app/classes:/app/libs/* gov.cms.dpc.aggregation.DPCAggregationService"
+
+
+echo "Running server via entrypoint!"
+exec ${CMDLINE} "$@"
```
nickrobison-usds: We need to make sure that we're not running Jacoco on each invocation. So we'll need to pass and check an environment variable.

dpc-attribution/docker/entrypoint.sh
```diff
@@ -2,7 +2,9 @@
 
 set -e
 
-CMDLINE="java -cp /app/resources:/app/classes:/app/libs/* gov.cms.dpc.attribution.DPCAttributionService"
+JACOCO="-javaagent:/org.jacoco.agent-runtime.jar=destfile=/jacoco-report/jacoco-it.exec"
+
+CMDLINE="java ${JACOCO} -cp /app/resources:/app/classes:/app/libs/* gov.cms.dpc.attribution.DPCAttributionService"
 
 run_java() {
```
nickrobison-usds: Let's remove this unused method.

pom.xml
```diff
@@ -436,6 +444,24 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
```
nickrobison-usds: Let's move this plugin invocation into the `integration-tests` profile and rename it to `ci`. That way we only bundle the extra jar during testing.

nickrobison-usds:
@ronakdev Looks good to me! Can we update the test bash file to pull in the newly generated jacoco reports?

PR #186, opened 2019-08-09
DPC-534: Refactor demo command

nickrobison-usds:
**Why**

The demo command had not been changed in a while, so it wasn't working. It was also really brittle not very useful, so we refactored it to make it a bit more robust, self contained, and useable for load testing.

**What Changed**

The Demo Command now creates a brand new Organization on each invocation, does all the rostering and then kicks off a single job per Provider. Which means we do 3 jobs for a total of 185 patients.

Cleaned up a couple of small bugs with the way we're handling Operation parameters, we can't just post a bundle, we have to wrap it in a Parameters object, because FHIR.

**Choices Made**

**Tickets closed**:

DPC-534: Demo command refactor
DPC-373: Bulk export client now does authentication correctly.
DPC-469: Demo command is failing in AWS.

**Future Work**

DPC-536: Fix slicing in FHIR profiles.
DPC-537: Make the expensive validator not load lazily.

**Checklist**


nickrobison-usds:
@RickHawesUSDS This is ready for review. Thanks!

PR #187, opened 2019-08-09
Update prodsbx to connected to peered BlueButton

ronaldheft-usds:
**Why**

Connected to the newly peered BlueButton URL.

### DIFF
--- a/dpc-aggregation/src/main/resources/prod-sbx.application.conf
+++ b/dpc-aggregation/src/main/resources/prod-sbx.application.conf
@@ -7,6 +7,7 @@ dpc.aggregation {
     password = dpc-safe
   }
   bbclient.keyStore.location = "/bb.keystore"
+  bbclient.serverBaseUrl = "https://internal-dpcwelb01-2074070868.us-east-1.elb.amazonaws.com/v1/fhir/"
   queue {
     singleServerConfig {
       address = "redis://redis.dpc-prod-sbx.local:6379"
--- a/dpc-aggregation/src/main/resources/test.application.conf
+++ b/dpc-aggregation/src/main/resources/test.application.conf
@@ -7,6 +7,7 @@ dpc.aggregation {
     password = dpc-safe
   }
   bbclient.keyStore.location = "/bb.keystore"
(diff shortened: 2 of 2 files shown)


dpc-aggregation/src/main/resources/prod-sbx.application.conf
```diff
@@ -7,6 +7,7 @@ dpc.aggregation {
     password = dpc-safe
   }
   bbclient.keyStore.location = "/bb.keystore"
+  bbclient.serverBaseUrl = "https://internal-dpcwelb01-2074070868.us-east-1.elb.amazonaws.com/v1/fhir/"
```
nickrobison-usds: This should just be `https://internal-dpcwelb01-2074070868.us-east-1.elb.amazonaws.com`

ronaldheft-usds: @nickrobison-usds serverBaseUrl in `application.conf` includes `/v1/fhir/`. Is that incorrect as well?

nickrobison-usds: Honestly, not entirely sure, I think you're right though. Let's try it with the /v1/fhir

ronaldheft-usds:
@nickrobison-usds Will be doing deploys to test later today to continue the work on the Jenkins deployment script. I can run a deploy with the peered URL in test. Should this also be set in test? Right now only the peered URL is used in prod-sbx.

nickrobison-usds:
@ronaldheft-usds Let's set this up in test and validate there. I checked the peering was working this morning.

ronaldheft-usds:
We're still having connectively issue in `test` and `prod-sbx`. Will be investigating more tomorrow.

RickHawesUSDS:
DPC is not on-boarded into BFD. You'll need to hit the new CCS environment when you do. The domain will be `dpc.bfdcloud.net`.

ronaldheft-usds:
This PR is no longer valid. We reconfigured things to accept a parameter via Docker for the BFD backend URL.

PR #189, opened 2019-08-09
Fix base url generation

nickrobison-usds:
Our siteURL provider now pulls the context from the config and builds that into the path.

The Demo command now takes a full url, rather than doing the funky appends, which should help make it more portable between environments.

### DIFF
--- a/dpc-api/src/main/java/gov/cms/dpc/api/cli/DemoCommand.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/cli/DemoCommand.java
@@ -52,8 +52,8 @@ public class DemoCommand extends Command {
                 .addArgument("--host")
                 .dest("hostname")
                 .type(String.class)
-                .setDefault("localhost:3002")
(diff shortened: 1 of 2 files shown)


dpc-api/src/main/java/gov/cms/dpc/api/cli/DemoCommand.java
```diff
@@ -52,8 +52,8 @@ public void configure(Subparser subparser) {
                 .addArgument("--host")
                 .dest("hostname")
                 .type(String.class)
-                .setDefault("localhost:3002")
-                .help("Set the hostname (including port number) for running the Demo against");
+                .setDefault("http://localhost:3002/v1")
```
ronaldheft-usds: Isn't our default path `/api/v1/` and not just `/v1/`?

nickrobison-usds: It is in production and sandbox, but not local yet.

PR #198, opened 2019-08-15
DPC-256: JMeter load test

nickrobison-usds:
**Why**

We need a way to run load and smoke tests against our AWS environments. This PR adds a really basic JMeter script which should accomplish this.

**What Changed**

Added a `dpc-smoketest` module, which contains the necessary classes for building a custom `JavaRequestSampler` which is how we can wrap our custom Java logic into something that JMeter can understand.

Also added `smoke/test` and `smoke/sbx` targets to the Makefile, which should make it nice and easy to run tests after deploys.

**Choices Made**

Everything is in a single class right now, which was easy to get working, but doesn't really give us deep insights into which parts of the application are slow, or how long individual requests take. We may want to improve that in the future, but JMeter makes it really hard to have nested Samples, so we we'll have to break things into individual classes and juggle the state.

JMeter tests are a little brittle, paths are relative to the application root and we can't easily override specific values (like iteration count) due to the way variables are passed. It's possible, but clunky.

We're currently running a single test iteration of _n_ number of threads. We default to 5, or it can be configured based on the `SMOKE_THREADS` param in the Makefile.

**Tickets closed**:

DPC-256

**Future Work**

**Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ smoke/test: ${JMETER}
 	@echo "Running Smoke Tests against Test env"
 	@${JMETER} -p src/main/resources/test.properties -Jthreads=${SMOKE_THREADS} -n -t src/main/resources/SmokeTest.jmx -l out.jtl
 
-.PHONY: smoke/sbx
-smoke/sbx: ${JMETER}
+.PHONY: smoke/prod-sbx
+smoke/prod-sbx: ${JMETER}
 	@echo "Running Smoke Tests against Sandbox env"
 	@${JMETER} -p src/main/resources/sbx.properties -Jthreads=${SMOKE_THREADS} -n -t src/main/resources/SmokeTest.jmx -l out.jtl
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 IG_PUBLISHER = ./.bin/org.hl7.fhir.publisher.jar
-JMETER = ./.bin/jmeter/bin/jmeter
 REPORT_COVERAGE ?= false
+
+JMETER = ./.bin/jmeter/bin/jmeter
+SMOKE_THREADS ?= 10
 
 ${IG_PUBLISHER}:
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -16,4 +27,14 @@ travis:
...
\ No newline at end of file
+	@docker build -f dpc-web/Dockerfile .
+
+.PHONY: smoke/test
+smoke/test: ${JMETER}
+	@echo "Running Smoke Tests against Test env"
+	@${JMETER} -p src/main/resources/test.properties -Jthreads=${SMOKE_THREADS} -n -t src/main/resources/SmokeTest.jmx -l out.jtl
+
+.PHONY: smoke/sbx
+smoke/sbx: ${JMETER}
```
ronaldheft-usds: For consistent naming and easy of parameterizing our Jenkins jobs, can this be renamed `smoke/prod-sbx`?

Makefile
```diff
@@ -16,4 +27,14 @@ travis:
...
+
+.PHONY: smoke/test
+smoke/test: ${JMETER}
+	@echo "Running Smoke Tests against Test env"
+	@${JMETER} -p src/main/resources/test.properties -Jthreads=${SMOKE_THREADS} -n -t src/main/resources/SmokeTest.jmx -l out.jtl
+
+.PHONY: smoke/prod-sbx
+smoke/prod-sbx: ${JMETER}
+	@echo "Running Smoke Tests against Sandbox env"
+	@${JMETER} -p src/main/resources/sbx.properties -Jthreads=${SMOKE_THREADS} -n -t src/main/resources/SmokeTest.jmx -l out.jtl
```
ronaldheft-usds: @nickrobison-usds I think this target will fail. Doesn't seem like there is a `sbx.properties` file. Should be `prod-sbx.properties`. The echo statement should be updated as well to reflect the environment.

PR #201, opened 2019-08-21
DPC-558: Fix testing race condition

ronaldheft-usds:
**Why**

We have a race condition during testing where the db sometimes will not be started when the attribution / aggregation sever starts. The docker container would exit and the attribution / aggregation server would not be started, causing the API tests to fail.

**What Changed**

This PR includes the DB and Redis server in our `start_dependencies` and `start_api` targets to ensure the ports are ready for traffic before starting any server that depends on them. This pattern was introduced in #199 but did not include the DB and Redis servers in the dependencies.

**Choices Made**

**Tickets closed**:

**Future Work**

**Checklist**

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -57,24 +57,25 @@ services:
       - export-volume:/app/data
       - ./jacocoReport/dpc-api:/jacoco-report
 
-  start_dependencies:
+  start_core_dependencies:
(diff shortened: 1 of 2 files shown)


ronaldheft-usds:
This has been testing in Jenkins and has resolved the race condition issue.

PR #214, opened 2019-09-03
Introduce InternalUser model and functionality

switzersc-usds:
**Why**

We need the ability for internal CMS users to log in and perform internal admin functions. 

**What Changed**

This PR introduces an InternalUser model that also uses Devise for authentication and `internal` namespace for internal user routes. There are associated views, routes, and controllers for auth and features to be accessible by InternalUsers. 

The internal login page, which does not have a Sign Up link:
image

Once an internal user has logged in, they see a Pending page for now:
image

**Choices Made**

I've created a separate model because there is no feature overlap between a user and an internal CMS user and we may even want different auth rules for each one (e.g. requiring a .gov email for internal CMS users). This should help us keep a clear distinction between types of users and also enable us to delineate different role structures for users (e.g. org admins, devs, etc) and internal users (e.g. CMS owner, tech support, etc).

**Future Work**

* Add internal user pages 

**Checklist**

~- [ ] Demo and Seed commands are working~
~- [ ] Swagger documentation has been updated~
~- [ ] FHIR documentation has been updated~


dpc-web/app/controllers/concerns/multi_model_login_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+module MultiModelLoginHelper
+  extend ActiveSupport::Concern
+  included do
+    before_action :check_user
+  end
+
+  protected
+
+  def check_user
+    # binding.pry
```
nickrobison-usds: Is this left here for a reason?

dpc-web/app/controllers/internal/devise/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Internal
+  module Devise
+    class OmniauthCallbacksController < Devise::OmniauthCallbacksController
```
nickrobison-usds: If we're using omni auth, is it possible to loop in Github logins?

dpc-web/app/models/internal_user.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class InternalUser < ApplicationRecord
```
nickrobison-usds: Are you planning on adding admin user details in a later PR?

dpc-web/app/views/internal/devise/registrations/new.html.erb
```diff
@@ -0,0 +1,29 @@
+<h2>Sign up</h2>
```
nickrobison-usds: Is this necessary if Internal users are not registerable?

switzersc-usds: Nope! Left over from the auto generation. Will remove.

dpc-web/app/models/internal_user.rb
switzersc-usds: Yeah, as needed. Right now I don't know what kind of attributes we need for internal users other than name and email. Unless you're referring to other types of user details?

dpc-web/app/controllers/internal/devise/omniauth_callbacks_controller.rb
switzersc-usds: We can do this. I left the omniauth controller here (from auto generation) as a basis for that work but thought this PR might grow too large if I included that functionality. Will our contractors be able to log in via Github? If not, then we probably can't do this, at least not for all internal users.

dpc-web/app/controllers/concerns/multi_model_login_helper.rb
switzersc-usds: Nope! Thanks for the catch =]

switzersc-usds:
@nickrobison-usds responses:

> Can we rely on Github authentication and user management? Remove the need to h�ave internal users setup a password and delegate the actual access management to Github, like we do with Jenkins?
See inline comment. 

> What's the differentiation between an internal user and an actual user, in terms of the login screen. Do they both see the same screen and then based on which table they're in get routed accordingly?
The login screenshot for internal users is above in the PR description. For external users, it's basically the same but with a "Request access" link:
image

Once they sign in, external users see their profile / DPC registration page, while internal users currently see a blank page that says "Pending" (also pictured above in the PR description). Once the users/orgs management work is done, internal users will probably see the Users table as the first page after login.

dpc-web/app/controllers/internal/devise/omniauth_callbacks_controller.rb
switzersc-usds: Discussed IRL and will add Github auth in next PR.

dpc-web/app/models/internal_user.rb
switzersc-usds: Discussed IRL.

PR #218, opened 2019-09-04
DPC-499: Queue V2

ronaldheft-usds:
**Why**

We'd like to redesign our queue to support faster processing of jobs split across multiple aggregators, reporting the progress of the jobs, the ability to shut down the aggregator cleanly, and support Patient Everything (processing single patient requests first).

The overall design and goals of Queue V2 has been documented in Confluence here: https://confluenceent.cms.gov/x/gQNCAg

**What Changed**

User-Facing Changes:
- The `X-Progress` header now reports on the actual status of the migration in the format `<STATUS>: <PERCENT>%` where PERCENT = X.XX format. This has been added to the documentation. The FHIR spec does not specify a format, but I am open to suggestions if there is a different way we want to present this.
- The file names of the job will be in the following format: `<batchID|uuid>-<sequence-count>.<resource-type>`
  - Previously the file format was using the jobID instead.
  - This will result in lots of different files, based on the batch count (currently 100). Since the results do not need to be ordered, this is okay.
- All previous jobs will no longer be retrievable.
  - They are stored in the old `job_queue` table.
  - If we'd like to ensure old jobs are still accessible, additional work is needed. Since we have not launched outside of synthetic data yet, this tradeoff appears to be acceptable.

Application Changes:
- Lots of refactoring to support Queue V2.
- The aggregator service now handles the queue database creation and migrations.
- Script added to our docker postgres setup to bootstrap multiple databases.
- Removed redis from our docker-compose and all application dependencies.
- Removes the old logic for file encryption.

**Choices Made**

Queue V2 was designed based off the design notes [in this Confluence document](https://confluenceent.cms.gov/x/gQNCAg). Along the way several other design decisions were made to make the original decision a reality.

- A separate database was create for the queue. This brought forth several bugs and issues with how our database management had been done with Dropwizard.
  - Migrating the database in the application `run` method was occurring after the queue started. This would cause the queue to crash. Migrations are now done in the docker `entrypoint.sh`.
  - The Hibernate Dropwizard module was refactored to fix an issue with duplicate `SessionFactory` creation.
  - The `@AdditionalProviders` functionality was refactored as a constructor to parameter, since it needed to be available before Guice started to prevent duplicate `SessionFactory` creation.

**Tickets closed**:

- DPC-499: This ticket for Queue V2
- DPC-190: Remove redis dependency from queue
- DPC-212: Remove pair class
- DPC-465: Test that was failing due to Queue v1 design
- DPC-642: Fix missing ndjson extension

**Future Work**

- DPC-639: Support multiple databases in Terraform **(blocker for deployment)** (Ready in https://github.com/CMSgov/dpc-ops/pull/45)
- DPC-603: Remove the old queue tables (ready in #244)
- DPC-640: Run multiple queue batches on a single aggregator
- DPC-641: Support FHIR export delete
- Remove Redis server from AWS?
- We will need to communicate the breaking changes on our Google Group. See: user-facing changes section.

**Checklist**


dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/DPCAggregationConfiguration.java
```diff
@@ -43,23 +46,16 @@
     @Max(100000) // Keep files under a GB
     private int resourcesPerFileCount = 10000;
 
-    // Enable file encryption per BCDA
```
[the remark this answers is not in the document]
nickrobison-usds: Did we remove parallel and encryption options?

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/DPCAggregationService.java
```diff
@@ -26,14 +31,26 @@ public String getName() {
...
         GuiceBundle<DPCAggregationConfiguration> guiceBundle = GuiceBundle.defaultBuilder(DPCAggregationConfiguration.class)
-                .modules(new DPCHibernateModule<>(),
-                        new AggregationAppModule(),
+                .modules(new AggregationAppModule(),
+                        new DPCQueueHibernateModule<>(hibernateBundle),
                         new JobQueueModule<>(),
                         new BlueButtonClientModule<>())
                 .build();
 
+        bootstrap.addBundle(hibernateBundle); // Needs to be before guice
```
nickrobison-usds: What's the reason for this? Can we expand the comment a bit?

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/DPCAggregationService.java
```diff
@@ -26,14 +31,26 @@ public String getName() {
...
         bootstrap.addBundle(new TypesafeConfigurationBundle("dpc.aggregation"));
+        bootstrap.addBundle(new MigrationsBundle<DPCAggregationConfiguration>() {
+            @Override
+            public DataSourceFactory getDataSourceFactory(DPCAggregationConfiguration dpcAggregationConfiguration) {
+                return dpcAggregationConfiguration.getQueueDatabase();
+            }
+
+            @Override
+            public String getMigrationsFileName() {
+                return "migrations/queue_v2.xml";
```
nickrobison-usds: Should we have a standard naming convention for migrations? e.g. queue.migrations.xml?

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
```diff
@@ -38,67 +34,55 @@
...
     private Disposable subscribe;
 
     /**
-     * Create an engine. Most likely this will be a singleton.
+     * Create an engine.
      *
+     * @param aggregatorID - The ID of the current working aggregator
      * @param bbclient    - {@link BlueButtonClient } to use
-     * @param queue       - {@link JobQueue} that will direct the work done
+     * @param queue       - {@link JobQueueInterface} that will direct the work done
```
nickrobison-usds: Should we standardize on a naming convention, e.g. IJobQueue?

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
```diff
@@ -115,77 +99,96 @@ public void stop() {
...
     private void pollQueue() {
-        subscribe = Observable.fromCallable(this.queue::workJob)
+        subscribe = Observable.fromCallable(() -> this.queue.workBatch(aggregatorID))
                 .doOnNext(job -> logger.trace("Polling queue for job"))
                 .filter(Optional::isPresent)
                 .map(Optional::get)
                 .repeatWhen(completed -> {
-                    logger.debug("No job, retrying in 2 seconds");
+                    logger.debug("No job, polling again in 2 seconds");
                     return completed.delay(2, TimeUnit.SECONDS);
```
nickrobison-usds: Can we pull this out as a config variable and potentially shorted it to 500 ms? We can make this a separate ticket, if we want.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
```diff
@@ -115,77 +99,96 @@ public void stop() {
...
 
-            final var errorCounter = new AtomicInteger();
-            final var results = Flowable.fromIterable(job.getResourceTypes())
-                    .flatMap(resourceType -> completeResource(job, resourceType, errorCounter))
-                    .toList()
-                    .blockingGet(); // Wait on the main thread until completion
+            Optional<String> nextPatientID = job.fetchNextBatch(aggregatorID);
+
+            // Stop processing when no patients or early shutdown
+            Boolean queueRunning = true;
```
nickrobison-usds: Does this need to be `volatile`?

dpc-aggregation/src/main/resources/banner.txt
```diff
@@ -0,0 +1,22 @@
+
```
nickrobison-usds: Excellent.

dpc-aggregation/src/main/resources/migrations/queue_v2.xml
```diff
@@ -0,0 +1,71 @@
...
+            <column name="status"></column>
+            <column name="aggregator_id"></column>
+            <column name="update_time"></column>
+        </createIndex>
+
+        <createTable tableName="JOB_QUEUE_BATCH_FILE">
+            <column name="batch_id" type="UUID">
+                <constraints nullable="false"/>
+            </column>
+            <column name="job_id" type="UUID">
```
nickrobison-usds: Do we need a foreign key constraint here?

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
```diff
@@ -59,70 +62,87 @@ public JobResource(JobQueue queue, @APIV1 String baseURL) {
...
-                    break;
-                }
-                default: {
-                    builder = builder.status(HttpStatus.ACCEPTED_202);
+        }
+
+        Response.ResponseBuilder builder = Response.noContent();
+        Set<JobStatus> jobStatusSet = batches.stream().map(JobQueueBatch::getStatus).collect(Collectors.toSet());
+
+        if ( jobStatusSet.contains(JobStatus.FAILED) ) {
```
nickrobison-usds: Can we pull this out into its own method?

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
```diff
@@ -59,70 +62,87 @@ public JobResource(JobQueue queue, @APIV1 String baseURL) {
...
+            // If any part of the job has failed, report a failed status
+            builder = builder.status(HttpStatus.INTERNAL_SERVER_ERROR_500);
+        } else if ( jobStatusSet.contains(JobStatus.RUNNING) || jobStatusSet.contains(JobStatus.QUEUED) ) {
+            // The job is still being processed
+            String progress = "QUEUED: 0.00%";
+            if ( jobStatusSet.contains(JobStatus.RUNNING) ) {
+                AtomicInteger done = new AtomicInteger();
```
nickrobison-usds: Do these need to be atomics? Seems like everything should be handled on a single thread, correct?

dpc-attribution/docker/entrypoint.sh
```diff
@@ -10,12 +10,13 @@ fi
 
 CMDLINE="java ${JACOCO} -cp /app/resources:/app/classes:/app/libs/* gov.cms.dpc.attribution.DPCAttributionService"
 
+echo "Migrating the database"
```
nickrobison-usds: Environment variable?

dpc-common/src/main/java/gov/cms/dpc/common/hibernate/DPCHibernateModule.java
```diff
@@ -24,16 +26,8 @@ public void configure(Binder binder) {
...
-    DPCManagedSessionFactory getSessionFactory(DPCHibernateBundle hibernate) {
-        // This is necessary because the session factory doesn't load on its own.
-        // I'm really not sure how to fix this, I think it's due to the interaction with the Proxy Factory
-        try {
-            hibernate.run(getConfiguration(), getEnvironment());
-        } catch (Exception e) {
-            throw new IllegalStateException(e);
-        }
-        return new DPCManagedSessionFactory(hibernate.getSessionFactory());
+    DPCManagedSessionFactory getSessionFactory() {
```
nickrobison-usds: So glad this is gone!

dpc-common/src/main/java/gov/cms/dpc/common/hibernate/DPCHibernateBundle.java
```diff
@@ -34,21 +29,16 @@
     private static final Logger logger = LoggerFactory.getLogger(DPCHibernateModule.class);
     public static String PREFIX_STRING = "gov.cms.dpc.common.entities";
 
-
-    // We have to use a Guice specific annotation to get optional binding of additional class paths to scan
-    @Inject(optional = true)
-    @AdditionalPaths
-    private static List<String> additionalPaths;
```
nickrobison-usds: Awesome! This needed to go away.

dpc-common/src/main/java/gov/cms/dpc/common/hibernate/DPCQueueHibernateBundle.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Custom Hibernate bundle, which allows us to inject the {@link org.hibernate.SessionFactory} into non-resource types.
+ * By default, the bundle scans the {@link DPCQueueHibernateBundle#PREFIX_STRING} for a list of annotated entities.
+ *
+ * @param <T> - Configuration class type
+ */
+@Singleton
+public class DPCQueueHibernateBundle<T extends Configuration & IDPCQueueDatabase> extends HibernateBundle<T> implements ConfiguredBundle<T> {
```
nickrobison-usds: Should this be a sub-class of HibernateBundle?

dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedBatchQueue.java
```diff
@@ -0,0 +1,293 @@
...
+            } finally {
+                tx.commit();
+            }
+        }
+    }
+
+    @Override
+    public long queueSize() {
+        try (final Session session = this.factory.openSession()) {
+            try {
```
nickrobison-usds: Can we pull this out so that the query is only built once? Maybe in a static or as part of the class instantiation?

dpc-queue/src/main/java/gov/cms/dpc/queue/JobQueueCommon.java
```diff
@@ -0,0 +1,53 @@
...
+                .blockingGet();
+
+        // Expect a single empty job when no patients passed
+        if ( jobBatches.isEmpty() && patients.isEmpty() ) {
+            jobBatches.add(this.createJobBatch(jobID, orgID, providerID, Collections.emptyList(), resourceTypes));
+        }
+
+        // Set the priority of a job batch
+        // Single patients will have first priority to support patient everything
+        final int priority = patients.size() == 1 ? 1000 : 5000;
```
nickrobison-usds: I like this!

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/DPCAggregationConfiguration.java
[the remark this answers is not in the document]
ronaldheft-usds: @nickrobison-usds Encryption options were removed per a discussion I had with Rick that encryption was not part of the FHIR spec.

Parallelization options were also removed for the time being. DPC-640 was created to add it back in. Technically, parallelization is happening right now as multiple aggregators will be working on a job, but to better use available resources, aggregators can also process multiple batches at a time (which is the goal of DPC-640).

The previous parallelization model of working on multiple patients at a time on a single aggregator was no longer feasible while supporting job resuming, as after each patient the aggregator "checks in", saves the status to the database, and writes out the patient data to the file. If we wanted to support processing multiple patients within a batch, we'd need a slightly different DB model that would track that status of each patient, and the output would need a separate file per patient.

If you think it's important for launch, we can hold on release Queue V2 until DPC-640 is ready.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/DPCAggregationService.java
ronaldheft-usds: @nickrobison-usds Improved the comment.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
ronaldheft-usds: @nickrobison-usds I don't think so. The variable is local to the method and not shared with other threads. The underlying variable is derived from `this.subscribe` which is an instance of `Disposable`, which is [thread safe](http://reactivex.io/RxJava/javadoc/io/reactivex/observers/DisposableObserver.html).

dpc-aggregation/src/main/resources/migrations/queue_v2.xml
ronaldheft-usds: @nickrobison-usds A foreign key constraint is defined on `batch_id`. https://github.com/CMSgov/dpc-app/pull/218/files/445d2331c4e107f6803e6b5de6b63ff971c8ef6f#diff-5e34fb4b3d78df68f85110b83aef829aR67

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
ronaldheft-usds: @nickrobison-usds I like the idea of standardizing on a name convention. Did see both `IServiceName` and `ServiceNameInterface` patterns used throughout the app. I like the `IServiceName` pattern however, as it's shorter, so I renamed it.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/DPCAggregationService.java
ronaldheft-usds: @nickrobison-usds I like this naming convention. Renamed it!

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
ronaldheft-usds: @nickrobison-usds Refactored them into their own methods.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/DPCAggregationConfiguration.java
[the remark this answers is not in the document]
nickrobison-usds: Sounds good to me! Thanks for the clarification, seems like the scale-out parallelization is definitely the right approach going forward.

dpc-aggregation/docker/entrypoint.sh
```diff
@@ -10,5 +10,8 @@ fi
 
 CMDLINE="java ${JACOCO} -cp /app/resources:/app/classes:/app/libs/* gov.cms.dpc.aggregation.DPCAggregationService"
 
+echo "Migrating the database"
```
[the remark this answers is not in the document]
ronaldheft-usds: @nickrobison-usds Added! There is now an environment variable `DB_MIGRATION`. It's set to enabled by default. With `DB_MIGRATION=0` the migrations will be skipped.

dpc-attribution/docker/entrypoint.sh
ronaldheft-usds: Done!

dpc-common/src/main/java/gov/cms/dpc/common/hibernate/DPCQueueHibernateBundle.java
ronaldheft-usds: @nickrobison-usds It is a sub-class of HibernateBundle, the same as the original DPCHibernateBundle.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
ronaldheft-usds: @nickrobison-usds This is currently on a single thread, but I was doing this in a stream before. Refactoring a bit.

dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedBatchQueue.java
ronaldheft-usds: @nickrobison-usds How would this work? CriteriaBuilder requires a database session.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
ronaldheft-usds: @nickrobison-usds Made it configurable!

ronaldheft-usds:
@nickrobison-usds Thanks!

The AggregatorID in provided in the JobQueueModule. See: https://github.com/CMSgov/dpc-app/pull/218/files#diff-0f972af4be048b7edf2f35a5c9040cd4R50

All requested changes should be addressed now. Let me know if I missed anything.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/OperationsConfig.java
```diff
@@ -4,40 +4,33 @@
...
-
-
-    public OperationsConfig(int resourcesPerFileCount,
-                            String exportPath,
-                            int retryCount,
-                            boolean encryptionEnabled,
-                            boolean parallelEnabled,
-                            float writeThreadFactor,
-                            float fetchThreadFactor) {
+    private int pollingFrequency;
```
nickrobison-usds: Thoughts on using BusyWait here? Could potentially reduce the latency for Patient/$everything operations. Not totally sure how much an improvement that would be though.

ronaldheft-usds: @nickrobison-usds I'm unsure of what you mean by BusyWait. Could you clarify?

I think our current pattern of polling and sleeping is sufficient, as in practicality across the cluster the polling time should naturally diverge.

While each queue thread may only be checking once every 500ms, across the cluster the threads will be completing their polling check at different 500ms intervals. The 500ms check time isn't locked a clock interval (like a cron job) and is instead based on 500ms from the last completed job. Additionally, when DPC-640 is completed and multiple queue threads are running on each aggregator, there will be even more queue threads to provide a further differentiation. Thus, 500ms should be the maximum time to pickup a job (if the queue is idle) and be even less in practicality.

If we do want to improve the speed of picking up jobs, we could implement Postgres WAIT and NOTIFY: https://www.postgresql.org/docs/9.3/libpq-notify.html

ronaldheft-usds:
DPC-642 (missing ndjson extension) has been included in this PR.

ronaldheft-usds:
TODO:
- Log a ticket to track changing Flowable
- Handle clearing files when restarting a stuck job

ronaldheft-usds:
Handling the file appending issues of restarting batches was fixed by disabling append mode on the `FileOutputStream` when writing a new file.

We do not need to remove Flowable, as it's still used to handle parallelizing multiple requests of a patient's data.

ronaldheft-usds:
@RickHawesUSDS @nickrobison-usds All requested changes should now be addressed.

@switzersc-usds Feel free to look over things. I can give you a run down of the changes tomorrow or Wednesday.

dpc-aggregation/pom.xml
```diff
@@ -18,10 +18,19 @@
...
+        <dependency>
+            <groupId>gov.cms.dpc</groupId>
+            <artifactId>dpc-common</artifactId>
+            <version>0.4.0-SNAPSHOT</version>
+        </dependency>
         <dependency>
             <groupId>io.dropwizard</groupId>
             <artifactId>dropwizard-core</artifactId>
         </dependency>
+        <dependency>
```
nickrobison-usds: Do we still need the migrations library?

dpc-attribution/pom.xml
```diff
@@ -177,6 +177,9 @@
                         <ports>
                             <port>8080</port>
                         </ports>
+                        <environment>
+                            <DB_MIGRATION>1</DB_MIGRATION>
+                        </environment>
```
nickrobison-usds: Can we pull the dropwizard-migrations library out?

dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedBatchQueue.java
nickrobison-usds: Good point. Ignore me.

nickrobison-usds:
So excited for this!

dpc-aggregation/pom.xml
ronaldheft-usds: @nickrobison-usds Yes, we do. The aggregation service now runs the `dpc_queue` migrations and the build fails without the dependency.

I thought of refactoring the migrations library into dpc-common, since 2 out of 4 projects require it, but I thought it better to explicitly include it each project.

dpc-attribution/pom.xml
ronaldheft-usds: @nickrobison-usds The main migrations are still being run out of dpc-attribution, so migrations for the `dpc_attribution` database require it.

ronaldheft-usds:
@nickrobison-usds Awesome. Will hold on the merge until we're ready to deploy, as this requires Terraform changes before we can roll it out.

PR #219, opened 2019-09-04
Enable InternalUser login via Github

switzersc-usds:
**Why**

We want to manage internal user authentication via Github teams.

**What Changed**

This PR introduces the ability for internal users to sign in (and sign up) via Github OAuth. 

New sign-in page for internal users:
image

**Choices Made**

* Uses Github Omniauth gem and Devise for OAuth flow
* Removes first/last names from InternalUsers because Github doesn't store these or provide them
* Saves github name and nickname (login) in case we want those
* Verifies that we are only signing in users on a pre-approved org and team

**Checklist**

~- [ ] Demo and Seed commands are working~
~- [ ] Swagger documentation has been updated~
~- [ ] FHIR documentation has been updated~


dpc-web/app/controllers/internal/auth/omniauth_callbacks_controller.rb
```diff
@@ -25,12 +30,24 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
       # end
+
+      def valid_org_team?
+        github_client.user_teams.any? do |team|
+          team[:id].to_s == ENV.fetch('GITHUB_ORG_TEAM_ID')
+        end
+      end
+
+      def github_client
+        @github_client ||= Octokit::Client.new(
```
switzersc-usds: We probably don't need the whole octokit library since we're just making one request but HTTP in raw Ruby is pretty annoying (we'd probably want to introduce an HTTP client library anyway) and this makes API auth and navigation way easier.

nickrobison-usds: I'm fine with that!

PR #220, opened 2019-09-05
DPC-545: Refactor token endpoints

nickrobison-usds:
**Why**

As a foundation for supporting more complex auth workflows as part of DPC-511, we need to refactor our Token endpoints into their own Resource in order to decouple auth operations from more generic Org changes.

**What Changed**

All token operations have been moved from the  `/Organization` endpoint to `/Token`, this should help separate concerns and make future refactoring easier (e.g. we may need to actually move the Token ops into the API layer). Also created a `TokenDAO` class which handles just the token operations but does support doing a check that the org actually exists before doing creation operations.

Finally, as part of these changes I was able to re-enable the OrganizationRegistration and OrganizationResource tests. This required some tweaks to the way we handle organization creation but should finally address DPC-413.

**Choices Made**

**Tickets closed**:

DPC-545: This ticket
DPC-413: Organization tests are no longer failing in CI.

**Future Work**

DPC-511: SMART backend auth

**Checklist**


dpc-api/src/main/java/gov/cms/dpc/api/cli/tokens/TokenList.java
```diff
@@ -60,28 +61,27 @@ public void run(Bootstrap<?> bootstrap, Namespace namespace) throws IOException
...
 
-    private void generateTable(List<TokenResponse> tokens) {
+    private void generateTable(List<TokenEntity> tokens) {
         // Generate the table
         final String[] headers = {"Token ID", "Type", "Expires"};
 
         System.out.println(FlipTable.of(headers, tokens
                 .stream()
-                .map(token -> new String[]{token.getId(), token.getType().toString(), token.getExpires()}).toArray(String[][]::new)));
+                .map(token -> new String[]{token.getId(), token.getTokenType().toString(), "token.get()"}).toArray(String[][]::new)));
```
ronaldheft-usds: Should `token.get()` be in quotes?

nickrobison-usds: Good catch, the refactor removed the ability to report expiration times, so I've pulled it out until DPC-617 gets merged in, which will re-add it.

PR #221, opened 2019-09-05
Resolve CVE-2015-9284 vulnerability

switzersc-usds:
**Why**

We need to make sure that we're mitigating the Cross Site Scripting
Forgery vulnerability by using a POST method to initiate the Omniauth
flow. For more info:

https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284
https://github.com/omniauth/omniauth/pull/809

**What Changed**

* Add a gem that enforces using `post` only with oauth initiation requests
* Change out github link to use the `post` method

### DIFF
--- a/dpc-web/Gemfile
+++ b/dpc-web/Gemfile
@@ -57,4 +57,5 @@ gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 gem 'delayed_job_active_record'
 gem 'daemons'
 gem 'omniauth-github'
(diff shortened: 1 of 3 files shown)


switzersc-usds:
While this mitigates the risk, it unfortunately doesn't get rid of Github's warning.

nickrobison-usds:
I was just going to ask!

nickrobison-usds:
Ok, I was able to dismiss the security warning via the 'fix has already been started' option. Not sure that's the best approach, but should work for now. I agree, we need better automated checking on PRs.

PR #222, opened 2019-09-05
Add bundler audit to CI

switzersc-usds:
**Why**

We want to get vulnerability warnings when a PR is opened.

**What Changed**

Add bundler-audit to codeclimate config. For more info:

https://docs.codeclimate.com/docs/bundler-audit
https://github.com/rubysec/bundler-audit

### DIFF
--- a/dpc-web-test.sh
+++ b/dpc-web-test.sh
@@ -16,7 +16,7 @@ docker-compose -f dpc-web/docker-compose.yml run web rails db:create db:migrate 
 docker-compose -f dpc-web/docker-compose.yml run web rails spec
 
 # Run bundler audit
-docker-compose -f dpc-web/docker-compose.yml run web bundle audit update && bundle audit check --ignore CVE-2015-9284
+docker-compose -f dpc-web/docker-compose.yml run web cd dpc-web && bundle audit update && bundle audit check --ignore CVE-2015-9284
 
 echo "┌──────────────────────────────────────────┐"
 echo "│                                          │"
--- a/dpc-web-test.sh
+++ b/dpc-web-test.sh
@@ -16,7 +16,7 @@ docker-compose -f dpc-web/docker-compose.yml run web rails db:create db:migrate 
 docker-compose -f dpc-web/docker-compose.yml run web rails spec
 
 # Run bundler audit
-docker-compose -f dpc-web/docker-compose.yml run web gem install bundler-audit && bundle audit update && bundle audit check --ignore CVE-2015-9284
+docker-compose -f dpc-web/docker-compose.yml run web bundle exec bundle audit update && bundle exec bundle audit check --ignore CVE-2015-9284
 docker-compose -f dpc-web/docker-compose.yml down
 
 echo "┌──────────────────────────────────────────┐"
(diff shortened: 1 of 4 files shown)


switzersc-usds:
I opened the pull request to test that out -- assuming the code climate part of the build would fail and then, if it did, seeing if there's a way to customize the config that's just not documented or adding it as a separate rake task. I'm a bit confused about why it's not failing though.

dpc-web-test.sh
```diff
@@ -3,7 +3,7 @@ set -e
 
 echo "┌──────────────────────────────────────────┐"
 echo "│                                          │"
-echo "│        Running Website Tests...          │"
+echo "│  Running Website Tests & Bundler Aduit   │"
```
nickrobison-usds: Typo here

dpc-web-test.sh
```diff
@@ -13,10 +13,16 @@ make website
...
+docker-compose -f dpc-web/docker-compose.yml run web gem install bundler-audit
+docker-compose -f dpc-web/docker-compose.yml run web bundle audit update
+docker-compose -f dpc-web/docker-compose.yml run web bundle audit check --ignore CVE-2015-9284
+
 docker-compose -f dpc-web/docker-compose.yml down
 
 echo "┌──────────────────────────────────────────┐"
 echo "│                                          │"
-echo "│        All Website Tests Complete        │"
+echo "│     All Website Tests & AuditComplete    │"
```
nickrobison-usds: Space here?

dpc-web-test.sh
switzersc-usds: Ha! Was rushing to catch the shuttle. Updating.

nickrobison-usds:
@switzersc-usds Any progress on this?

switzersc-usds:
I ran into a blocker and dropped it to get some other work done. Can pick this up again this week. @nickrobison-usds

dpc-web-test.sh
```diff
@@ -15,8 +15,11 @@ docker-compose -f dpc-web/docker-compose.yml up start_core_dependencies
 docker-compose -f dpc-web/docker-compose.yml run web rails db:create db:migrate db:seed
 docker-compose -f dpc-web/docker-compose.yml run web rails spec
 
+# Run bundler audit
+docker-compose -f dpc-web/docker-compose.yml run web cd dpc-web && bundle audit update && bundle audit check --ignore CVE-2015-9284
```
switzersc-usds: I'm a little confused about why CDing into the web directory is necessary here but not in the above commands...

switzersc-usds:
@nickrobison-usds this should be good to go now.

dpc-web-test.sh
switzersc-usds: @ronaldheft-usds or @nickrobison-usds any ideas on this? The workdir for the web container is dpc-web so I'm not sure why I need to `cd` into that dir, but it appears I do.

nickrobison-usds: That's a really great question, I'm not sure either. Seems like a simple solution, definitely something to keep an eye on in the future, in case it crops up again.

PR #228, opened 2019-09-11
As a CMS user I want to view users who have requested access.

switzersc-usds:
**Why**

CMS users need to see users to start managing them.

**What Changed**

Adds an internal path for users index and show so that an internal user can view a list of users who have requested access to DPC and see a user's individual details.

*Index*
Screen Shot 2019-09-11 at 8 59 44 AM
Screen Shot 2019-09-11 at 9 08 01 AM
Screen Shot 2019-09-11 at 9 08 08 AM

*Show*
Screen Shot 2019-09-11 at 8 56 04 AM
Screen Shot 2019-09-11 at 8 56 19 AM

**Choices Made**

The main goal of this PR is to get an MVP of the pages for viewing users. The styles match the mockups at about 85% and the pages are functional though not particularly beautiful on mobile. The index page leaves space on the left side for the internal user nav menu which is coming soon. 

I tried to use CMS design guide as much as possible and to follow the css class naming conventions when I had to create new styles; I'm not a front-end person though so they could probably use some help.

**Future Work**

- Improve the styles to fully match mockups and be more responsive
- Add filtering and pagination

**Checklist**


dpc-web/app/controllers/internal/users_controller.rb
```diff
@@ -3,6 +3,13 @@
 module Internal
   class UsersController < ApplicationController
     before_action :authenticate_internal_user!
-    def index; end
+
+    def index
+      @users = User.order('created_at DESC').page params[:page]
```
nickrobison-usds: Do we have an index on create_at time?

dpc-web/app/views/internal/users/index.html.erb
```diff
@@ -1,3 +1,48 @@
...
+<% title "Manage Users" %>
+<%= render partial: 'shared/alerts', locals: { notice: flash[:notice], alert: flash[:alert] } %>
+
+<div class="ds-l-row">
+  <div class="ds-l-row--12 ds-l-md-col--3">
+  </div>
+
+  <div class="ds-l-row--12 ds-l-md-col--9">
+    <div class="ds-l-row">
+      <h1 class="ds-h1">Viewing <b><%= @users.total_count %></b> provider users</h1>
```
nickrobison-usds: Does this value take into account pagination? E.g. 20 of 200 users?

dpc-web/app/controllers/internal/users_controller.rb
switzersc-usds: No. Shall I add one here?

nickrobison-usds: Yes, please.

PR #241, opened 2019-09-17
DPC-236: Add support for organization public keys

nickrobison-usds:
**Why**

In order to fully support SMART Backend Services authentication, we need the ability to store and managed public keys for organizations, which will be used for signing the JWT token requests.

**What Changed**

This PR adds a `/Key` endpoint to the API service, which supports basic CREATE, GET, DELETE functionality for each organization.

This also brings along some of the logic required for parsing PEM encoded keys into the raw Byte arrays that Java understands. This is pretty tricky, but by pulling in the `bouncy castle` library, we can abstract most of the nasty bits. One of the unit tests confirms that when doing the key round trip, that it's still usable for decryption and signature verification. Which is kind of important.

**Choices Made**

The decision was made to pull the auth functionality out of the Attribution Service and into the API layer. This should simplify the implementation and avoid an extra round trip when doing basic auth functionality.

**Tickets closed**:

DPC-236: 

**Future Work**

Migrate the tokens and root_keys table from the `public` schema to the new `auth` db, based on the work in #218.

Implement public key validations (e.g. algorithm and length)

DPC-663: Services should use unique usernames and passwords for DB connections.

**Checklist**


dpc-api/src/main/java/gov/cms/dpc/api/DPCAPIModule.java
```diff
@@ -86,7 +91,7 @@ public String provideV1URL(@ServiceBaseURL String baseURL) {
     @Provides
     @AdditionalPaths
     public List<String> provideAdditionalPaths() {
-        return List.of("gov.cms.dpc.queue.models");
+        return List.of("gov.cms.dpc.queue.models", "gov.cms.dpc.api.entities");
```
ronaldheft-usds: If this doesn't get merged in before #218, I'll show you where this change needs to go in the refactored HibernateBundle.

dpc-api/src/test/java/gov/cms/dpc/api/auth/PublicKeyHandlerTest.java
```diff
@@ -0,0 +1,86 @@
...
+        final KeyPairGenerator kpg = KeyPairGenerator.getInstance("RSA");
+        final KeyPair keyPair = kpg.generateKeyPair();
+        final String encoded = Base64.getMimeEncoder().encodeToString(keyPair.getPrivate().getEncoded());
+
+        final String key = String.format("-----BEGIN RSA PRIVATE KEY-----\n%s\n-----END RSA PRIVATE KEY-----\n", encoded);
+        assertThrows(PublicKeyException.class, () -> PublicKeyHandler.parsePEMString(key));
+    }
+
+    @Test
+    void testEncryptionRoundTrip() throws NoSuchPaddingException, NoSuchAlgorithmException, InvalidKeyException, BadPaddingException, IllegalBlockSizeException, IOException, InvalidKeySpecException {
```
ronaldheft-usds: Nice test and great to see the round-trip encryption working!

dpc-attribution/src/main/resources/auth.migrations.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ronaldheft-usds: If we break this out into its own database, are you still anticipating having the attribution server run the migration? For #218 I've gone with the pattern of having the aggregation server run the queue database migrations. Maybe it will make sense to have the API server run the authentication schema migrations?

nickrobison-usds: Agreed. I'd like to pull the auth functionality entirely into the api service, so we'll need it to run its own migrations.

dpc-api/src/main/java/gov/cms/dpc/api/DPCAPIModule.java
nickrobison-usds: I'm going to hold this until #218 is done, then we'll huddle up and try again.

nickrobison-usds:
@ronaldheft-usds Would you mind taking a look at this again? I wired in the Auth DB to the API service and made a couple of tweaks to the Hibernate Modules and Bundles. I want to make sure I did it correctly!

nickrobison-usds:
The code coverage numbers are fine for now, the integration tests will come when we land the rest of DPC-511.

dpc-queue/src/main/java/gov/cms/dpc/queue/models/JobQueueBatch.java
```diff
@@ -278,13 +278,13 @@ public void setRunningStatus(UUID aggregatorID) {
      * Returns null if at the end of the list.
      */
     public Optional<String> fetchNextPatient(UUID aggregatorID) {
-        if ( this.status != JobStatus.RUNNING ) {
+        if (this.status != JobStatus.RUNNING) {
```
ronaldheft-usds: There are a bunch of lines that changed because our code formatting is different in IntelliJ (and IntelliJ updated the formatting during your refactor). We should define an .editorconfig for the project so we keep consistent formatting across developers.

nickrobison-usds: Yes. Let's do that!

PR #242, opened 2019-09-17
As a CMS user, I want to tag users.

switzersc-usds:
**Why**

We want to use tags to help internal users manage users.

**What Changed**

Add tags to the user page so that an internal user can see and add tags to the user. Also add UI for creating tags themselves so that the internal user adds a tag before they can tag the user with that tag (eventually we probably want to change this UI but this was the fastest for now).

Managing tags:
Screen Shot 2019-09-17 at 12 24 11 PM

When you try to delete a tag that is being used:
Screen Shot 2019-09-17 at 12 25 51 PM

Adding tags to a user:
Screen Shot 2019-09-17 at 11 37 49 AM

Tags visible from the index page:
Screen Shot 2019-09-17 at 1 26 20 PM

**Choices Made**

I made this a polymorphic association with a `has_many: :through` association, so that we can use tags for organizations, vendors, etc, eventually rather than creating separate tables for those taggings. This architecture is more extensible and DRYer. 

**Tickets closed**:

Part of DPC-510

**Future Work**

- Improve UI
- Add filtering on users#index by tag

### DIFF
--- a/dpc-web/app/views/internal/tags/index.html.erb
+++ b/dpc-web/app/views/internal/tags/index.html.erb
@@ -1,0 +1,68 @@
+<% title "Manage Tags" %>
+
+
+<div class="ds-l-row">
+  <div class="ds-l-row--12 ds-l-md-col--3">
+    <ul class="ds-c-vertical-nav">
+      <li class="ds-c-vertical-nav__item">
+        <%= link_to "Users", internal_users_path, class: "ds-c-vertical-nav__label" %>
+      </li>
+      <li class="ds-c-vertical-nav__item">
+        <%= link_to "Tags", internal_tags_path, class: "ds-c-vertical-nav__label ds-c-vertical-nav__label--current" %>
+      </li>
+    </ul>
+  </div>
+
+  <div class="ds-l-row--12 ds-l-md-col--9 ds-u-padding-bottom--2">
+
+    <div class="ds-l-row">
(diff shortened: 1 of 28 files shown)


switzersc-usds:
The code climate coverage change is related to a java class untouched in this PR.

dpc-web/app/views/internal/tags/index.html.erb
```diff
@@ -0,0 +1,68 @@
...
+                  <svg class="icon" xmlns="http://www.w3.org/2000/svg" aria-hidden="true">
+                    <use xlink:href="/assets/solid.svg#times"></use>
+                  </svg>
+                <% end %>
+              </span>
+
+            <% else %>
+
+              <span class="ds-c-badge ds-u-fill--primary-alt">
+                <%= tag.name %> - <%= tag.taggings.count %>
```
nickrobison-usds: Is the `count` method provided by default via Rails? Just curious.

nickrobison-usds:
@switzersc-usds Go ahead and merge this, the code coverage stuff is fine. Thanks!

dpc-web/app/views/internal/tags/index.html.erb
switzersc-usds: It's a method on Ruby `Array` but Rails ActiveRecord adds a method `count` on ActiveRecord Relations that converts this to SQL `COUNT` to be more performant.

PR #245, opened 2019-09-24
DPC-564: Expire attributions rather than removing them

nickrobison-usds:
**Why**

We're currently removing attributions from the database as soon as they're expired, which is less useful than it can be. Ideally, we'd expire patients and leave them as part of the roster (in an inactive state) to allow organizations to reconcile their roster memberships.

**What Changed**

Refactored the attribution logic to allow us to add and remove patients by setting their status to be `inactive`, rather than simply removing them from the database. This should be much more flexible and the SQL is less pathological as we're now handling membership on a relationship level, rather than at the higher, Roster level. We now also return the period in which the patient is/was a member of the roster.

Expiration job now auto-expires membership and removes and expired relationships older than 6 months.

Cleaned up the add/remove Roster operations, they were `PUT`s rather than `POST`s, as they should have been. Adding a patient that is already a group member means their expiration window lengthens, which is how providers can 're-attest'.

During this process we also fixed DPC-634, which means we're now clearly returning error messages when trying to attribute a patient that doesn't exist in the database.

**Choices Made**

**Tickets closed**:

DPC-564: This ticket
DPC-634: Clear error messages for missing patients.
DPC-643: Missing NPI system
DPC-650: Correct start/end temporals in member periods.

**Future Work**

DPC-651: Add inactive patients to export response as operation outcomes.

**Checklist**


dpc-attribution/src/test/java/gov/cms/dpc/attribution/AttributionTestHelpers.java
```diff
@@ -16,7 +16,7 @@
 
     public static Practitioner createPractitionerResource(String NPI) {
         final Practitioner practitioner = new Practitioner();
-        practitioner.addIdentifier().setValue(NPI);
+        practitioner.addIdentifier().setValue(NPI).setSystem(DPCIdentifierSystem.NPPES.getSystem());
```
ronaldheft-usds: Is this correct per <https://groups.google.com/forum/#!topic/dpc-api/DZuFWIVh4wI>?

dpc-attribution/src/test/java/gov/cms/dpc/attribution/SharedMethods.java
```diff
@@ -24,7 +24,7 @@ public static Bundle createAttributionBundle(String providerID, String patientID
 
         // Create the provider with the necessary fields
         final Practitioner practitioner = new Practitioner();
-        practitioner.addIdentifier().setValue(providerID).setSystem(DPCIdentifierSystem.MBI.getSystem());
+        practitioner.addIdentifier().setValue(providerID).setSystem(DPCIdentifierSystem.NPPES.getSystem());
```
ronaldheft-usds: Same as: https://groups.google.com/forum/#!topic/dpc-api/DZuFWIVh4wI

dpc-common/src/main/java/gov/cms/dpc/common/utils/SeedProcessor.java
```diff
@@ -90,7 +90,7 @@ public static Bundle generateAttributionBundle(Map.Entry<String, List<Pair<Strin
 
         // Create the provider with the necessary fields
         final Practitioner practitioner = new Practitioner();
-        practitioner.addIdentifier().setValue(entry.getKey());
+        practitioner.addIdentifier().setValue(entry.getKey()).setSystem(DPCIdentifierSystem.NPPES.getSystem());
```
ronaldheft-usds: Same as https://groups.google.com/forum/#!topic/dpc-api/DZuFWIVh4wI

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/entities/ProviderEntityConverter.java
```diff
@@ -25,7 +24,10 @@ public static Practitioner convert(ProviderEntity entity) {
                 setFamily(entity.getProviderLastName())
                 .addGiven(entity.getProviderFirstName());
 
-        practitioner.addIdentifier().setValue(entity.getProviderNPI());
+        practitioner.addIdentifier()
+                .setSystem(DPCIdentifierSystem.NPPES.getSystem())
+                .setValue(entity.getProviderNPI());
+
```
ronaldheft-usds: Same as https://groups.google.com/forum/#!topic/dpc-api/DZuFWIVh4wI

dpc-attribution/src/test/java/gov/cms/dpc/attribution/AttributionTestHelpers.java
nickrobison-usds: This applies to the serialization of the practitioner resources themselves, we were omitting the system for the ID, which is not right, but linked post refers to the examples of the attribution rosters where I actually copy/pasted incorrectly. Also applies to the other notes as well.

ronaldheft-usds: Cool. Thanks for the clarification!

nickrobison-usds: Yeah, I messed this up earlier, identifiers have _values_ and concepts have _code_. The documentation was updated in #251 to reflect this.

PR #246, opened 2019-09-26
DPC-562: Update docker entrypoint to download S3 config files

nickrobison-usds:
**Why**

Now that we have private keys and certs that need to be setup for connecting to the BB backends, we need the containers to be able to grab the files from S3 on startup.

**What Changed**

Updated the Docker entrypoint to install the AWSCLI and sync the S3 config bucket on startup. This is behind the BOOTSTRAP environment variable, so it should only run when running in AWS.

**Choices Made**

Rather than rebasing our Docker image onto something with the cli already installed, we just grab it on startup each time. It should only had a half a minute or so to the startup time, and we can always optimize later.

**Tickets closed**:

**Future Work**

Ongoing work as part of DPC-560

**Checklist**

### DIFF
--- a/dpc-aggregation/docker/entrypoint.sh
+++ b/dpc-aggregation/docker/entrypoint.sh
@@ -21,7 +21,7 @@ else
 fi
 
 if [ -n "$BOOTSTRAP" ]; then
-  echo "hello"
+  echo "Bootstrapping image from S3"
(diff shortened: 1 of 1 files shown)


nickrobison-usds:
Done. DPC-563 should be tracking it.

PR #253, opened 2019-10-02
style collapsible left nav, static html for later use

jaredcunha-usds:
Screen Recording 2019-10-02 at 11 17 AM

**Why**

Creating separate pages for all users, unassigned users, and assigned users will improve the usability and interaction design of all the filter menus. This means we had to create a collapsible left nav menu to support the additional levels.

**What Changed**

Added CSS for the left navigation and provide sample HTML to use as we build out the internal user experience.

**Choices Made**

Use the accordions javascript for the interactions so we don't have to write new javascript. This is consistent with the way the US Web Design system handles the mobile navigation menus. 

**Future Work**

Hook up to actual views

### DIFF
--- a/dpc-web/app/assets/stylesheets/application.scss
+++ b/dpc-web/app/assets/stylesheets/application.scss
@@ -35,6 +35,7 @@
  @import "components/footer";
  @import "components/hero";
  @import "components/icons";
+ @import "components/left-nav";
  @import "components/modals";
  @import "components/module-containers";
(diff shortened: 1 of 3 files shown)


jaredcunha-usds:
Since the left nav was recently moved into a partial, I went ahead an hooked it up slightly. Still shows pages that don't yet exist (which I think is okay for now) and some of the current state handling still needs to finished as well (which I think also think is okay for now)

PR #261, opened 2019-10-08
DPC-678: Move Auth functions out of attribution service.

nickrobison-usds:
**Why**

Our current authentication design involves both the API service and the Attribution service, which is not only really complex, but means important data is spread across multiple databases. As part of the work for DPC-511, we need to try and centralize all the auth functionality into the API service.

**What Changed**

This ended up being a really big change, as we had to touch almost all of the authenticated tests in both the API and Attribution services.

All auth functionality has been moved out of the attribution service. Moved the `TokenResource` class up to the API service and removed the remaining token functionality from the `OrganizationResource` class, as it's no longer needed.

The Auth handlers now take care of validating the Macaroon, rather than delegating that to an external service.

Created a `dpc_auth` database to hold the necessary authentication information. This will require some downtime during deployment, as we'll need to manually migrate the existing token information from the `dpc_attribution` database to `dpc_auth`. The change _should_ be transparent to the users once it's complete.

Now that we don't handle token management in the attribution service, we need another way for the tests to generate the necessary access tokens and for project admins to onboard organizations. This PR also adds a number of admin `tasks` which can directly manage tokens and organizations. These tasks are now called from the CLI commands, so to the project admin, the workflow should be the same as always. @mamurphy @ronaldheft-usds we'll need to make sure that the API admin port (9900) is accessible via the VPN, but not to external users. 

Added the ability to configure the public facing URL, which is used for Third Party caveats and callback values. Previously, it was generated based on the incoming request context, which was clunky and prone to error if proxy forwarding wasn't configured correctly.

**Choices Made**

Using Dropwizard's task functionality seems to be the best way to handle the fact that we need to bootstrap some tokens and allow for testing of secure functionality, but we may have to revisit in the future.

**Tickets closed**:

DPC-678: Move auth functionality into API service.

**Future Work**

DPC-682: Patient entity is not persisting Gender field.
DPC-683: Patient updating is returning dirty data
DPC-697: Code coverage not reporting correctly for auth related tasks.

**Checklist**


dpc-api/src/main/java/gov/cms/dpc/api/auth/DPCAuthFilter.java
```diff
@@ -68,28 +73,30 @@ public void filter(final ContainerRequestContext requestContext) throws IOExcept
...
 
-        logger.trace("Found {} matching organizations", returnedBundle.getTotal());
-        if (returnedBundle.getTotal() == 0) {
+        // Lookup the organization by Macaroon id
+        final UUID macaroonID = UUID.fromString(m1.identifier);
+
+
+        final UUID orgID = this.dao.findOrgByToken(macaroonID);
+
+        // TODO: This is probably the point to handle the actual Macaroon validation
```
ronaldheft-usds: Is this still a TODO? Looks like validation is taking place on line 93.

dpc-api/src/main/java/gov/cms/dpc/api/tasks/GenerateClientTokens.java
```diff
@@ -0,0 +1,65 @@
...
+        this.bakery = bakery;
+        this.resource = resource;
+    }
+
+    @Override
+    public void execute(ImmutableMultimap<String, String> parameters, PrintWriter output) throws Exception {
+
+        final ImmutableCollection<String> organizationCollection = parameters.get("organization");
+        if (organizationCollection.isEmpty()) {
+            logger.warn("CREATING UNRESTRICTED MACAROON. ENSURE THIS IS OK");
```
ronaldheft-usds: 🚨 🚨 🚨 👍

dpc-api/src/test/java/gov/cms/dpc/api/AbstractSecureApplicationTest.java
```diff
@@ -38,13 +37,19 @@ protected String getBaseURL() {
...
+        return String.format("http://localhost:%d/tasks/", APPLICATION.getAdminPort());
+    }
+
     @BeforeAll
     public static void setup() throws Exception {
         APITestHelpers.setupApplication(APPLICATION);
         ctx = FhirContext.forDstu3();
         // Register a test organization for us
+        // First, create a Golden macaroon for admin uses
+        final String goldenMacaroon = APITestHelpers.createGoldenMacaroon();
```
ronaldheft-usds: 🌅

src/main/resources/server.conf
```diff
@@ -32,12 +32,12 @@ server {
...
-    type = console
-    logFormat = "%-5p [%d{ISO8601,UTC} - %t] %c: %replace(%m){'\n','\r'}\r%replace(%rEx){'\n','\r'}"
-  }]
-}
+//logging {
+//  appenders = [{
+//    type = console
+//    logFormat = "%-5p [%d{ISO8601,UTC} - %t] %c: %replace(%m){'\n','\r'}\r%replace(%rEx){'\n','\r'}"
+//  }]
+//}
```
ronaldheft-usds: Why is this commented out? This was previously required to format the logs in a CloudWatch friendly way.

nickrobison-usds: Bad commit, I'll pull it out. Thanks!

dpc-api/src/main/java/gov/cms/dpc/api/auth/DPCAuthFilter.java
nickrobison-usds: Good catch, I'll remove it.

nickrobison-usds:
@ronaldheft-usds Thanks for the review! I'm going to check in a SQL script to do the token migration, I'm envisioning enabling maintenance mode, deploying the change, migrating the tokens, and reenabling access. Does that make sense? The table structure is the same, so it should be a really simple migration, but I'm going to test it locally today.

ronaldheft-usds:
@nickrobison-usds Couldn't find any instances of separate connection pools in tests, so I'm not sure what I was remembering.

I did check that by default the Postgres Docker image allows 100 connections, which I think would be more than sufficient. Our connection pool is configured at a max of 10 connections, so with all 1 db running in aggregation, 1 db in attribution, 3 dbs in api, we should be looking at max 50 connections (and never see that in a test like we're seeing here). Granted, some of the servers are using both Hibernate and JDBI, so that may open multiple connection pools, but still.

It does look like `APPLICATION.after()` cleans everything up, and it's even shown in an example in the docs to use that for tests.

nickrobison-usds:
@ronaldheft-usds I think each test suite creates its own connection pool to the database, is it possible Postgres isn't evicting the connections in time to allow new ones to connect?

ronaldheft-usds:
@nickrobison-usds Certainly possible. Looks like there is a good resource here for adding extra assertions to tests using Dropwizard to check for connection leaks: https://phillbarber.blogspot.com/2015/02/how-to-test-for-connection-leaks.html

nickrobison-usds:
@ronaldheft-usds Added the leak detection to the integration tests, didn't find anything. However, when I resolved the issue with the MacaroonBakery exception the tests passed all the way through. My guess is that when the tests hard-crash they don't always clean up properly.

I watched the Postgres dashboard during the integration and E2E runs, everything seemed to be behaving properly.

nickrobison-usds:
@ronaldheft-usds I haven't been able to figure out the code coverage issue, seems like there's some really deep wonkiness with JaCoCo, I've created a ticket to track the issue. Thoughts on merging this PR as is? I don't really like seeing the coverage numbers drop so much, but this PR is getting pretty big and sprawling already.

dpc-attribution/src/test/java/gov/cms/dpc/attribution/AbstractAttributionTest.java
```diff
@@ -22,11 +34,31 @@ public static void initDB() throws Exception {
...
+        // Ensure there are no active connections left
+        checkAllConnectionsClosed(String.format("http://localhost:%s", APPLICATION.getAdminPort()));
         APPLICATION.after();
     }
 
     protected String getServerURL() {
         return String.format("http://localhost:%s/v1", APPLICATION.getLocalPort());
     }
+
+    private static void checkAllConnectionsClosed(String adminURL) throws IOException {
```
ronaldheft-usds: Nice to see this test!

PR #262, opened 2019-10-08
DPC-644: Fix provider deletion

nickrobison-usds:
**Why**

A bug was reported where calling `DELETE` on a provider was resulting in a 500 error. This was because the deletion logic wasn't correctly remove attribution relationships. So our unit tests were passing, but the actual functionality was borked.

**What Changed**

Simple tweak to iterate through the rosters linked to a given provider and remove all the attribution relationships and the rosters themselves.

Also verified that patient removal is working as well.

**Choices Made**

**Tickets closed**:

DPC-644: This ticket

**Future Work**

**Checklist**

### DIFF
--- a/dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/ProviderDAO.java
+++ b/dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/ProviderDAO.java
@@ -1,14 +1,11 @@
 package gov.cms.dpc.attribution.jdbi;
 
-import gov.cms.dpc.common.entities.ProviderEntity;
+import gov.cms.dpc.common.entities.*;
 import gov.cms.dpc.common.hibernate.attribution.DPCManagedSessionFactory;
 import io.dropwizard.hibernate.AbstractDAO;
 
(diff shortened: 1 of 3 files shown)


dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/ProviderDAO.java
```diff
@@ -71,7 +68,18 @@ public ProviderEntity persistProvider(ProviderEntity provider) {
...
+    /**
+     * Remove the {@link ProviderEntity} and all associated resources.
+     *
+     * @param provider - {@link ProviderEntity} to remove
+     */
     public void deleteProvider(ProviderEntity provider) {
+
+        // For each roster remove the attributions and the roster itself
+        provider.getAttributionRosters()
+                .forEach(roster -> removeRoster(roster.getId()));
```
ronaldheft-usds: This works, but alternatively we could have added `orphanRemoval=true` and `cascade=CascadeType.ALL` on `PatientEntity.attributionRosters` and Hibernate would handle removing the relationship automatically.

PR #287, opened 2019-10-09
DPC-658: Improve export job response with data integrity elements

embh:
**Why**
In order to ensure data integrity, checksums and byte lengths should be provided for files in the job status response.

**What Changed**
Output files JSON objects in job status responses will now include the `extension` field, containing fields `checksum` for a checksum (SHA-256 currently) and `length` for byte length.

**Choices Made**
This change could have been made as part of the aggregation process or in the API. After discussing with @ronaldheft-usds , API side seemed like a simpler option due to the possibility of jobs pausing and resuming.

**Tickets closed**:
- DPC-658
  - DPC-659: Add file checksums (subtask)
  - DPC-660: Add file byte length (subtask)

**Future Work**
If API performance becomes an issue, determining checksum and file length should be moved into aggregation.

**Checklist**

  - Seed works. Demo receives a 404 on the roster upload step, unrelated to the changes in this branch.


dpc-api/src/main/java/gov/cms/dpc/api/models/JobCompletionModel.java
```diff
@@ -55,6 +61,38 @@ public ResourceType getType() {
...
+         */
+        private String sha256;
+
+        /**
+         * File byte length.
+         */
+        private Long length;
+
+        public OutputEntryExtension(String sha256, Long length) {
+            this.sha256 = sha256;
```
nickrobison-usds: Does it make sense to use a hash prefix? e.g. sha256:{hash value}

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/JobResource.java
```diff
@@ -173,10 +182,31 @@ public Response checkJobStatus(@Auth OrganizationPrincipal organizationPrincipal
...
+            byte[] digest = new SHA256.Digest().digest(fileInputStream.readAllBytes());
+            return Hex.toHexString(digest);
+        } catch (Exception e) {
+            logger.error("Failed to generate checksum", e);
+            return null;
+        }
+    }
+
+    private Long generateFileLength(File file) {
+        return file != null ? file.length() : null;
```
nickrobison-usds: Rather than having a null value, should we thrown an exception, or use a default like 0?

dpc-api/src/main/java/gov/cms/dpc/api/models/JobCompletionModel.java
nickrobison-usds: Rather than calling this sha256, let's just call it checksum, it's a bit more generic and lets us change the hash algorithm as we go along.

dpc-web/app/views/pages/reference.md
```diff
@@ -409,7 +413,7 @@ curl -v https://sandbox.dpc.cms.gov/api/v1/jobs/42 \
 }
 ~~~
 
-Claims data can be found at the URLs within the output field.
+Claims data can be found at the URLs within the output field. The output includes file integrity information in an `extension` object. It contains `sha256`, a SHA-256 checksum, and `length`, the file length in bytes.
```
nickrobison-usds: This level of documentation is fine. If we do make the checksum changes, make sure we explain the first part is the algorithm and the remainder is the actual checksum.

PR #288, opened 2019-10-09
Include demo command in Travis CI tests

ronaldheft-usds:
**Why**

We mention in our PR templates the demo command and seed commands should be tested. We're not currently validating this happens, so it's possible for the commands to rot.

**What Changed**

- Adds the seed and demo commands to our CI script. They will only run on Travis CI.

**Choices Made**

- Places the seed and demo commands behind a `SEED_COMMAND` environment variable so they don't run during the Jenkins build process.

**Tickets closed**:

**Future Work**

- Demo command needs to be fixed up before this is merged.

**Checklist**


ronaldheft-usds:
@nickrobison-usds Looks like the demo command is broken because it's interacting with the attribution server directly and not providing the organization id. I assume we updated the attribution server at some point to require an organization id, as I see there is an APIHelpers method in the API to add it.

Also, looks like the seed command is not broken, but is outputting a ton of warnings. It appears Swagger is doing some type of reflecting on optional dependencies that don't exist. Any idea why that might be happening?

nickrobison-usds:
@ronaldheft-usds Let's remove the demo command, since it's superfluous now, along with any of its dependencies.

We shouldn't need to actually run the seed command, it's automatically run as part of the attribution integration tests, so we can pull that out of the build process and live with the Swagger errors, unless we come up with an easier solution.

PR #312, opened 2019-10-15
Code Hygiene and build cleanups

nickrobison-usds:
**Why**

We've accumulated a number of warnings and errors as part of our build process. This PR takes some time to clean things up and try to squash the warnings.

**What Changed**

Addressed all outstanding errors in our most recent LGTM run.
Added missing Javadoc comments to all methods that were throwing warnings during the build.
Silence some old debug warnings that were set to `WARN` level.

Created a `dpc-testing` module which will be useful in the future. Right now, it holds a new `RingBufferLogger` class which we can use to quiet down the build output. It works by allocating 100 `ILoggingEvent` slots, capturing all of the logging output and then dumping the logs when a test fails. This involved some wiring with JUnit, in order to trigger the log dump, and still needs some better integration with Dropwizard, but it should help a bit and might be useful in production in the future (e.g. silently logging everything and then dumping the last 100 log events when an error occurs).

A couple of small tweaks to the IG.

**Choices Made**

I'm not entirely sure how to quiet down logging for Hibernate and Dropwizard yet, so we'll have to look at improving those in the future.

The build errors for the IG are acceptable, we don't have a publish box requirement, because we're not actually publishing our IG, plus, the tool seems to be broken because it won't accept a copy/paste of the expected output.

**Tickets closed**:

None, just housekeeping

**Future Work**

**Checklist**


nickrobison-usds:
@ronaldheft-usds Sounds good! Thanks for all your help on this!

PR #322, opened 2019-10-23
DPC-712: When I create an organization in the web app, create it in the API.

switzersc-usds:
**Why**

We need to specify which API environments an organization has access to and then create that organization record in the API. 

**What Changed**

* Adds golden macaroon support
* Adds `api_environments` attribute as an array column to `organizations` to keep track of desired envs easily 
* Create a RegisteredOrganization record for every env the org is registered with, which also stores the ID for that org from that env's API
* Update organizations UI to support api_environments, profile endpoints, and NPI
* Send (and save) a fake profile endpoint if it is not provided by user
* Filter input for api_environments to make sure all incoming values
  are arrays without blank items

When editing/creating an org:
Screen Shot 2019-11-05 at 12 29 43 PM
Screen Shot 2019-11-05 at 11 49 43 AM

When an org has an env enabled:
Screen Shot 2019-11-06 at 3 01 03 PM

**Choices Made**
See commit messages. 

**Tickets closed**:
DPC-712

**Future work**:
DPC-757 - UI for adding multiple FHIR endpoints for an org.


dpc-web/app/services/organization_registrar.rb
```diff
@@ -0,0 +1,47 @@
...
+    existing_envs.sort == api_environments.sort
+  end
+
+  def remove_old_registered_organizations
+    removed_envs = existing_envs - api_environments
+    removed_reg_orgs = organization.registered_organizations.where(api_env: removed_envs)
+
+    removed_reg_orgs.each do |registered_org|
+      if APIClient.new(registered_org.api_env).delete_organization(registered_org)
+        pp "Destroyed"
```
switzersc-usds: Need to remove this

dpc-web/app/services/api_client.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def request(uri_string, json, token)
+    uri = URI.parse uri_string
+    headers = { 'Content-Type': 'application/json' }.merge(auth_header(token))
+    http = Net::HTTP.new(uri.host, uri.port)
+    request = Net::HTTP::Post.new(uri.request_uri, headers)
+    request.body = json
+
+    Rails.logger.info 'Making request to FHIR API:'
+    Rails.logger.info request
```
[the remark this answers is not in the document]
switzersc-usds: For sure -- we probably don't need this logging at all actually. Let's touch base tomorrow on what logging already happens for these systems.

dpc-web/app/models/organization.rb
```diff
@@ -6,12 +6,61 @@ class Organization < ApplicationRecord
...
   has_many :organization_user_assignments
   has_many :users, through: :organization_user_assignments
+  has_one :profile_endpoint
+  has_many :registered_organizations
 
   enum organization_type: ORGANIZATION_TYPES
 
   validates :organization_type, inclusion: { in: ORGANIZATION_TYPES.keys }
   validates :name, uniqueness: true, presence: true
+  validate :api_environments_allowed
```
switzersc-usds: Add validation for NPI presence when api_environments are present

dpc-web/Gemfile
```diff
@@ -62,3 +62,5 @@ gem 'omniauth-github'
 gem 'omniauth-rails_csrf_protection', '~> 0.1'
 gem 'octokit'
 gem 'kaminari'
+gem 'fhir_client'
```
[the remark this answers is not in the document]
nickrobison-usds: If we pull this out, we'll need to make sure that we still include the STU3 resources, is that a separate gem?

dpc-web/app/models/organization.rb
```diff
@@ -6,12 +6,61 @@ class Organization < ApplicationRecord
   has_one :address, as: :addressable
   has_many :organization_user_assignments
   has_many :users, through: :organization_user_assignments
+  has_one :profile_endpoint
```
nickrobison-usds: This should probably be a `has_many` relationship. We only require 1 endpoint, but it's possible that organizations would register a val or testing environment as well.

dpc-web/app/models/profile_endpoint.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class ProfileEndpoint < ApplicationRecord
+  belongs_to :organization
+
+  enum connection_type: {
```
nickrobison-usds: Currently, we only support (and plan to only support FHIR Rest), do we still need an enum for that?

dpc-web/app/models/profile_endpoint.rb
```diff
@@ -0,0 +1,40 @@
...
+    'test' => 0,
+    'active' => 1,
+    'suspended' => 2,
+    'error' => 3,
+    'off' => 4,
+    'entered-in-error' => 5
+  }
+
+  validates :name, :uri, :status, :connection_type, presence: true
+  validate :uri_is_valid_format
```
nickrobison-usds: These are great! We'll definitely be adding some more validations in the future.

dpc-web/app/views/internal/organizations/_fields.html.erb
```diff
@@ -1,43 +1,125 @@
...
+
+
+<fieldset class="ds-l-row field-group">
+  <div class="ds-l-col--12 ds-l-md-col--3">
+    <legend class="field-group__heading">Profile endpoint</legend>
+
+  </div>
+
+  <div class="ds-l-col--12 ds-l-md-col--9 field-group__fields">
+    <span class="ds-c-field__hint">You are not required to enter profile endponts unless the organization is graduating from Sandbox.</span>
```
nickrobison-usds: What does _profile endpoint_ mean? Should this be organization FHIR endpoint, or something similar?

dpc-web/app/views/internal/organizations/show.html.erb
```diff
@@ -42,10 +53,23 @@
         <%= @organization.num_providers %> providers
       </div>
 
+      <div class="ds-u-margin-bottom--1">
```
nickrobison-usds: This should probably have an _NPI:_ prefix, it's a little unclear as to what this value is. To me, it looks like the Organization ID, but seems like it's not.

dpc-web/app/views/internal/organizations/_fields.html.erb
switzersc-usds: @nickrobison-usds -- I didn't want to use just `endpoint` (words that begin with `end` are a pain to deal with when using editors that auto-format `end`s) and I didn't know we were only supporting `FHIR` endpoints (as opposed to the other types), so picked `ProfileEndpoint` as a more generic option. I can change this to `FhirEndpoint` though.

dpc-web/app/views/internal/organizations/show.html.erb
switzersc-usds: I was trying to continue the pattern of using an icon rather than a label for this. @jaredcunha-usds - thoughts on a good icon for NPI is, or if none, thoughts on the best style for using a text label here insted? (e.g. maybe we should move this to a different section, or keep it here but make the label inline with the icons, not sure)

dpc-web/app/models/profile_endpoint.rb
switzersc-usds: ah okay, sure, we can remove this then. Yay for removing complexity!

dpc-web/app/models/organization.rb
switzersc-usds: How would we know which to send for which environment the API is registred for then? Or we would post all endpoints in the same payload for every environment we create the organization in? @nickrobison-usds

dpc-web/Gemfile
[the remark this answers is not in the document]
switzersc-usds: @nickrobison-usds not sure what you're talking about. I added this gem in this PR and ended up not using it. Are you using it elsewhere for something?

dpc-web/app/models/organization.rb
nickrobison-usds: Right now, we'll have to duplicate the payload for each environment. I'm ok with holding of on changing this right now, but let's open a ticket to track this improvement.

switzersc-usds: Gotcha. Okay well I'll go ahead and change the association but leave the UI to only support 1 and create a ticket to change it. I have more questions about this and we can talk more about it when we get to that ticket.

dpc-web/app/serializers/organization_submit_serializer.rb
```diff
@@ -0,0 +1,75 @@
...
+          entry: [
+            organization_resource,
+            endpoint_resource
+          ]
+        }
+      }
+    ]
+  end
+
+  def organization_resource
```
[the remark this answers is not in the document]
switzersc-usds: breaking up the serializer would make it harder to read IMO

switzersc-usds:
@nickrobison-usds - I made updates. Let me know what you think!

dpc-web/app/views/internal/organizations/show.html.erb
jaredcunha-usds: Yeah, i'd go iconless on this. There's nothing that really conveys NPI that wouldn't necessarily be confused with something else.

dpc-web/app/serializers/organization_submit_serializer.rb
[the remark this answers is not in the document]
nickrobison-usds: Silenced!

switzersc-usds:
Alrighty addressed the codeclimate issues and made the label change. Will merge on green!

PR #323, opened 2019-10-23
DCP-511: Implement SMART Backend Services

nickrobison-usds:
**Why**

This PR implements required changes in order for us to support the SMART backend services. The details of the spec are listed [here](http://hl7.org/fhir/us/bulkdata/2019May/authorization/index.html).  In short, this new flow requires users to exchange their `client_token` for an `access_token` by submitting a self-signed JWT attesting to their identity and requesting a specific set of access permissions.

This is critical to our ability to roll out the Open Sandbox and to further or support for interoperability.

It's a big PR, so buckle up enjoy the ride!

**What Changed**

Added a `Token/auth` endpoint, which handles generating an _access token_ from the supplied _client token_, this requires the user to supply a signed JWT in the `client_assertion` query param, in order to get the discharge. We've already added support for uploading public keys in #241. Right now, all of the query params are static, we don't support fine-grained scopes. Only the `system/*:*` scope, which grants all access. That's fine for sandbox, but we'll definitely need to improve that before going full production (shouldn't be a blocker for pilot access).

We do some basic verification of the JWT itself, ensuring that it's not been used before (though our JTI cache is local to the instance, rather than global) and that it hasn't expired. We also verify the signature against the public key that was previously uploaded.

Made some breaking changes to the `dpc-macaroons` module, since our access tokens are now the client_token plus a discharge macaroon, we now need to handle Lists of Macaroons instead of just singletons. Updated all the method signatures and the underlying macaroons library. See: https://github.com/nickrobison-usds/jmacaroons/pull/17.

Public Keys now support (and require) labels, which we can use for matching the `kid` in the JWT header with the corresponding public key in the DB.

Significantly improved our handling of Java errors by cleaning up the way we build and inject our exception handlers. Due to the FHIR requirement to return an `OperationOutcome` in the event of an error, we have to override all of Dropwizard's default exception mappers. This was working in a really hacky manner, but now it's much better. Our custom handlers can inherit from `AbstractFHIRExceptionHandler` and give us control over how the error is reported and generated. We're using this for custom `PersistenceException` and `JerseyViolationException` error messaging.

Smaller cleanups and changes:

Refactored the `auth` package to make it a little cleaner, since we now have a ton of new helpers and classes that we have to keep track of.

I started using IntelliJ for as my git client. Which means all of my commits are now run through their _code analyzer_ so it suggested lots of little fixes and cleanups as I went along. I apologize for the PR noise.

Aligned the ports in the `application.local.conf.sample` file to match what we have in our Docker compose. (also opened the admin ports in the Docker compose file.

**Choices Made**

This PR sets the default token version to `2`, which means it will now generate an unfulfilled third-party caveat for every client token it generates. That forces the user to use the JWT auth flow. However, existing client tokens (e.g. those generated for the sandbox) are version `1` which mean they do not have the third-party caveat and thus are not required to use the JWT auth flow. I made this decision in order to keep support for our existing customers and give them a pathway towards upgrading. Any new tokens generated (e.g. via the new self-service functionality or for production) will require the JWT flow.

**Tickets closed**:

DPC-511: This ticket
DPC-277: Improved handling of Hibernate errors
DPC-736: Fix swagger data models for Keys and Tokens
DPC-738: Public keys now have labels

**Future Work**

DPC-540: Add default logging to exceptions.
DPC-729: Add the ability to remove token root keys from db
DPC-747: Improve available scopes and authorizations
DPC-748: Move JTI cache to Redis
DPC-758: Add SMART .wellknown file
DPC-759: Add public key compliance checks
DPC-760: Add support for ECC keys

**Checklist**


dpc-api/src/main/java/gov/cms/dpc/api/auth/filters/AdminAuthFilter.java
```diff
@@ -46,15 +49,15 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
...
+            this.bakery.verifyMacaroon(m1);
         } catch (BakeryException e) {
             logger.error("Macaroon verification failed", e);
             throw new WebApplicationException(unauthorizedHandler.buildResponse(BEARER_PREFIX, realm));
         }
 
         // Ensure that we don't have any organization IDs
         // Since we ALWAYS generate organization_id caveats for tokens, its absence indicates that its a Golden Macaroon
-        final boolean isGoldenMacaroon = this.bakery.getCaveats(m1)
+        final boolean isGoldenMacaroon = this.bakery.getCaveats(m1.get(0))
```
ronaldheft-usds: Should this be in a try/catch since m1 _theoretically_ could be an empty list?

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/TokenResource.java
```diff
@@ -136,11 +170,99 @@ public Response deleteOrganizationToken(
...
+            logger.error("JWT has invalid signature", e);
+            throw new WebApplicationException(INVALID_JWT_MSG, Response.Status.UNAUTHORIZED);
+        } catch (JwtException e) {
+            logger.error("Malformed JWT", e);
+            throw new WebApplicationException(INVALID_JWT_MSG, Response.Status.UNAUTHORIZED);
+        }
+    }
+
+    private void validateJWTQueryParams(String grantType, String clientAssertionType, String scope) {
+        if (!grantType.equals("client_credentials")) {
```
ronaldheft-usds: Will this line still be reached with the `allowableValues` option in the `@ApiParam` annotation?

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/TokenResource.java
```diff
@@ -136,11 +170,99 @@ public Response deleteOrganizationToken(
...
+    }
+
+    private Macaroon generateMacaroon(TokenPolicy policy, UUID organizationID) {
         // Create some caveats
-        final List<MacaroonCaveat> caveats = List.of(
-                new MacaroonCaveat("", new MacaroonCondition("organization_id", MacaroonCondition.Operator.EQ, organizationID.toString()))
-        );
+        final Duration tokenLifetime = policy
+                .getExpirationPolicy().getExpirationUnit()
+                .getDuration().multipliedBy(policy.getExpirationPolicy().getExpirationOffset());
```
ronaldheft-usds: I might be having a brain fart as I'm deep in this code right now, but is multiplying what we want to do here?

dpc-api/src/main/java/gov/cms/dpc/api/auth/jwt/CaffeineJTICache.java
```diff
@@ -0,0 +1,40 @@
...
+import com.github.benmanes.caffeine.cache.Caffeine;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import java.util.concurrent.TimeUnit;
+
+@Singleton
+public class CaffeineJTICache implements IJTICache {
```
ronaldheft-usds: Something we might want to consider to get this to work across instances is setup an SNS topic and publish the JTI to the topic when it's inserted in the cache. Then other nodes could subscribe to the topic, and when receiving a new JTI, insert it into their own cache.

dpc-api/src/main/java/gov/cms/dpc/api/auth/jwt/JwtKeyResolver.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        final PublicKeyEntity keyByLabel;
+        try {
+             keyByLabel = this.dao.findKeyByLabel(keyId);
+        } catch (NoResultException e) {
+            throw new WebApplicationException(String.format("Cannot find public key with label: %s", keyId), Response.Status.UNAUTHORIZED);
+        }
+
+        // TODO: Should be moved into a helper class
```
ronaldheft-usds: 👀

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/KeyResource.java
```diff
@@ -111,7 +121,13 @@ public PublicKeyEntity submitKey(@ApiParam(hidden = true) @Auth OrganizationPrin
         publicKeyEntity.setOrganization_id(organizationEntity.getId());
         publicKeyEntity.setId(UUID.randomUUID());
         publicKeyEntity.setPublicKey(publicKey);
+        publicKeyEntity.setLabel(keyLabel);
 
         return this.dao.persistPublicKey(publicKeyEntity);
     }
+
+    private String buildDefaultKeyID(UUID organizationID) {
+        final int newKeyID = this.dao.fetchPublicKeys(organizationID).size() + 1;
```
ronaldheft-usds: Are keys ever removed from the DB that a duplicate keyID could be issued? If so, is that an issue?

dpc-common/src/main/java/gov/cms/dpc/fhir/dropwizard/handlers/exceptions/DefaultFHIRExceptionHandler.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.inject.Inject;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.Provider;
+
+/**
+ * Core error handler for differentiating between FHIR and standard HTTP errors.
+ * This overrides *all* of Dropwizard's error handling, but for any non-FHIR resources, we simply delegate back to the root {@link LoggingExceptionMapper}
+ */
+@Provider
+public class DefaultFHIRExceptionHandler extends AbstractFHIRExceptionHandler<Throwable> {
```
ronaldheft-usds: 🎉

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/KeyResource.java
nickrobison-usds: Keys can be removed from the DB. Each key ID needs to be unique to the organization (enforced by a unique constraint) and the key itself need to be unique to the environment. Does that answer your question? Or are you asking how we handle deleting a key and then adding a new one with the same label?

dpc-api/src/main/java/gov/cms/dpc/api/auth/jwt/CaffeineJTICache.java
nickrobison-usds: That's clever! What would be your thoughts about using Redis for this? Would that make things simpler?

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/TokenResource.java
nickrobison-usds: I believe so, because we're apply x number of iterations of a given duration. So in this case it would be a year unit multiplied by the number of years to key is valid (e.g. 1), so the offset would be 1 year.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/TokenResource.java
nickrobison-usds: I think the allowableValues annotation only applies to the Swagger docs, you can still pass anything you want to the actual API endpoint. So I think we do still need the check.

dpc-api/src/main/java/gov/cms/dpc/api/auth/filters/AdminAuthFilter.java
nickrobison-usds: That's a good question. I don't think it's possible for that to happen, but I'll add a check for it.

dpc-api/src/main/java/gov/cms/dpc/api/auth/MacaroonHelpers.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    private static List<CaveatSupplier> generateV1Caveats(Duration tokenLifetime, UUID organizationID) {
+        return generateDefaultCaveats(1, tokenLifetime, organizationID);
+    }
+
+    private static List<CaveatSupplier> generateV2Caveats(Duration tokenLifetime, UUID organizationID) {
+        final List<CaveatSupplier> caveatSuppliers = new ArrayList<>(generateDefaultCaveats(2, tokenLifetime, organizationID));
+        caveatSuppliers.add(() -> new MacaroonCaveat("local", new MacaroonCondition(ORGANIZATION_CAVEAT_KEY, MacaroonCondition.Operator.EQ, organizationID.toString())));
```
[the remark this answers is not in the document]
nickrobison-usds: We just need some payload body in order for the caveat to deserialize correctly. It's just a placeholder, we could use any other value. I'll add a comment to make it clearer.

dpc-api/src/main/java/gov/cms/dpc/api/auth/DPCAuthFilter.java
```diff
@@ -75,22 +73,47 @@ private DPCAuthCredentials validateMacaroon(String macaroon, UriInfo uriInfo) {
...
+            this.bakery.verifyMacaroon(m1, String.format("organization_id = %s", orgID));
         } catch (BakeryException e) {
             logger.error("Macaroon verification failed", e);
             throw new WebApplicationException(unauthorizedHandler.buildResponse(BEARER_PREFIX, realm));
         }
 
         return buildCredentials(macaroon, orgID, uriInfo);
     }
+
+    private UUID extractMacaroonID(List<Macaroon> macaroons) {
```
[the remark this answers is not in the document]
nickrobison-usds: Good call!

dpc-api/src/main/java/gov/cms/dpc/api/auth/DPCAuthFilter.java
```diff
@@ -75,22 +73,47 @@ private DPCAuthCredentials validateMacaroon(String macaroon, UriInfo uriInfo) {
...
+        final Macaroon rootMacaroon = macaroons.get(0);
+        // If we're provided a Golden Macaroon, the ID won't match, so we'll need to actually pull the org_id from the
```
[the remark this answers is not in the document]
nickrobison-usds: Hah! yup. I'll fix it.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/KeyResource.java
ronaldheft-usds: Ok, then it looks there is a bug where a duplicate keyID could be issued. During this flow:

1. Two existing keys exist.
1. New key created (keyID: "key:3")
1. Either of the original two keys is deleted
1. New key created (keyID: "key:3")

We'd end up with a duplicate "key:3" because our ID seed is only the current count of keys in the database.

dpc-api/src/main/java/gov/cms/dpc/api/auth/jwt/CaffeineJTICache.java
ronaldheft-usds: Redis would also be a great solution if we're keeping our Redis cluster around.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/TokenResource.java
ronaldheft-usds: Ah, that makes sense.

dpc-api/src/main/java/gov/cms/dpc/api/auth/filters/AdminAuthFilter.java
nickrobison-usds: Looking at this more, if an empty array is passed to the verifier (on line 52) the verification will fail, so there's no way for us to get an empty array at this point. I've added an assertion to help make that clearer.

dpc-api/src/main/java/gov/cms/dpc/api/auth/jwt/CaffeineJTICache.java
nickrobison-usds: I think so, we'll address this with DPC-742, but I'll set it as a blocker for our beta production release.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/KeyResource.java
nickrobison-usds: Good catch!

nickrobison-usds: I've switched the label generation to use a random number generation. I think this should address the bug. The UI won't let users create public keys without a label, so the only way to trigger the default label generation is to manually create a key via the API and not set the label.

ronaldheft-usds:
There are some test failures to clean up, however, but when tests are passing, I'm good with a merge.

dpc-api/src/main/java/gov/cms/dpc/api/auth/DPCAuthFilter.java
```diff
@@ -49,9 +47,9 @@
     protected abstract DPCAuthCredentials buildCredentials(String macaroon, UUID organizationID, UriInfo uriInfo);
 
     @Override
-    public void filter(final ContainerRequestContext requestContext) throws IOException {
+    public void filter(final ContainerRequestContext requestContext) {
         final UriInfo uriInfo = requestContext.getUriInfo();
-        final String macaroon = AuthHelpers.extractMacaroonFromRequest(requestContext, unauthorizedHandler.buildResponse(BEARER_PREFIX, realm));
+        final String macaroon = MacaroonHelpers.extractMacaroonFromRequest(requestContext, unauthorizedHandler.buildResponse(BEARER_PREFIX, realm));
 
         // If we have a path authorizer, do that, otherwise, continue
```
DeirdreHolub: I can't figure out what this comment is trying to tell me. Are you saying that the method called on the next line does some kind of auth validation that might throw an unauthorized exception?

dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/TokenResourceTest.java
```diff
@@ -131,15 +142,18 @@ void testTokenDeletion() throws IOException {
...
-            httpDelete.addHeader(HttpHeaders.AUTHORIZATION, String.format("Bearer %s", ORGANIZATION_TOKEN));
+            httpDelete.addHeader(HttpHeaders.AUTHORIZATION, String.format("Bearer %s", this.fullyAuthedToken));
             try (CloseableHttpResponse response = client.execute(httpDelete)) {
                 assertEquals(HttpStatus.OK_200, response.getStatusLine().getStatusCode(), "Should have succeeded");
             }
 
 
             // Fetching token should throw an exception
+            // Reauthorizing should fail
+//                assertThrows(InternalServerErrorException.class, () -> APITestHelpers.jwtAuthFlow(getBaseURL(), ORGANIZATION_TOKEN, KEY_ID, privateKey));
```
DeirdreHolub: Why is this assertion commented out?

dpc-api/src/main/java/gov/cms/dpc/api/auth/DPCAuthFilter.java
nickrobison-usds: That's an old comment that should have come out. Good catch, thanks!

dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/TokenResourceTest.java
nickrobison-usds: It was replaced by the test right below it, which correctly checked for missing tokens, it should have been removed, and it now is!

nickrobison-usds:
@DeirdreHolub Thanks for the review!

Regarding the .wellknown endpoint, I've created DPC-758 to track adding that file, we definitely should have one, even though we're not really a SMART app.

The public key enforcements are waiting for another PR, I think I have a ticket tracking it, otherwise, DPC-759 now is. Fortunately, the JWT library will fail when trying to verify a JWT with a key that doesn't meet the length requirements, so it's not a security risk, just a poor user experience.

I think I addressed your previous comments, which are just poorly cleaned up code sections.

PR #324, opened 2019-10-24
DPC-726: Add Organization/$submit endpoint to API service

nickrobison-usds:
**Why**

I suddenly realized that we had never pulled the Organization creation logic into the API service, which meant the website didn't have the ability to actually create orgs, only do things with them once they have been created. So we needed to fix that.

**What Changed**

Added the `$submit` operation to the API service, which works just like the corresponding operation in the Attribution service.

The trick here is that we only want Golden Macaroons to be able to create orgs. To solve that, I created an `AdminOperation` annotation, which checks the incoming Macaroon to see if it has an `organization_id` caveat present. Since we ALWAYS create org tokens with that caveat, its absence indicates it's really a Golden Macaroon and should be allowed to shine.

This is verified by trying to create an org as both an admin, and as a normal organization.

**Choices Made**

Very simple changes right now, as this is blocking the website development. I was able to re-use most of the existing auth and Organization logic, so the actual set of changes is really small. A lot of it is simply refactoring common functions into helper classes.

The actual endpoint is not shown in Swagger, in order to not tip our hand too much.

**Tickets closed**:

DPC-726: Create organizations from the API service.

**Future Work**

DPC-729:  Add the ability to manually remove tokens from the database.

**Checklist**


nickrobison-usds:
@ronaldheft-usds @DeirdreHolub this PR is ready to review, a couple of loose ends to cleanup, but the core logic is in place. Thanks!

dpc-api/src/main/java/gov/cms/dpc/api/auth/AdminAuthFilter.java
```diff
@@ -0,0 +1,68 @@
...
+            logger.error("Macaroon verification failed", e);
+            throw new WebApplicationException(unauthorizedHandler.buildResponse(BEARER_PREFIX, realm));
+        }
+
+        // Ensure that we don't have any organization IDs
+        // Since we ALWAYS generate organization_id caveats for tokens, its absence indicates that its a Golden Macaroon
+        final boolean isGoldenMacaroon = this.bakery.getCaveats(m1)
+                .stream()
+                .map(MacaroonCaveat::getCondition)
+                .anyMatch(cond -> cond.getKey().equals("organization_id"));
```
ronaldheft-usds: Apologies if I'm missing something in how we create golden macaroons, but from my understand we're only going to create a golden macaroon by calling the attribution service directly, correct? We then manually configure the website with the golden macaroon.

On the client side, the website will call the API to create a macaroon using the golden macaroon, but the API will never return a macaroon without an organization_id assigned to it.

If that's the case, could be enhance the security a bit by adding an extra check to the OrganizationRegistration cli command to verify the token returned back has an organization_id. It doesn't look like there is any path in the code currently to accidentally create a golden macaroon, but I would feel safer if we have that extra check if something unintentionally breaks in the future.

nickrobison-usds: That's a good suggestion! The only way to create a golden macaroon is to call the `generate-token` admin task without an organization ID, but it makes sense to ensure that we never create one by accident.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/TokenResource.java
```diff
@@ -99,6 +99,10 @@ public String createOrganizationToken(
 
         final Macaroon macaroon = generateMacaroon(organizationID);
 
+        // Ensure that each generated Macaroon has an associated Organization ID
+        // This way we check to make sure we never generate a Golden Macaroon
+        ensureOrganizationPresent(macaroon);
```
ronaldheft-usds: Yay! This is an awesome addition!

PR #330, opened 2019-10-25
DPC-727: Fix admin endpoints and smoke tests

nickrobison-usds:
**Why**

During our deployments on Wednesday, we can into some issues with the admin tasks not being reachable by client code. This turned out to be operator error with specifying the wrong base URL for the CLI commands.

While digging into this, discovered that the SmokeTests are actually failing and not reporting the status as such. Turns out, JMeter always succeeds, you have to `grep` the summary report to determine if a failure did indeed occur.

Also realized that the fix we applied in #326 was only masking the root issue, which was that our public URL was set to `http://`, when adding the port, that caused the redirection to fail.

**What Changed**

Updated the CLI commands to be explicit about what URLs are required (e.g. admin commands need to have `/task` in the URL. Also set a couple of parameters to be required, as they should have been all along.

Reworked some of the smoke tests, they had rotted after a series of changes, so I think they're all updated. They also now return a status code if any of the tests fail. It's a really clunky hack, so we might want to consider [this Jenkins plugin](https://www.blazemeter.com/blog/how-to-use-the-jenkins-performance-plugin/).

**Choices Made**

Duplicated some of the helper code until #323 is merged.

**Tickets closed**:

DPC-727

**Future Work**

DPC-511: Refactor the helper code for creating organizations and macaroons into the `dpc-testing` 
module.

**Checklist**


nickrobison-usds:
@ronaldheft-usds Taurus is merged, great find!

What's the processing for ensuring that `virtualenv` is installed on the Jenkins box? Add it to the ansible build?

Also, the smoke tests will fail in Dev until we get our certificate from EnTrust (which should hopefully be in the next couple of days), we can either disable strict SSL checking in the HAPI client, or disable them in Dev. Preferred approach?

ronaldheft-usds:
@nickrobison-usds That's an awesome addition with Taurus!

We can add it to Jenkins by [installing the dependency in the Docker file here](https://github.com/CMSgov/dpc-ops/blob/master/playbooks/jenkins/files/Dockerfile#L34), then running a packer build to generate a new AMI and deploying out the management module.

I would vote just letting the smoke tests fail if it's only going to be a couple of days.

PR #343, opened 2019-10-30
cleanup layout and views

jaredcunha-usds:
**Why**

Most of the page designs are moving towards the type of layout that's been used for internal users. This work will clean up some of the existing page designs to reflect this.

**What Changed**

Some pages will reflect the new layout design

Screen Shot 2019-10-30 at 1 16 21 PM

Screen Shot 2019-10-30 at 1 16 29 PM

Screen Shot 2019-10-30 at 1 16 42 PM

**Choices Made**

- Created a new layout file for table indexes 
- Removed `internal_user?` helper method and related test

**Tickets closed**:

- [DPC-666](https://jiraent.cms.gov/browse/DPC-666) AKA "The Bug From Hell"


dpc-web/app/controllers/internal/organizations_controller.rb
```diff
@@ -71,5 +72,14 @@ def organization_params
         address_attributes: %i[id street street_2 city state zip]
       )
     end
+
+    def resolve_layout
```
jaredcunha-usds: I have this private method defined in two places. @switzersc-usds do you prefer this goes into application controller?

dpc-web/app/controllers/internal/users_controller.rb
```diff
@@ -43,5 +44,14 @@ def download
     def user_params
       params.fetch(:user).permit(:first_name, :last_name, :email, :organization_ids)
     end
+
+    def resolve_layout
```
switzersc-usds: Since there's only one index action per controller, I think this adds a layer of complexity and code duplication that maybe isn't necessary. Instead, you should be able to add `render layout: 'table_index'` to the `index` action itself after assigning the `@users` variable.

jaredcunha-usds: Good call. This is done.

PR #345, opened 2019-10-30
DPC-653: Create custom Docker image for services

rnagle:
**Why**

Rather than installing common dependencies via each service's `entrypoint.sh`, we want to build a baseline image with said dependencies that individual service images can be based on.

Assuming the base image changes infrequently, this should mean marginally faster build times for individual services. It also reduces external requirements for services to start successfully.

**What Changed**

- Added `docker-compose.base.yml` and `docker/Dockerfiles/Dockerfile.base`
- `Dockerfile.base` builds from the `openjdk:11-jdk-slim` (same that was previously used in `pom.xml` as the base for jib builds)
- Moved `apt-get` commands out of api and aggregation `entrypoint.sh` files
- Added `make docker-base`, made `docker-base` part of `make ci-app` target to ensure the base image is available when jib builds images for services
- Updated base config for Docker images in `pom.xml` to use `docker://dpc-base:lastest` (uses the image from local Docker daemon)
- Updated `README.md`

**Choices Made**

Probably the most notable choice was to use local Docker daemon to source `dpc-base:latest` when jib builds images, rather than shouldering the additional complexity of pushing to/pulling from a remote repo (e.g., Docker Hub or ECR).

**Tickets closed**:

https://jiraent.cms.gov/browse/DPC-653

**Future Work**

None planned at present.

**Checklist**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -400,7 +400,7 @@
                             </tags>
                         </to>
                         <from>
-                            <image>docker://755619740999.dkr.ecr.us-east-1.amazonaws.com/dpc-base</image>
+                            <image>docker://dpc-base:latest</image>
                         </from>
                     </configuration>
                     <executions>
--- a/pom.xml
+++ b/pom.xml
@@ -400,7 +400,7 @@
                             </tags>
                         </to>
                         <from>
-                            <image>openjdk:11-jdk-slim</image>
+                            <image>docker://755619740999.dkr.ecr.us-east-1.amazonaws.com/dpc-base</image>
                         </from>
                     </configuration>
                     <executions>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -400,7 +400,7 @@
                             </tags>
                         </to>
                         <from>
-                            <image>openjdk:11-jdk-slim</image>
+                            <image>docker://755619740999.dkr.ecr.us-east-1.amazonaws.com/dpc-base</image>
```
rnagle: Using the `docker://` prefix lets us use the image from the local Docker daemon. We might not have to push this to a remote repo, in which case we can change the name/label. Anyone have strong feelings on this?

nickrobison-usds: Sounds good to me! We'll just need to make sure we note it in the README.

PR #339, opened 2019-10-29
DPC-693: 1-800-MEDICARE ETL

embh:
**Why**
This PR sets up the 1-800-MEDICARE file ETL process that will allow DPC to provide Consent resources.

**What Changed**
* `SuppressionFileImport` job added to the consent module and supporting `SuppressionFileUtils`
* `ConsentEntity` changes:
  * Added `bfdPatientId` for the patient ID used by BFD (For discussion: is this necessary or is MBI enough?)
  * Modified `effectiveDate` to be a date rather than a timestamp, to match the 1-800 data
* New `JobTestUtils` for shared `startJob()`, `stopJob()`, and `resetScheduler()` test helpers
* Created four synthetic 1-800 files to be used by `SuppressionFileImportTest`
* `suppressionFileDir` can be set in config files

**Choices Made**
Getting a non-HICN ID from BFD is required for consent records. To limit the scope of work in this PR, that part has been separated into a new ticket, DPC-740. 

**Tickets closed**:
[DPC-693](https://jiraent.cms.gov/browse/DPC-693): 1-800-MEDICARE/Consent ETL

**Future Work**
[DPC-740](https://jiraent.cms.gov/browse/DPC-740): Integrate BFD request for MBI into 1-800-MEDICARE ETL
[DPC-746](https://jiraent.cms.gov/browse/DPC-746): Create audit records for 1-800-MEDICARE ETL

**Checklist**

- [x] All tests are passing via `make ci-app` (app change) and `make ci-web` (website change)
- [x] Swagger documentation has been updated (N/A)
- [x] FHIR documentation has been updated (N/A)
- [ ] Any required dpc-ops changes have a PR submitted and mentioned in this ticket (N/A)
- [x] Any manual migration steps are documented, scripts written (where applicable), and tested
- [ ] Before merging, any required dpc-ops changes have been approved and merged into master of the dpc-ops repo (N/A)

PR #359, opened 2019-11-05
DPC-690: API for Consent Service

DeirdreHolub:
**Why**

This PR exposes the Consent Service for use by other internal services only. There are open spec and code questions, discussed in **Choices Made**.

The original specification for behavior can be found [here](https://confluenceent.cms.gov/display/DAPC/1-800+MEDICARE+Suppression+Data).

**Example Resource**
```
{
  "resourceType": "Consent",
  "id": "3a51c388-a9b0-42e2-afe4-688a2b6cef95",
  "text": {
    "status": "generated",
    "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\">Consent status of OPTIN for the patient with identifiers [http://bluebutton.cms.hhs.gov/identifier#hicnHash|hicn_1 ], [https://bluebutton.cms.gov/resources/variables/bene_id|mbi_1 ]</div>"
  },
  "status": "active",
  "category": [
    {
      "coding": [
        {
          "system": "http://loinc.org",
          "code": "64292-6",
          "display": "Release of information consent Document"
        }
      ]
    }
  ],
  "patient": {
    "reference": "http://localhost:3200/api/v1/Patient?identity=|mbi_1"
  },
  "dateTime": "2019-10-28T18:00:00-06:00",
  "organization": [
    {
      "reference": "https://localhost",
      "display": "Data at the Point of Care"
    }
  ],
  "policyRule": "http://hl7.org/fhir/ConsentPolicy/opt-in"
}
```
_Note: Using logical patient ref until we decide whether ETL or API will look up UUID._

**What Changed**

- Consent service endpoints added
- SeedCommand added to load consent records from a csv file

**Choices Made**

The specification was written based on the R4 version; this PR implements the R3 version of the Consent Resource. The differences between the two are documented in the spec page referenced above. Given the minor differences and that most of the data is fixed, I think using the R3 version works.

I kept the ConsentResource in one file, rather than following the pattern in other APIs because this API has so few methods. However, that may be an annoying dissonance. 

**Tickets closed**:

**Future Work**

**Checklist**


dpc-common/src/main/java/gov/cms/dpc/common/consent/entities/ConsentEntity.java
```diff
@@ -8,28 +8,48 @@
...
+    public static final String LOINC_CATEGORY = "64292-6";
+    public static final String OPT_IN = "OPTIN";
+    public static final String OPT_OUT = "OPTOUT";
+    public static final String TREATMENT = "TREAT";
+    public static final String SCOPE_CODE = "patient-privacy";
 
-    private static final long serialVersionUID = 42L;
+    private static final long serialVersionUID = 8702499693412507926L;
+
+    public ConsentEntity() { }
```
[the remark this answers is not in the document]
nickrobison-usds: I usually just add `//Hibernate required` to the annotation, to make it clear why it's there.

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/entities/ConsentEntityConverter.java
```diff
@@ -0,0 +1,75 @@
...
+    private static final String OPT_OUT = "OPTOUT";
+
+    private ConsentEntityConverter() {}
+
+    private static String patientIdentifier(DPCIdentifierSystem type, String value) {
+        return String.format("%s|%s ", type.getSystem(), value);
+    }
+
+    private static Narrative narrativeText(String inOrOut, String hicn, String mbi) {
+        StringBuilder sb = new StringBuilder("Words about the ");
```
nickrobison-usds: Should we change this to something like _consent status for patient_?

dpc-consent/src/main/java/gov/cms/dpc/consent/resources/ConsentResource.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @GET
+    @Path("/Consent")
+    @FHIR
+    @UnitOfWork
+    @ApiOperation(value = "Search for Consent Entries", notes = "Search for Consent records. " +
+            "<p>Must provide ONE OF Consent ID as an _id or identifier, or a patient MBI or HICN to search for.")
+    @ApiResponses(@ApiResponse(code = 400, message = "Must provide Consent or Patient id"))
+    public Bundle search(
+            @ApiParam(value = "Consent resource _id") @QueryParam("_id") Optional<UUID> id,
```
nickrobison-usds: Can we use the search param constants from the `Consent` class?

dpc-consent/src/main/java/gov/cms/dpc/consent/resources/ConsentResource.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private final ConsentDAO dao;
+
+    @Inject
+    ConsentResource(ConsentDAO dao) {
+        this.dao = dao;
+    }
+
+    @GET
+    @Path("/Consent")
```
nickrobison-usds: This should just be GET on the root path.

dpc-consent/src/main/java/gov/cms/dpc/consent/resources/ConsentResource.java
```diff
@@ -0,0 +1,132 @@
...
+    @FHIR
+    @UnitOfWork
+    @ApiOperation(value = "Search for Consent Entries", notes = "Search for Consent records. " +
+            "<p>Must provide ONE OF Consent ID as an _id or identifier, or a patient MBI or HICN to search for.")
+    @ApiResponses(@ApiResponse(code = 400, message = "Must provide Consent or Patient id"))
+    public Bundle search(
```
nickrobison-usds: These operations should had metric annotations, like the `dpc-attribution` resources.

dpc-consent/src/main/java/gov/cms/dpc/consent/resources/ConsentResource.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private Bundle bundleOf(List<ConsentEntity> consentEntities) {
+
+        if (consentEntities.isEmpty()) {
+            throw new WebApplicationException("Cannot find patient with given ID", javax.ws.rs.core.Response.Status.NOT_FOUND);
+        }
+
+        Bundle bundle = new Bundle();
+        bundle.setType(Bundle.BundleType.COLLECTION);
```
nickrobison-usds: This should be a `Search` bundle

dpc-consent/src/main/resources/banner.txt
```diff
@@ -0,0 +1,9 @@
+                                                    (    (     
+   (                                   )     (      )\ ) )\ )  
+   )\                     (         ( /(     )\    (()/((()/(  
+ (((_)   (    (     (    ))\  (     )\()) ((((_)(   /(_))/(_)) 
+ )\___   )\   )\ )  )\  /((_) )\ ) (_))/   )\ _ )\ (_)) (_))   
+((/ __| ((_) _(_/( ((_)(_))  _(_/( | |_    (_)_\(_)| _ \|_ _|  
+ | (__ / _ \| ' \))(_-</ -_)| ' \))|  _|    / _ \  |  _/ | |   
+  \___|\___/|_||_| /__/\___||_||_|  \__|   /_/ \_\ |_|  |___|
```
nickrobison-usds: Awesome!

dpc-consent/src/main/resources/application.conf
```diff
@@ -5,4 +5,7 @@ dpc.consent {
     title = "DPC Consent Service"
     resourcePackage = gov.cms.dpc.consent.resources
   }
+
+  // Disable FHIR validation
+  fhir.validation.enabled = false
```
nickrobison-usds: Why is validation disabled?

DeirdreHolub: I used attribution's conf as the model to follow, since it is also a backend service. It's validation is also disabled.

docker-compose.yml
```diff
@@ -57,9 +57,14 @@ services:
...
   consent:
     image: ${ECR_HOST:-755619740999.dkr.ecr.us-east-1.amazonaws.com/dpc-consent}:latest
+    depends_on:
+      - db
     environment:
       - ENV=local
       - JACOCO=${REPORT_COVERAGE}
+    ports:
+      - "3333:3002"
+      - "9999:9900"
```
nickrobison-usds: Can we do 9904 for this port? To keep it consistent with the other services.

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/entities/ConsentEntityConverter.java
```diff
@@ -0,0 +1,80 @@
...
+        if (OPT_OUT.equals(value) || OPT_IN.equals(value)) {
+            return value;
+        }
+        throw new IllegalArgumentException(String.format("invalid policyCode %s", value));
+    }
+
+    private static List<CodeableConcept> category(String loincCode) {
+        // there must be a way to look up the code systems used in these CodeableConcept values. what is it?
+        CodeableConcept category = new CodeableConcept();
+        category.addCoding().setSystem("http://loinc.org").setCode(loincCode);
```
nickrobison-usds: Can we set the `display` value as well? To avoid forcing the user to memorize, or lookup, LOINC codes? The display value can be whatever we want it to be.

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/entities/ConsentEntityConverter.java
```diff
@@ -0,0 +1,80 @@
...
+
+        c.setId(consentEntity.getId().toString());
+
+        // there is no consent status in entity, so we are defaulting to active. Correct?
+        c.setStatus(Consent.ConsentState.ACTIVE);
+        c.setDateTime(Date.from(consentEntity.getEffectiveDate().toInstant()));
+        c.setText(narrativeText(consentEntity.getPolicyCode(), consentEntity.getHicn(), consentEntity.getMbi()));
+
+        // hicn or mbi? both?
+        c.setPatient(new Reference(new IdType("Patient", consentEntity.getMbi())));
```
nickrobison-usds: The reference has to point to the ID of the Patient resource, not the business identifier. So this would be the UUID primary key of the PatientEntity.

dpc-consent/pom.xml
```diff
@@ -14,6 +14,7 @@
 
     <properties>
         <mainClass>gov.cms.dpc.consent.DPCConsentService</mainClass>
+        <jooq.version>3.12.1</jooq.version>
```
nickrobison-usds: Now that we have 2 services using JOOQ (attribution as well), can we pull this variable up into the root pom?

dpc-common/src/main/java/gov/cms/dpc/common/consent/entities/ConsentEntity.java
```diff
@@ -8,28 +8,50 @@
...
 import java.io.Serializable;
 import java.time.OffsetDateTime;
+import java.time.ZoneId;
+import java.util.Optional;
 import java.util.UUID;
 
-@Entity
+@Entity(name = "consent")
 public class ConsentEntity implements Serializable {
+    public static final String LOINC_CATEGORY = "64292-6";
```
embh: What is meant by `CATEGORY` here? Could it be `CODE` to match how LOINC describes the value? https://loinc.org/64292-6/

dpc-common/src/main/java/gov/cms/dpc/fhir/DPCIdentifierSystem.java
```diff
@@ -11,6 +11,7 @@
     NPPES("http://hl7.org/fhir/sid/us-npi"),
     MBI("https://bluebutton.cms.gov/resources/variables/bene_id"),
     DPC("https://dpc.cms.gov/organization_id#"),
+    HICN("http://bluebutton.cms.hhs.gov/identifier#hicnHash"),
```
embh: Should we have both `HICN` and `HICN_HASH`? The unhashed HICN system is `http://hl7.org/fhir/sid/us-medicare`: https://github.com/CMSgov/beneficiary-fhir-data/blob/master/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerConstants.java#L205

@nickrobison-usds @ronaldheft-usds The `MBI` value is referencing the BFD patient ID system; MBI is `http://hl7.org/fhir/sid/us-mbi`. https://github.com/CMSgov/beneficiary-fhir-data/blob/master/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerConstants.java#L212 Is it set that way intentionally?

nickrobison-usds: @embh It was set that way intentionally, per this [ticket](https://jiraent.cms.gov/browse/DPC-529). If that's not correct, we'll need to make the change, but it will be a breaking change, so we'll need to go through the correct process.

Could you open a ticket to track this question?

embh: That's interesting. I can't see the ticket it links to, but the value it changed from is also BFD's own patient ID, AFAIK. New ticket here: https://jiraent.cms.gov/browse/DPC-762

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/entities/ConsentEntityConverter.java
DeirdreHolub: Just to be clear, you're saying this has to be the UUID of the DPC PatientResource? In that case, I think we should update the ETL to add that to the table during import; otherwise, we'll have to query for it from attribution with every request.

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/entities/ConsentEntityConverter.java
DeirdreHolub: The official short name is 'Release of information consent'. Does that work?

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/entities/ConsentEntityConverter.java
nickrobison-usds: Yes, it has to be the UUID of the DPC PatientResource, we can modify the ETL process or pull that in, or fetch it from the attribution service when we need it.

Another option would be to use a logical id, which would be something like (https://dpc.cms.gov/api/v1/Patient?identifier=|{patient MBI}). I don't really like that, but it is possible.

dpc-common/src/main/java/gov/cms/dpc/common/consent/entities/ConsentEntity.java
DeirdreHolub: How about CATEGORY_LOINC_CODE? I'm trying to anchor the name to its purpose in this resource. In other words, this isn't a general purpose constant.

docker-compose.yml
DeirdreHolub: 9904 and 3600 to be consistent in both ports.

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/entities/ConsentEntityConverter.java
embh: Would the 1-800 ETL use the dpc_attribution database directly? If not, is there a way to retrieve a patient by MBI or BFD ID in the attribution service?

DeirdreHolub: Does this also imply that the patient id they submit might be a UUID or a reference? (The spec  doesn't anticipate that.)

DeirdreHolub:
Open issue from spec: what organization id should we use?

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/entities/ConsentEntityConverter.java
DeirdreHolub: > Yes, it has to be the UUID of the DPC PatientResource, we can modify the ETL process or pull that in, or fetch it from the attribution service when we need it.
> 
> Another option would be to use a logical id, which would be something like ([https://dpc.cms.gov/api/v1/Patient?identifier=|{patient](https://dpc.cms.gov/api/v1/Patient?identifier=%7C%7Bpatient) MBI}). I don't really like that, but it is possible.

We're already using identifier as an alternate way to submit the consent record identifier. I'm assuming that we can't use it for both purposes.

DeirdreHolub: > Would the 1-800 ETL use the dpc_attribution database directly? If not, is there a way to retrieve a patient by MBI or BFD ID in the attribution service?

a GET on /Patient?identifier=<mbi goes here>

nickrobison-usds: @DeirdreHolub You probably solved this by now, but a Reference can be a logical reference, which can be the query string necessary in order to return the resource or a bundle containing a resource. Using the IdType constructor won't work correctly, but I believe there's another overload that will.

I'm not sure I'm following regarding the duplicate patient identifier.

nickrobison-usds:
> Open issue from spec: what organization id should we use?

Per the spec, this should be the organization which owns the actual resource itself. In this case, it would be a Reference to https://dpc.cms.gov. Eventually, this may change to be BFD.

nickrobison-usds:
@DeirdreHolub Looks good! I merged #339, so once we get the conflicts resolved, we can merge it. Thanks!

nickrobison-usds:
@DeirdreHolub Could you please open a follow-on ticket for creating a FHIR profile and updating our IG?

PR #365, opened 2019-11-07
make providers current when provider orgs selected

jaredcunha-usds:
**Why**

The providers left nav dropdown should use the `current_class` when viewing organizations 

**What Changed**

Screen Shot 2019-11-07 at 8 25 27 AM


dpc-web/db/schema.rb
```diff
@@ -98,6 +90,15 @@
...
 
+  create_table "profile_endpoints", force: :cascade do |t|
+    t.string "name", null: false
+    t.integer "status", null: false
+    t.integer "connection_type", null: false
+    t.string "uri", null: false
+    t.integer "organization_id", null: false
+    t.index ["organization_id"], name: "index_profile_endpoints_on_organization_id"
+  end
+
```
jaredcunha-usds: Not sure if this file matters. looks like the block was just moved during a migration.

jaredcunha-usds: Ran a migration, but i don't know. should this be committed?

dpc-web/app/views/shared/internal/_sidenav.html.erb
```diff
@@ -11,7 +11,7 @@
 
 <ul class="ds-c-vertical-nav">
   <li class="ds-c-vertical-nav__item usa-accordion">
-    <button class="usa-accordion__button ds-c-vertical-nav__label ds-c-vertical-nav__label--parent <%= current_sidenav_class?(:users, current) %>" aria-controls="nav-providers-main" aria-expanded="true">
+    <button class="usa-accordion__button ds-c-vertical-nav__label ds-c-vertical-nav__label--parent <%= current_sidenav_class?(:users, current) %> <%= current_sidenav_class?(:organizations, current) %>" aria-controls="nav-providers-main" aria-expanded="true">
```
jaredcunha-usds: Sort of a lazy approach, but wasn't sure if we should build more complexity into the helper method. I'm not a ruby expert, obvs, but is there a more concise way of passing  `:users` **OR** `:organizations` into the argument?

PR #376, opened 2019-11-12
DPC-787: JobCompletionModel extensions should have qualifiers

embh:
**Why**
Keys in the `extension` field in job creation model entries should be namespaced.

**What Changed**
Extension format has been updated to match FHIR specs as seen [here](https://www.hl7.org/fhir/STU3/extensibility-definitions.html), with the following URLs:
* `checksum` -> `https://dpc.cms.gov/checksum`
* `length` -> `https://dpc.cms.gov/file_length`

**Choices Made**
I chose to use URLs rather than reverse domain name notation, to be consistent with Blue Button/Beneficiary FHIR Data Server. For the values, I used `valueString` for the checksum and `valueDecimal` for the file length, based on [FHIR datatypes](https://www.hl7.org/fhir/STU3/datatypes.html). ~`BigDecimal` was used for the decimal value as suggested in the Notes section of that page.~ The decimal value type is `long`, which is what was used previously for the file length field.

**Tickets closed**:
[DPC-787](https://jiraent.cms.gov/browse/DPC-787)

**Future Work**
Extensions need to be added to FHIR IG; this will be a separate task.
The [FHIR Extensibility](https://www.hl7.org/fhir/STU3/extensibility.html) page says:
> The StructureDefinition that defines the extension SHOULD be available at the identified location so that resource processors can access the definition from the URL.

So we should consider creating those as well. Should they be part of the implementation guide, with URLs changed to the expected paths?

**Checklist**

- [x] FHIR documentation has been updated - see Future Work above


dpc-common/src/main/java/gov/cms/dpc/common/models/JobCompletionModel.java
```diff
@@ -62,40 +67,40 @@ public Integer getCount() {
...
 
-        public String getChecksum() {
-            return checksum;
+        public String getValueString() {
+            return valueString;
         }
 
-        public Long getLength() {
-            return length;
+        public BigDecimal getValueDecimal() {
```
nickrobison-usds: Even though we're using FHIR decimal, let's keep this value as a long, the decimal allows for integer values as well.

PR #379, opened 2019-11-13
DPC-593: As a user, I want to create client tokens in the dashboard.

switzersc-usds:
Why

Users need to create client tokens to use to authenticate with our API.

What Changed

* Changes Dashboard view for signed in users
* Adds UI for creating and viewing client tokens

Create a token:
Screen Shot 2019-11-14 at 11 45 09 AM

Screen Shot 2019-11-14 at 10 44 23 AM

View tokens (direct GET from API):
Screen Shot 2019-11-14 at 11 43 40 AM

Choices Made

* Removes DpcRegistration model: not sure what the original intent was for this, and I'm not using it for anything
* Does not offer delete token functionality yet - not needed for sandbox

Tickets closed:

DPC-593

Future Work

* delete functionality
* public keys


dpc-web/app/services/api_client.rb
```diff
@@ -32,6 +46,16 @@ def auth_header(token)
     { 'Authorization': "Bearer #{token}" }
   end
 
+  def delegated_macaroon(reg_org_id)
+    # Temp fix - the macaroon shouldn't need to be decoded
```
nickrobison-usds: I believe this was fixed in ea3d6cc5a8f0f8c1dd79e0260e9db6f8ca4a9cd9, which I merged last night. Can you confirm?

dpc-web/app/views/client_tokens/new.html.erb
```diff
@@ -0,0 +1,32 @@
...
+  <section class="box">
+    <div class="box__content">
+
+      <h2 class="box__heading">New token for <%= @organization.name %></h2>
+      <%= form_tag organization_client_tokens_path(organization_id: params[:organization_id]), method: :post do %>
+
+        <div class="field">
+          <%= label_tag :label, "Label", class: "ds-c-label" %>
+          <%= text_field_tag :label, params[:label], class: "ds-c-field" %>
+          <p class="ds-c-field__hint">Choose a descriptive name that includes environment, such as “EMR synthetic sandbox”</p>
```
nickrobison-usds: The API service limits this text to 25 characters. Would it be simple to impose that limit here? Otherwise, the API _should_ return a useable error message.

dpc-web/app/services/api_client.rb
switzersc-usds: oh! will do!

dpc-web/app/views/client_tokens/new.html.erb
switzersc-usds: yeah we can add a validation!

dpc-web/app/services/api_client.rb
switzersc-usds: correct, no longer needed! thanks!

switzersc-usds:
@nickrobison-usds I have updated the PR to remove the token decoding and to add a max length to the client token label input field

PR #390, opened 2019-11-14
Add some simple DB seeds to speed up development

switzersc-usds:
**Why**

Right now when we are developing the app, if we need to reset the database, we have to go through the process of creating users and organizations all over again. 

**What Changed**

This PR adds seeds that create an assigned user and an unassigned user, and the unassigned user also has a tag. This should be enough for getting basic dev up and running for UI changes or QA. 

**Choices Made**

Adds seeds. The seeds file gets run whenever you run `rails db:reset` in terminal.


switzersc-usds:
@jaredcunha-usds let me know if you think there's anything else we should add to the seeds in this PR. 

I hesitate to add things related to the API in the seeds because it might create more confusion and problems than it's worth.

PR #393, opened 2019-11-15
DPC-817: JobCompletionModel extension fields should not be included if no value

embh:
**Why**
In job response extensions, only fields with values should be included. There was a bug in https://github.com/CMSgov/dpc-app/pull/376 due to changing the `valueDecimal` type to `long`.

**What Changed**
Changed `valueDecimal` to `Long` instead, so it can be null and thus ignored when converted to JSON. Postman tests updated to confirm that each object in the `extension` array has only two fields.

**Tickets closed**:
[DPC-817](https://jiraent.cms.gov/browse/DPC-817)

**Checklist**


nickrobison-usds:
Why would there be null values here?

embh:
> Why would there be null values here?

For the checksum object, `valueDecimal` isn't set, so it shouldn't be included in the JSON. As a `long`, it ends up with a value of `0`.

nickrobison-usds:
@embh Ah, I see now. Thanks!

PR #394, opened 2019-11-15
DPC-815: Aggregator crashes / fails to starts on invalid job

ronaldheft-usds:
**Why**

We are seeing the aggregator fail to start when the job status is queued (0) with a claimed aggregatorID.

This should not occur. The aggregator should skip invalid jobs, or at the very least, fail them into a failed state and then resume claiming new jobs.

**What Changed**

* The aggregator will now resubscribe to the queue if kicked out via an exception
* The queues will mark job an invalid job as failed if claimed (so they don't keep reclaiming invalid jobs)

**Choices Made**

* There were protections in place when marking a job as failed to ensure the same aggregator is  doing so. I removed these protections because a job should never get stuck and unable to be marked as failed. The other methods still check the aggregator ids before changing status.

**Tickets closed**:

DPC-815: Aggregator fails to start when an aggregator job is an invalid state

**Future Work**

DPC-814: The underlying cause for how these invalid jobs ended up in the queue

**Checklist**


dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
```diff
@@ -107,7 +107,10 @@ private void pollQueue() {
                     logger.debug(String.format("No job, polling again in %d milliseconds", operationsConfig.getPollingFrequency()));
                     return completed.delay(operationsConfig.getPollingFrequency(), TimeUnit.MILLISECONDS);
                 })
-                .subscribe(this::processJobBatch, error -> logger.error("Unable to complete job.", error));
+                .subscribe(this::processJobBatch, error -> {
+                    logger.error("Unable to complete job.", error);
+                    this.pollQueue(); // Continue polling the queue
```
nickrobison-usds: Probably a dumb question, but will this recursive call potentially blow out the stack? Is it possible for us to get a couple of days in and suddenly get a StackOverflow?

If so, still happy to merge this in, let's just get a ticket open to track it.

ronaldheft-usds: Great point about the StackOverflow! That would definitely happen if we recovering from exceptions on a regular. I read over the RxJava docs more and figured out how to keep the queue running in an RxJava way.

ronaldheft-usds:
Added an AtomicBoolean to track if the queue is running, which will come in handy for DPC-816.

PR #395, opened 2019-11-15
DPC-824: Fix fragile test strategy for synthetic 1-800 files

embh:
**Why**
The tests for the 1-800-MEDICARE ETL involve importing synthetic files. Because the ETL cleans up the import directory, the synthetic files need to be copied before the tests are run to preserve the originals. A `/copy` directory with only a `.gitignore` file was created for that purpose, but the `.gitignore` file ends up deleted after tests run, causing a bit of a mess when looking at changes in Git.

**What Changed**
The `/copy` directory in `src/test/resources/synthetic-1800-files` is now created and deleted in the testing process.

**Tickets closed**:
[DPC-824](https://jiraent.cms.gov/browse/DPC-824)

**Checklist**


embh:
> I'm not sure I understand this PR, is it possible to use the maven-dependencies plugin to copy the necessary test files on each invocation? Rather than setting up a manual copy process?

One test expects a set of valid files and the other uses invalid files, so they would have to be copied into separate directories rather than reusing the `/copy` path. As a result, the injector for the input directory (`suppressionFileDir`) would have to be reconfigured per test. Would that be preferable?

nickrobison-usds:
@embh No, I think that's fine then. Thanks for clarifying!

PR #397, opened 2019-11-15
DPC-721: Enable internal user to create org from user page

switzersc-usds:
**Why**

It's going to take internal users ages to create organizations with the current user experience, so they have requested the ability to create orgs from the user's input information.

**What Changed**

If a user requests an org and that org doesn't exist, the internal user
managing that user can create an organization directly from the user's
edit page. The internal user clicks a link and then the new organization
form is populated on the organziations#new page with the user's
requested org info, so the internal user can modify or add anything as
needed. Then the internal user saves the org and gets redirected back to
the user's edit page and they can carry on from there.

**Choices Made**

This utilizes the pages we already have by pre-populating forms and redirecting back to the user page, rather than create new pages or service objects to do this conversion from requested org to org behind the scenes. 

**Tickets closed**:

DPC-721


jaredcunha-usds:
Screen Shot 2019-11-15 at 4 40 35 PM

PR #406, opened 2019-11-18
DPC-818: As a user I want to edit my organization's NPI and vendor from the dashboard.

switzersc-usds:
**Why**

Empower users to add their own NPI and vendor for the organization so that our internal admin team doesn't have to reach out to ask for these from each organization. 

Right now it's expected that this is is just the sandbox NPI; we will discuss any other requirements for production NPIs when we get there.

**What Changed**

* Adds form to user dashboard to let them edit an organization's NPI and vendor (they can only edit the NPI and vendor, nothing else)
* Adds text to the dashboard letting users know they need an NPI to start managing credentials for the API
* Kicks of registering the org in the API if there is an NPI (set by user or admin) *and* an api environment set by an admin

When no NPI or vendor:
Screen Shot 2019-11-21 at 2 44 21 PM

Form:
Screen Shot 2019-11-21 at 2 44 30 PM

With the NPI and sandbox enabled but no token or key:
Screen Shot 2019-11-21 at 4 58 06 PM

With NPI and vendor:
Screen Shot 2019-11-21 at 3 28 23 PM

View from internal admin dashboard:
image

**Choices Made**

See above and commit messages.

**Tickets closed**:

DPC-818

**Future Work**

Need to revisit this once after open sandbox launch.


dpc-web/spec/models/organization_spec.rb
```diff
@@ -135,4 +135,30 @@
       expect(org.api_credentialable?).to be false
     end
   end
+  
+  describe '#notify_users_of_sandbox_access' do
```
switzersc-usds: weird, this is supposed to be from master

ronaldheft-usds: @switzersc-usds Is something not right in master?

PR #407, opened 2019-11-18
DPC-837: Add support for provenance resources

nickrobison-usds:
**Why**

As part of our on-going attribution work and as required for DPC-264, we need the ability to extract provenance resources for specific requests, from the the appropriate header.

This is required for #329.

**What Changed**

Added the necessary Dropwizard plumbing to support a `@ProvenanceHeader` annotation, which will automatically parse and inject the FHIR resource from the `X-Provenance` header, per the FHIR spec. Individual methods can then make use of this annotation and have their request automatically fail if it's not provided.

**Choices Made**

Just the basic plumbing for now, it's not actually implemented anywhere, but will be in the future. This is mostly to unblock DPC-264.

**Tickets closed**:

DPC-837: Provenance header

**Future Work**

DPC-843: Create FHIR profile for provenance

**Checklist**


DeirdreHolub:
In our mob code review, there was some discussion about the amount of abstraction we introduce when adding code like this that "hides" parameter conversion. Having read through the description of Provenance, I think it's quite likely to be more widely used than it is at present, so I think this implementation was the right choice.

nickrobison-usds:
@DeirdreHolub Great suggestions, changes made!

dpc-common/src/test/java/gov/cms/dpc/fhir/parameters/ParamResourceFactoryTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+        final IParser parser = Mockito.mock(IParser.class);
+        Mockito.when(parser.parseResource(Mockito.eq(Parameters.class), Mockito.any(InputStream.class))).thenAnswer(answer -> {
+            throw new IOException();
+        });
+
+        final ParamResourceFactory factory = new ParamResourceFactory(mockInjector, null, parser);
+
+        final WebApplicationException exception = assertThrows(WebApplicationException.class, factory::provide, "Should throw exception");
+        assertAll(() -> assertEquals(HttpStatus.INTERNAL_SERVER_ERROR_500, exception.getResponse().getStatus(), "Should be a bad request"),
```
ronaldheft-usds: Are we wanting to return a 500 or a 400 here? The test says it should be a bad request but the assertion is expecting a 500.

nickrobison-usds: Good question, that's a copy-pasta bug, I'll fix it. Thanks!

PR #411, opened 2019-11-18
DPC-814: Fix invalid job state

ronaldheft-usds:
**Why**

We are finding cases of aggregator jobs that fail into a status of queued with the aggregatorID still set. This is an invalid state, and leads to further issues in the application.

**What Changed**

The issue turned out to be the stuck batch logic. It was resetting the status without resetting the aggregatorID (or clearing files, which could have led to a separate issues). Both of those issues have been fixed and tests were written to verify.

The fact that the stuck batch logic was triggering is concerning, so I added additional logging as well for us to hopefully track down the reason these batches get stuck. We had a ton of issues with deployments lately, so it could be related to those, but I would not be surprised if there are more cases that lead to stuck batches.

**Choices Made**

**Tickets closed**:

DPC-814: Aggregator jobs can fail into an invalid state

**Future Work**

**Checklist**

### DIFF
--- a/dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedBatchQueue.java
+++ b/dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedBatchQueue.java
@@ -169,6 +169,7 @@ public class DistributedBatchQueue extends JobQueueCommon {
             final List<JobQueueBatch> stuckJobList = session.createQuery(query).getResultList();
 
             for ( JobQueueBatch stuckJob : stuckJobList ) {
+                logger.warn(String.format("Restarting stuck batch... batchID=%s", stuckJob.getBatchID()));
                 stuckJob.restartBatch();
                 session.merge(stuckJob);
             }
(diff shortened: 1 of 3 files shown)


dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedBatchQueue.java
```diff
@@ -169,6 +169,7 @@ private void restartStuckBatches(Session session) {
             final List<JobQueueBatch> stuckJobList = session.createQuery(query).getResultList();
 
             for ( JobQueueBatch stuckJob : stuckJobList ) {
+                logger.warn(String.format("Restarting stuck batch... batchID=%s", stuckJob.getBatchID()));
```
embh: Any reason for choosing `String.format` here over `{}` (as used [here](http://www.slf4j.org/manual.html#typical_usage))? Fine either way - just wondering if you know something I don't. 😄

embh:
So I understand the context: is having the aggregator ID set when a job is queued a problem because that prevents the job from being claimed and started/resumed?

dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedBatchQueue.java
ronaldheft-usds: Force of habit. `String.format` allows more flexibility in the output formatting of variables, so I've always defaulted to using it.

PR #413, opened 2019-11-20
trap focus in modals, dialogs, etc.

jaredcunha-usds:
**Why**

For keyboard users, whenever a modal, flyout or any other "top layer" element is enabled, the focus state should be confined to that element. Otherwise, the focus state is lost. 

In addition, when that element is triggered, the focus state should also be moved to the close button for that element in the event that it was accidentally opened.

**What Changed**

- Added a javascript utility to trap focus
- Updated modal js so that focus state is transferred to the close button when it is opened

The keyboard navigation should function as such:

Screen Recording 2019-11-19 at 08 56 PM

Screen Recording 2019-11-19 at 08 57 PM

**Choices Made**

A JS utility is generic and can be used in a number of different instances.

### DIFF
--- a/dpc-web/app/assets/javascripts/utils/_trap-focus.js
+++ b/dpc-web/app/assets/javascripts/utils/_trap-focus.js
@@ -1,0 +1,32 @@
+// When dialogs, modals, or any other areas that appear on top
+// of the main canvas, we want to trap the focusable areas
+// for folks using keyboard navigation so that the focused
+// elements are not lost.
+
+function trapFocus(element) {
+    var focusableEls = element.querySelectorAll('a[href]:not([disabled]), button:not([disabled]), textarea:not([disabled]), input[type="text"]:not([disabled]), input[type="radio"]:not([disabled]), input[type="checkbox"]:not([disabled]), select:not([disabled]), input[type="submit"]:not([disabled])'),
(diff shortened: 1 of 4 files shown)


dpc-web/app/assets/javascripts/utils/_trap-focus.js
```diff
@@ -0,0 +1,32 @@
+// When dialogs, modals, or any other areas that appear on top
+// of the main canvas, we want to trap the focusable areas
+// for folks using keyboard navigation so that the focused
+// elements are not lost.
+
+function trapFocus(element) {
+    var focusableEls = element.querySelectorAll('a[href]:not([disabled]), button:not([disabled]), textarea:not([disabled]), input[type="text"]:not([disabled]), input[type="radio"]:not([disabled]), input[type="checkbox"]:not([disabled]), select:not([disabled]), input[type="submit"]:not([disabled])'),
```
switzersc-usds: just curious about the reasoning not to include all `input` (without specifying `type`) here.

jaredcunha-usds: you know, i pulled the script from elsewhere, but I suppose it can be all `input` types. But really in the DOM, it's either going to be `a` or a submit button or input. I can modify this in another branch.

PR #415, opened 2019-11-20
Do not verify cert in some envs

switzersc-usds:
**Why**

Dev env has some cert issues and we want to skip past that during some testing.

**What Changed**

Use env var to determine whether to verify ssl cert of API.

### DIFF
--- a/dpc-web/app/services/api_client.rb
+++ b/dpc-web/app/services/api_client.rb
@@ -108,9 +108,13 @@ class APIClient
 
   def http_request(request, uri)
     http = Net::HTTP.new(uri.host, uri.port)
-    http.use_ssl = use_ssl?
-    response = http.request(request)
 
+    if use_ssl?
+      http.use_ssl = true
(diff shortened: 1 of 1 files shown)


dpc-web/app/services/api_client.rb
```diff
@@ -126,4 +130,8 @@ def headers(token)
   def use_ssl?
     !(Rails.env.development? || Rails.env.test?)
   end
+
+  def verify_ssl_cert?
+    ENV.fetch('VERIFY_SSL_CERT') == 'true'
```
ronaldheft-usds: Can we change this to `ENV.fetch('VERIFY_SSL_CERT') != 'false'`?

Would feel safer that the environment defaults to verifying SSL certs unless the ENV var is specifically set to `false`.

switzersc-usds: yeah sounds good - updated

PR #417, opened 2019-11-20
DPC-439: Implementing streaming file downloads

nickrobison-usds:
**Why**

As we get ready for our production launch, we need to improve the way we handle large file downloading, so as to avoid reading gigabytes of files into memory. In addition, it would be really nice to have support for things like retriable and partial downloads.

**What Changed**

This PR essentially rewrites the `DataResource` class to move it from an in-memory file downloader to a more full-featured streaming file server.

Abstracted the file handling logic into a `FileManager` class, which makes it easier to encapsulate the file system access and improve our testability.

A lot of the logic was borrowed from various places on the internet, specifically, [this](http://www.nurkiewicz.com/2015/06/writing-download-server-part-i-always.html) series of helpful articles. I believe I've been able to expunge all of the non-streaming filters and methods from the actual download path. 

The API now sends along a bunch of additional metadata when returning a file download. Including: 
- `ETag` header with the SHA256 checksum. 
- `Content-Length` header with the total file size
- `Last-Modified` header with the creation timestamp of the file.

The `DataResource` now supports HTTP range requests, which means users can provide a `Range` header with a set of byte values and the API will only return that set of bytes, along with a `206` status. I believe the implementation fulfills the IETF spec, but it's a bit unclear in some places. The `HttpRangeHeaderParamConverter` class has the specific implementation along with some additional links.

The `DataResource` also allows the user to specific the `If-Non-Match` and `If-Unmodified-Since` headers and the API will return a `304` if either the supplied checksum matches the file, or if the `creationTimestamp` of the file is equal to or before the since header value. The checksum values are run through a _weak etag_ filter, which removes any trailing `--gzip` values that are added by Jersey when applying gzip compression to responses.

Added a `HEAD` method to the `DataResource` which mirrors the existing GET method, but without returning a body.

The Aggregator now handles doing the file checksumming before marking the job as complete, rather than requiring the API service to perform the same operation on each request to the Job status API. This should really help performance when handling multi-gigabyte exports. The implementation is a little hacky right now, 

**Choices Made**

The implementation of the aggregation checksumming is not great, but it works and we can improve it later.

Jersey really does not like to send along manual `Content-Length` headers, it continually strips them off which means that header is missing when doing a range download.

**Tickets closed**:

**Future Work**

DPC-891: Fix content length of download responses
DPC-892: Add conditional partial data response

**Checklist**


dpc-api/src/main/java/gov/cms/dpc/api/core/FileManager.java
```diff
@@ -0,0 +1,104 @@
...
+            query.setParameter("fileName", fileID);
+            query.setParameter("org", organizationID);
+            final List objects = query.getResultList();
+            final Object[] objectArray = (Object[]) objects.get(0);
+            creationTime = (OffsetDateTime) objectArray[0];
+            batchFile = (JobQueueBatchFile) objectArray[1];
+        } catch (NoResultException e) {
+            throw new WebApplicationException("Cannot find file", Response.Status.NOT_FOUND);
+        }
+
```
RickHawesUSDS: I suggest that this block should be a method in the Queue implementation to keep the encapsulation of the Queue package from leaking out.

dpc-api/src/main/java/gov/cms/dpc/api/converters/HttpRangeHeaderParamConverter.java
```diff
@@ -0,0 +1,53 @@
...
+    public RangeHeader fromString(String value) {
+        // If the range request is completely empty, treat it as if it was never set
+        if (value == null || value.isEmpty()) {
+            return null;
+        }
+        final Matcher matcher = RANGE_REGEX.matcher(value);
+        if (matcher.matches()) {
+            final RangeHeader rangeHeader = new RangeHeader();
+            rangeHeader.setUnit(matcher.group(1));
+            rangeHeader.setStart(Long.parseLong(matcher.group(2)));
```
RickHawesUSDS: I think bytes are the only units accepted. I was expecting some form of unit validation here.

nickrobison-usds: Good catch. Added a check to the `DataResource` unit tests.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
```diff
@@ -145,8 +149,10 @@ protected void processJobBatch(JobQueueBatch job) {
             }
 
             // Finish processing the batch
-            if ( queueRunning ) {
+            if (queueRunning) {
                 logger.info("COMPLETED job {} batch {}", job.getJobID(), job.getBatchID());
+                // Calculate metadata for the file (length and checksum)
+                calculateFileMetadata(job);
```
ronaldheft-usds: This seems like a great spot to do the meta data calculations. 👍

nickrobison-usds:
@ronaldheft-usds That's correct, existing jobs will just return empty values. But hopefully that would only be for a bit longer!

PR #418, opened 2019-11-20
DPC-819: Send an email when users are added to a sandbox-enabled organization

switzersc-usds:
**Why**

We want users to get info on how to use sandbox once they belong to an organization that is sandbox enabled.

**What Changed**

* Send notification email to users once they are added to an org already sandbox enabled
* Send notification email to users already assigned to an org as soon as that org becomes sandbox enabled

image
image
image
image

**Choices Made**

**Tickets closed**:

DPC-819


dpc-web/app/controllers/internal/organizations_controller.rb
```diff
@@ -86,5 +89,9 @@ def organization_params
         fhir_endpoints_attributes: %i[id name status uri]
       )
     end
+
+    def sandbox_added?
+      (organization_params[:api_environments] - @organization.api_environments).include?('0')
```
nickrobison-usds: This seems clever, but I'm not sure I'm following the logic, could you explain what this is checking for? Sorry to be dense!

switzersc-usds: no worries! good question. it takes the params (form fields) that are coming in that were input by the user and sees what the user selected for API environments, and compares that to the organization's existing API environments. If sandbox (0) is present in the params but not the existing list saved for the org, that means the user has added sandbox to the org. If that's the case, we want to notify the org's existing users that they now have sandbox access.

nickrobison-usds: Nice! Would it be possible to change the 0 to an enum value, or at least add a comment indicating what the 0 value indicates?

switzersc-usds: I can add a comment. Because this column is an array of integers, not an enum, it's not quite as simple to change this to a string (would have to add mapping or change the array to be one of strings, not integers).

switzersc-usds: For posterity (and I believe this info is in a previous PR or commit message), I went with an array of integers for this column because of extensibility (if names change) and storage size, and hopefully one day we can use some of Rails' enum utilities (if we have time to adapt them or Rails comes out with support for enum in arrays). The sacrifice though is human readability and having to do lots of mapping and remembering between names and numbers, and an array of strings probably would've made just as much sense. Ah well 🤷‍♀

PR #419, opened 2019-11-21
DPC-849: Fix thread safety issue with HAPI Validator

nickrobison-usds:
**Why**

A user reported in the Google group that they were experiencing issues with the validator rejecting resources because it was associating them with an incorrect profile.

This issue occurred previously and is related to the fact that none of the HAPI validator classes are thread-safe. I had thought it was fixed, but apparently not.

In addition, our smoke tests are passing even though the validator was throwing a 422 exception.

**What Changed**

Updated the smoke tests to _hopefully_ improve error handling, we're now throwing errors from the main `SmokeTest` class, rather than deeper in the helper utils.

Also, staggered the startup time so the tests ramp up a bit slower and hopefully surface more threading and resource issues.

The SmokeTests now use FHIR creations when uploading resources, previously we were using the `$submit` operation, which was fine, but now we're stressing the system a bit more.

Updated the ClientUtils class to ensure that there aren't any errors returned as part of the export process, this should hopefully catch the patient ID issue that we're tracking with DPC-850.

**Choices Made**

Still not sure what's the root cause of the exception swallowing, but at least I've been able to verify that some exceptions are actually thrown correctly.

**Tickets closed**:

DPC-849

**Future Work**

**Checklist**

### DIFF
--- a/dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/ClientUtils.java
+++ b/dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/ClientUtils.java
@@ -50,6 +50,10 @@ public class ClientUtils {
                 .map(npi -> exportRequestDispatcher(exportClient, npi))
                 .map(search -> (Group) search.getEntryFirstRep().getResource())
                 .map(group -> jobCompletionLambda(exportClient, httpClient, group))
+                .peek(jobResponse -> {
+                    if (jobResponse.getError().size() > 0)
+                        throw new IllegalStateException("Errors reported during export");
+                })
                 .forEach(jobResponse -> jobResponse.getOutput().forEach(entry -> {
                     jobResponseHandler(httpClient, entry);
                 }));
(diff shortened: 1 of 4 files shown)


dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/ClientUtils.java
```diff
@@ -50,6 +50,10 @@ static void handleExportJob(IGenericClient exportClient, List<String> providerNP
                 .map(npi -> exportRequestDispatcher(exportClient, npi))
                 .map(search -> (Group) search.getEntryFirstRep().getResource())
                 .map(group -> jobCompletionLambda(exportClient, httpClient, group))
+                .peek(jobResponse -> {
+                    if (jobResponse.getError().size() > 0)
+                        throw new IllegalStateException("Errors reported during export");
```
ronaldheft-usds: Nice! I like the extra assertions to verify no errors.

dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/ClientUtils.java
```diff
@@ -254,15 +261,11 @@ private static JobCompletionModel jobCompletionLambda(IGenericClient exportClien
         return patientReferences;
     }
 
-    static List<String> submitPractitioners(String providerBundleFilename, Class<?> baseClass, FhirContext ctx, IGenericClient exportClient) {
+    static List<String> submitPractitioners(String providerBundleFilename, Class<?> baseClass, FhirContext ctx, IGenericClient exportClient) throws IOException {
         final Bundle providerBundle;
 
-        try {
```
ronaldheft-usds: 🤔 You would think a RuntimeException would be good enough for JMeter.

dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/SmokeTest.java
```diff
@@ -75,6 +75,9 @@ public SampleResult runTest(JavaSamplerContext javaSamplerContext) {
         final IGenericClient adminClient = APIAuthHelpers.buildAdminClient(ctx, hostParam, goldenMacaroon, true);
 
         final SampleResult smokeTestResult = new SampleResult();
+        smokeTestResult.setSampleLabel("Smoke Test");
+        // False, unless proven otherwise
```
ronaldheft-usds: Well said.

src/test/smoke_test.yml
```diff
@@ -2,9 +2,10 @@ settings:
   artifacts-dir: bzt-out/%Y-%m-%d_%H-%M-%S.%f
 
 execution:
-  - iterations: 1
+  - iterations: 3
```
ronaldheft-usds: Shouldn't we apply these settings in DEV as well, so we can catch concurrency issues during our normal CD process?

nickrobison-usds: The root config is applied to all the smoke test environments. So this covers dev, local and test. This file gets merged with the environment specific file, e.g. `test.smoke_test.yml`

ronaldheft-usds: Ah yeah, totally makes sense. Brain fart moment seeing `*_test.yml` 🤣

PR #421, opened 2019-11-21
Smoke test cleanups

nickrobison-usds:
**Why**

As we've been in the perennial debug loop for our JMeter tests, I made a couple of small tweaks to make life better.

**What Changed**

Reduced the test ramp-up time from 120s to 60s, this was a change that I missed on an earlier branch, still catches the threading issues but makes the tests run a bit quicker.

Improved the error message when an export job completes but has error outcomes.

Updated the `done` handling in the export job monitoring. Now, if a non-202 status is returned we check to see if it's an error response and if so, parse out the error as a string.


ronaldheft-usds:
Merged because the smoke test keeps on timing out. Seeing if this improves.

PR #431, opened 2019-11-22
DPC-850: Bugfix - Save number of providers on registration

switzersc-usds:
**Why**

We're not saving number of providers when users register. The param needed to be added to the devise params in Application Controller from when I renamed this column.

**What Changed**

Updated allowed devise params. Changed validation to require number be greater than 0 (not greater than or equal to).

**Choices Made**

See commit messages.

**Tickets closed**:

DPC-850

### DIFF
--- a/dpc-web/app/models/user.rb
+++ b/dpc-web/app/models/user.rb
@@ -23,7 +23,8 @@ class User < ApplicationRecord
   validates :requested_organization_type, inclusion: { in: ORGANIZATION_TYPES.keys }
   validates :last_name, :first_name, presence: true
   validates :requested_organization, presence: true
-  validates :requested_num_providers, numericality: { only_integer: true, greater_than_or_equal_to: 0, allow_nil: true },
+  validates :requested_num_providers, numericality: { only_integer: true, greater_than_or_equal_to: 0,
+                                                      allow_nil: true },
                                       if: -> { health_it_vendor? }
   validates :requested_num_providers, numericality: { only_integer: true, greater_than: 0 },
                                       unless: -> { health_it_vendor? }
--- a/dpc-web/app/models/user.rb
+++ b/dpc-web/app/models/user.rb
@@ -23,7 +23,7 @@ class User < ApplicationRecord
   validates :requested_organization_type, inclusion: { in: ORGANIZATION_TYPES.keys }
   validates :last_name, :first_name, presence: true
   validates :requested_organization, presence: true
-  validates :requested_num_providers, numericality: { only_integer: true, greater_than_or_equal_to: 0, allow_nil: true }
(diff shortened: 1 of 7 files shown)


dpc-web/app/models/user.rb
```diff
@@ -23,7 +23,7 @@ class User < ApplicationRecord
   validates :requested_organization_type, inclusion: { in: ORGANIZATION_TYPES.keys }
   validates :last_name, :first_name, presence: true
   validates :requested_organization, presence: true
-  validates :requested_num_providers, numericality: { only_integer: true, greater_than_or_equal_to: 0, allow_nil: true }
+  validates :requested_num_providers, numericality: { only_integer: true, greater_than: 0 }
```
ronaldheft-usds: Yay! Required!

ronaldheft-usds: Is there any area where text would reference this is optional?

switzersc-usds: I don't think so, but I'll double check.

switzersc-usds: Oh, yup, the field says optional. Removing.

switzersc-usds: Oh crap, just remembered vendors can leave it blank. Hmm, I need to add a conditional to only validate this if they are not a vendor

switzersc-usds:
@ronaldheft-usds - Just made a couple updates

dpc-web/app/models/user.rb
```diff
@@ -23,7 +23,10 @@ class User < ApplicationRecord
   validates :requested_organization_type, inclusion: { in: ORGANIZATION_TYPES.keys }
   validates :last_name, :first_name, presence: true
   validates :requested_organization, presence: true
-  validates :requested_num_providers, numericality: { only_integer: true, greater_than_or_equal_to: 0, allow_nil: true }
+  validates :requested_num_providers, numericality: { only_integer: true, greater_than_or_equal_to: 0 },
```
DeirdreHolub: It looks like you've removed the option to leave it nil. Does this mean a user will be forced to enter a 0, or will it default to 0? FWIW, a nil seems like a reasonable way to indicate "I don't know how to answer this question now".

switzersc-usds: Good catch. For vendors, we definitely want them to be able to leave this blank and have it default to 0, so I'm adding `allow_nil: true` back for them. This will let them leave it blank (which is good because this field doesn't even show up for them, so it has to be left blank) and then the DB will default the blank column to 0.

For providers, my understanding is that we don't want them to put 0 or leave blank if they don't know -- they have to put *something*, even if it's a guesstimate.

ronaldheft-usds:
Looks great now! Thanks for the changes!

PR #434, opened 2019-11-22
DPC-866: Fix token key signing security flaw

nickrobison-usds:
**Why**

Discovered a bug in the authentication system where it was possible to sign an organization client_token with *any* key pair registered with the system. That meant if a client_token leaked, any other organization could use it gain access to that org's data.

This was because the key lookup for the JWT was not taking into account the organization id.

**What Changed**

Updated the `JWTKeyResolver` method to account for the organization id, when searching by key id.

This requires us to parse the provided Macaroon for the `organization_id` caveat and use that in the lookup. It's a bit clunky, but actually secure.

To support this, I made a couple of methods on the `MacaroonBakery` static, so we could use their logic in other places.

Added an additional integration test to verify keys and tokens have to be within the same organization.

**Choices Made**

**Tickets closed**:

DPC-866

**Future Work**

**Checklist**


ronaldheft-usds:
Looks like we have failing tests. Holding on the merge until those are passing.

ronaldheft-usds:
@nickrobison-usds So I've been trying to get these tests passing, and there seems to be a couple of issues.

1) Some tests need the organization id mocked because they're using a fake macaroon, and that fake macaroon is not deserializing with the organizationID. (I have this issue fixed locally.)
2) Any test that is using a golden macaroon.

The issue with the golden macaroon seems pretty serious, but maybe there is a simple solution to it.

From my observations, the tests were working previously because they were actually picking up the incorrect public key. They had a key collision with `integration-test-key`, which belonged to the organization token and not the golden macaroon. Now the collision doesn't work, because the golden macaroon does not have an organization set, so the `PublicKeyDao` return null.

I'm thinking `JwtKeyResolver` shouldn't even be running for golden macaroons, and it was probably a bug it was running previously. Thoughts?

nickrobison-usds:
@ronaldheft-usds I believe I've solved the test failures, everything is passing on my local machine.

Looks like the tests had a couple of issues, mostly related to me rushing the PR out on Friday evening. The _PatientResource_ test wasn't using the correct organization ID, which was causing the public keys to be associated to the wrong organizations. Simple fix.

The _OrganizationResource_ test needed to use the _buildAdminClient_ method, which takes a golden macaroon and allows for doing admin things, like creating organizations in the first place.

PR #424, opened 2019-11-21
Sandbox documentation updates

nickrobison-usds:
**Why**

This makes some required updates to the documentation ahead of the sandbox release. Trying to follow along with the related confluence [doc](https://confluenceent.cms.gov/display/DAPC/Updates+to+screen+text+after+token+changes).

**What Changed**

Updated the attribution section in the documentation to be explicit about what constitutes a treatment related purpose. Will definitely need a wording review.

Fixed several little issues with the Swagger docs:
- Updated the scheme to use https (except for running locally) which avoids the redirect
- Removed the example from the `/metadata` endpoint, which was causing the browser to freeze due to the size of the json payload.
- Added explicit `@Path` annotations to all of our resources that mark them under the `/v1` root, not sure why this wasn't been propagated correctly, but it is now.
- Added missing Auth annotations to the `/Practitioner` resource (just the annotations were missing, not the actual auth!)

**Choices Made**

This change does not include the remapping of the MBIs, I wanted to get it out quickly for the wording review and for touching all the resources.

**Checklist**

- [x] All tests are passing via `make ci-app` (app change) and `make ci-web` (website change)
- [x] Swagger documentation has been updated
- [x] FHIR documentation has been updated
- [x] Any required dpc-ops changes have a PR submitted and mentioned in this ticket
- [x] Any manual migration steps are documented, scripts written (where applicable), and tested
- [x] Before merging, any required dpc-ops changes have been approved and merged into master of the dpc-ops repo

PR #441, opened 2019-11-25
Revert "Sandbox documentation updates"

ronaldheft-usds:
Reverts CMSgov/dpc-app#424

This PR contained failing tests.


ronaldheft-usds:
Appears like we have an intermittent failing test, and this PR wasn't the root cause for the failure.

PR #448, opened 2019-11-26
Rough command for importing organizations into the website

ronaldheft-usds:
There is a lot to be desired in this script, but it _should_ handle our import for this one off case.

This PR will **not**  be merged and is for reference only.

### DIFF
--- a/dpc-attribution/src/main/java/gov/cms/dpc/attribution/cli/ImportCommand.java
+++ b/dpc-attribution/src/main/java/gov/cms/dpc/attribution/cli/ImportCommand.java
@@ -1,0 +1,97 @@
+package gov.cms.dpc.attribution.cli;
+
+import gov.cms.dpc.attribution.DPCAttributionConfiguration;
+import io.dropwizard.Application;
+import io.dropwizard.cli.EnvironmentCommand;
+import io.dropwizard.db.ManagedDataSource;
+import io.dropwizard.db.PooledDataSourceFactory;
+import io.dropwizard.setup.Environment;
+import net.sourceforge.argparse4j.inf.Namespace;
+import net.sourceforge.argparse4j.inf.Subparser;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.Statement;
+
+public class ImportCommand extends EnvironmentCommand<DPCAttributionConfiguration> {
+
+    public ImportCommand(Application<DPCAttributionConfiguration> application) {
+        super(application, "import", "Import into the website");
+    }
+
+    @Override
+    public void configure(Subparser subparser) {
+        super.configure(subparser);
+    }
+
+    @Override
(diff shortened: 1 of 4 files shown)


nickrobison-usds:
Before we go too far down this road, I just want to ask one dumb question. Is this a process that we will need to replicate for our production pilots as well when we go live?

If so, does it make more sense to implement #447 so we have an existing solution for that time?

ronaldheft-usds:
@nickrobison-usds We won't. The website will function as expected creating their tokens in production because the website will already have its internal organization created and will be then responsible for creating the org in production.

dpc-attribution/src/main/java/gov/cms/dpc/attribution/cli/ImportCommand.java
```diff
@@ -0,0 +1,120 @@
...
+            String name = resultSet.getString(3);
+            String endpointId = resultSet.getString(4);
+
+            this.importIntoWebsite(webConnection, id, npi, name, endpointId);
+        }
+    }
+
+    protected void importIntoWebsite(Connection webConnection, String id, String npi, String name, String endpointId) throws Exception {
+        String orgInsert = String.format(
+                "INSERT INTO organizations(name, npi, organization_type, num_providers, api_environments, created_at, updated_at) VALUES ('%s', '%s', 0, 0, '{0}', now(), now())",
```
switzersc-usds: just wanna note that the org name should be unique. is that enforced in the api at all or do we risk inserting non-unique names into the website db?

dpc-attribution/src/main/java/gov/cms/dpc/attribution/cli/ImportCommand.java
```diff
@@ -0,0 +1,120 @@
...
+        );
+        System.out.println(fhirInsert);
+
+        if ( RUN_INSERTS ) {
+            PreparedStatement fhirInsertStatement = webConnection.prepareStatement(fhirInsert);
+            fhirInsertStatement.executeUpdate();
+        }
+
+        String addressInsert = String.format(
+                "INSERT INTO addresses(street, city, state, zip, addressable_type, addressable_id, created_at, updated_at, address_use, address_type) VALUES ('123 Fake St.', 'Washington', 'DC', '20000', 'Organization', %s, now(), now(), 0, 0)",
```
switzersc-usds: Lol okay

dpc-attribution/src/main/java/gov/cms/dpc/attribution/cli/ImportCommand.java
ronaldheft-usds: It's not enforced in the API, but it appears the website is okay with duplicate names. I was able to import the same org multiple times in DEV.

switzersc-usds: It's okay at the database level but at the model level those records will be considered invalid, so a user will not be able to update them until the name is changed to be unique.

ronaldheft-usds:
This was run successfully.

PR #501, opened 2019-12-19
DPC-302: Add support for transforming resource lists into bundles

nickrobison-usds:
**Why**

This PR continues our efforts in abstracting away the custom nature of FHIR resources from the core Dropwizard/Jersey resources. Currently, the endpoints only support returning single FHIR resources, which means means in order to return collections, we need to manually convert them to `Bundle` resources before leaving the method.

This PR adds support for directly returning collections of FHIR resources from endpoints, and converting them to `Bundle` resources using a custom `MessageBodyWriter` class.

**What Changed**

Add a `BundleHandler` class which takes care of converting java `Collections` of FHIR `Resources` into a corresponding `Bundle`. This also takes care of setting the bundleType and total properties correctly.

Also added a `BundleReturnOptions` annotation, which lets us customize the resulting bundle. Currently, it only allows us to modify the type, but more features can be added in the future.

Also updated all the resource signatures to return `List<Resource>` rather than a raw Bundle. We can still return actual bundle resources, if desired.

**Choices Made**

One drawback of this approach is that Swagger is no longer able to infer the actual return type of the method. So we need add a the `response=Bundle.class` to each of the `ApiOperation` annotations. It's a bit annoying, but works.

**Tickets closed**:

DPC-302: Return collections instead of bundles.

**Future Work**

**Checklist**


nickrobison-usds:
@embh @DeirdreHolub This PR is ready to go, I'm not sure why the total code coverage number dropped, I think it's because it touched so many classes, but the newly added ones are fully tested, so I'm ok with merging it.

PR #503, opened 2019-12-19
DPC-921: Update request filters to reject non-FHIR content types

nickrobison-usds:
**Why**

On the Google Group, a user reported an issue where they were getting a weird exception when trying to submit their practitioners. It turned out to be due to our request filters accepting incorrect content types, but then it gets delegated to the Jackson parser which fails to deserialize it.

**What Changed**

Improved the request checking in the `FHIRRequestFilter` class to verify that both the `Accepts` header and the `Content-Type` header are correct. Really, content-type is the most important one, because it's the one that determines which `MessageBodyReader` gets called, and we need it to correctly call our custom FHIR reader.

**Choices Made**

**Tickets closed**:

DPC-921: Reject non-FHIR requests to FHIR endpoints.

**Future Work**

**Checklist**

### DIFF
--- a/dpc-common/src/main/java/gov/cms/dpc/fhir/dropwizard/filters/FHIRRequestFilter.java
+++ b/dpc-common/src/main/java/gov/cms/dpc/fhir/dropwizard/filters/FHIRRequestFilter.java
@@ -20,10 +20,10 @@ public class FHIRRequestFilter implements ContainerRequestFilter {
 
     @Override
     public void filter(ContainerRequestContext requestContext) {
-        // Ensure the Accepts header is set to a FHIR media type
+        // Ensure the `Accept` header is set to a FHIR media type
         checkAccepts(requestContext);
         // The content type header is optional, but if it's present, it has to be a FHIR resource type
-        // HAPI does NOT set the Content-Type, so we can't require it as part of our requests, otherwise our test suite breaks
+        // HAPI does NOT set the Content-Type by default, so we can't require it as part of our requests, otherwise our test suite breaks
         checkContentType(requestContext);
     }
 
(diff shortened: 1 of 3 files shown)


dpc-common/src/main/java/gov/cms/dpc/fhir/dropwizard/filters/FHIRRequestFilter.java
```diff
@@ -6,26 +6,51 @@
...
+    FHIRRequestFilter() {
+        // Not used
+    }
+
     @Override
     public void filter(ContainerRequestContext requestContext) {
+        // Ensure the Accepts header is set to a FHIR media type
+        checkAccepts(requestContext);
+        // The content type header is optional, but if it's present, it has to be a FHIR resource type
+        // HAPI does NOT set the Content-Type, so we can't require it as part of our requests, otherwise our test suite breaks
```
embh: Does this comment mean the `Content-Type` header can't be set with the HAPI client that we use in tests?

dpc-common/src/main/java/gov/cms/dpc/fhir/dropwizard/filters/FHIRRequestFilter.java
```diff
@@ -6,26 +6,51 @@
...
 
 public class FHIRRequestFilter implements ContainerRequestFilter {
 
+    FHIRRequestFilter() {
+        // Not used
+    }
+
     @Override
     public void filter(ContainerRequestContext requestContext) {
+        // Ensure the Accepts header is set to a FHIR media type
```
embh: Not a blocker, but we might want to change the references to `Accepts` to `Accept` for clarity.

dpc-common/src/main/java/gov/cms/dpc/fhir/dropwizard/filters/FHIRRequestFilter.java
nickrobison-usds: I can, but you have to manually add the header, which is kind of awkward. I meant it does not set the header by default

PR #504, opened 2019-12-20
DPC-276: Unify Hibernate entity conversion utils.

nickrobison-usds:
**Why**

As the number of FHIR resources and corresponding Java class increases, we need a better way of handling the conversions and related utilities. Rather than sticking the methods on the Java entities themselves, it would be nice to have an extensible framework for centralizing the conversion logic.

**What Changed**

Created a new `FHIREntityConverter` class, which provides two methods for converting to and from FHIR (toFHIR() and fromFHIR()). This allows registering `FHIRConverter` interfaces with the Java service loader, which will be automatically registered with the converter and used going forward.

Moved all the various conversion logic out of the entity classes and into these helpers. There are still a couple of places where this isn't possible, namely because we need some additional context when doing the conversion that's not readily available. Eventually these can be pulled in, but we'll need to make some changes to the method signatures.

All of the resources now have the `FHIREntityConverter` as an injection parameter, in order to support easy conversion.

**Choices Made**

This isn't a full reflection based solution, like Jackson. Instead it takes a similar approach the `StdSerializer` classes, which give fine grained control over individual classes as well as allowing delegation back to the main converter.

**Tickets closed**:

DPC-276: Centralize entity conversion

**Future Work**

DPC-935: Improve testing of the conversion classes themselves.

**Checklist**


dpc-common/src/main/java/gov/cms/dpc/fhir/converters/FHIRConverter.java
```diff
@@ -0,0 +1,14 @@
+package gov.cms.dpc.fhir.converters;
+
+import org.hl7.fhir.dstu3.model.Base;
+
+public interface FHIRConverter<R extends Base, C> {
+
+    C fromFHIR(FHIREntityConverter converter, R resource);
+
+    R toFHIR(FHIREntityConverter converter, C javaClass);
```
embh: I would expect something called `javaClass` to be of type `Class`. Isn't `C` the Hibernate entity?

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/FHIREntityConverter.java
```diff
@@ -0,0 +1,148 @@
...
+        final List<FHIRConverter<?, ?>> converters = ServiceLoaderHelpers.getLoaderStream(FHIRConverter.class)
+                .map(l -> (FHIRConverter<?, ?>) l)
+                .collect(Collectors.toList());
+
+        return FHIREntityConverter.initialize(converters);
+
+    }
+
+    /**
+     * Create a new {@link FHIRConverter} with the given {@link Collection} of {@link FHIRConverter}s
```
embh:
```suggestion
     * Create a new {@link FHIREntityConverter} with the given {@link Collection} of {@link FHIRConverter}s
```

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/FHIRResourceConverter.java
```diff
@@ -9,14 +11,14 @@
  * @param <C> - Java class to convert to
  */
 @FunctionalInterface
-public interface FHIRResourceConverter<R extends BaseResource, C> {
+public interface FHIRResourceConverter<R extends Base, C> {
 
     /**
      * Convert FHIR {@link BaseResource} into a corresponding Java class
      *
      * @param resource - {@link BaseResource} to convert
```
embh: Should these links be to `Base` now?

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/TestPatientEntityConverter.java
```diff
@@ -0,0 +1,41 @@
+package gov.cms.dpc.fhir.converters;
+
+import gov.cms.dpc.common.entities.PatientEntity;
+import org.hl7.fhir.dstu3.model.Patient;
+
+import java.util.UUID;
+
+public class TestPatientEntityConverter implements FHIRConverter<Patient, PatientEntity> {
```
embh: How/where is this used?

nickrobison-usds: It's not, should have been removed. Thanks!

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/FHIRConverter.java
nickrobison-usds: Good call. Thanks!

dpc-common/src/test/java/gov/cms/dpc/common/entities/OrganizationEntityTest.java
```diff
@@ -14,11 +16,14 @@
 class OrganizationEntityTest {
 
     @Test
+    @Disabled
```
embh: Is this test (and `testSimpleRoundTrip`) disabled because an organization entity needs more context when converting?

nickrobison-usds: That's correct, I've removed one of the duplicated tests and added a comment as to why the remaining test is disabled. We'll pick it up as part of DPC-935.

PR #513, opened 2019-12-27
DPC-922: Update validator to check for correct expiration time format.

nickrobison-usds:
**Why**

A GG user reported a `500` exception when trying to request an access token. The logs indicated it was due to a full timestamp being submitted in the `exp` claim. Unfortunately, this passed the validator, which isn't great.

**What Changed**

Update the validator to ensure the supplied expiration time is in seconds since epoch. We do this by trying to cast the value to an integer and throwing an exception if that fails.

**Choices Made**

**Tickets closed**:

DPC-922: Have validator check exp format.

**Future Work**

**Checklist**


dpc-api/src/test/java/gov/cms/dpc/api/auth/jwt/JWTUnitTests.java
```diff
@@ -620,6 +622,32 @@ void testInvalidKID() throws NoSuchAlgorithmException {
...
+            final KeyPair keyPair = APIAuthHelpers.generateKeyPair();
+
+            final String id = UUID.randomUUID().toString();
+            final String jwt = Jwts.builder()
+                    .setHeaderParam("kid", UUID.randomUUID())
+                    .setAudience("localhost:3002/v1/Token/auth")
+                    .setIssuer(m)
+                    .setSubject(m)
+                    .setId(id)
+                    .claim("exp", Instant.now().plus(1, ChronoUnit.MINUTES).atOffset(ZoneOffset.UTC).format(DateTimeFormatter.ISO_OFFSET_DATE_TIME))
```
embh: Is there a test that sets `exp` directly as an integer? It looks like the other tests use `setExpiration(Date)`, but I may have missed it.

nickrobison-usds: Good catch, added!

PR #516, opened 2019-12-31
DPC-453: Submit attribution attestation

nickrobison-usds:
**Why**

CMS is requiring users to submit an _attestation_ of treatment purposes each time they add a patient to a physician roster. 

**What Changed**

Added an additional FHIR profile that verifies the required information is present in the `X-Provenance` header. Specifically we need a couple of things:

- `recorded` a timestamp of when the attestation occured
- `reason` the reason for the attestation which is currently fixed to _http://hl7.org/fhir/v3/ActReason#TREAT_
- `agent` the organization performing the action _onBehalfOf_ a given provider.

```
{
  "resourceType": "Provenance",
  "meta": {
    "profile": [
      "https://dpc.cms.gov/api/v1/StructureDefinition/dpc-profile-attestation"
    ]
  },
  "recorded": "1990-01-01T00:00:00.000-05:00",
  "reason": [
    {
      "system": "http://hl7.org/fhir/v3/ActReason",
      "code": "TREAT"
    }
  ],
  "agent": [
    {
      "role": [
        {
          "coding": [
            {
              "system": "http://hl7.org/fhir/v3/RoleClass",
              "code": "AGNT"
            }
          ]
        }
      ],
      "whoReference": {
        "reference": "Organization/{organization id}"
      },
      "onBehalfOfReference": {
        "reference": "Practitioner/{practitioner id}"
      }
    }
  ]
}
```

The `/Group` endpoints now require this profile as a header value.

**Choices Made**

Right now, the attestation is merely logged and not persistent anywhere. Eventually we'll need to wire this into the backend to store the attestation records, but that will probably come as part of the auditing framework.

By adding the attestation profile, we need to disable validation against the base resource definitions, since we no longer require the `target` value to be set. This shouldn't be a big deal, as we merge our differential profiles with the underlying snapshot, so the base should still be validated correctly.

**Tickets closed**:
DPC-453

**Future Work**

- Wire attestation into auditing framework
- Add additional profile for individual patient _$everything_ calls, which adds back the `target` field.

**Checklist**


dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
```diff
@@ -355,4 +342,51 @@ private static void checkExportRequest(String outputFormat, String since) {
...
+        final Coding reason = provenance.getReasonFirstRep();
+
+        final Provenance.ProvenanceAgentComponent performer = FHIRExtractors.getProvenancePerformer(provenance);
+        final List<Reference> attributedPatients = attributionRoster
+                .getMember()
+                .stream()
+                .map(Group.GroupMemberComponent::getEntity)
+                .collect(Collectors.toList());
+
+        logger.info("Organization {} is attesting a {} purpose between provider {} and patient(s) {}", performer.getWhoReference(), performer.getOnBehalfOfReference(), reason, attributedPatients);
```
DeirdreHolub: Shouldn't `reason` be the second param, and performer.getOnBehalfOfReference() be the third param? 

Also, do we really want the entire list of attributed patients (up to 5000 iirc) in the log? This doesn't seem like an actionable result. That is, if we found an issue that required resolution via looking at the logs, we could have an overwhelming amount of not-easily-parsed information. 

Finally, isn't this dumping PII into the log?

nickrobison-usds: Good catch on the param ordering! I'll fix that.

I don't believe references are PII, they serve the same function as BFD IDs. And I do believe we want the entire attestation in the logs for auditing purposes. If you have a better suggestion, I'm definitely open to it. This is also a temporary solution until we have our auditing framework fleshed out.

embh: Is it possible to include an example of the log message in the PR description? I'm wondering what the list of patient references will look like. I don't see a `toString()` on `Reference` to print something more useful than the class and hash.

DeirdreHolub: Clarifying question: the attestation is not going to contain a list of patients itself, correct? The list is whatever roster of patients was submitted with the request. That is, when an organization creates a roster for the first time, the list can have up to 5,000 entries; if the organization is updating a roster, it might have 1.

DeirdreHolub: I think we should add the group id and roster id that are being updated. With those, we would have enough information to write a database query if we needed to validate the operation.

nickrobison-usds:
Thank you both for your attention to detail! This was a really sloppy PR, my apologies for not testing better.

I've corrected the toString methods and added a new set of Unit tests to verify the output. I've also added the rosterID to the log output, when it's available.

@DeirdreHolub Your understanding is correct, this will only log the patients being _added_ to a roster, or when a roster is replaced. So the initial log could be up to 5000 patients, but then it would mostly be smaller.

PR #523, opened 2020-01-06
DPC-926: Updated Backend Services to POST form params

nickrobison-usds:
**Why**

One of the Google Group uses reported that they were unable to generate access tokens from the `Token/auth` endpoint. The issue was that their query params were being stripped and never making it to our endpoint. I'm not sure why their library wasn't working, but it revealed a poor implementation of the Backend Services spec where we should have been receiving the auth values as url encoded form params, in the post body, rather than as query params.

This will be a breaking change for our users, but hopefully a simple one.

**What Changed**

Updated authentication implementation to move from requiring query params to requiring encoded form params in the body. This actually matches the backend services spec, our previous implementation was incorrect.

All of our code and helper scripts have been updated to account for this. This WILL be a breaking change for users, but it should also fix some of the issues that have been reported.

**Choices Made**

**Tickets closed**:

DPC-926

**Future Work**

**Checklist**


nickrobison-usds:
Confirmed at the FHIR connectathon that this fix works correctly. We'll need to work out a messaging strategy to ensure we handle the breaking change appropriately.

dpc-testing/src/main/java/gov/cms/dpc/testing/APIAuthHelpers.java
```diff
@@ -132,6 +144,21 @@ public static AuthResponse jwtAuthFlow(String baseURL, String macaroon, UUID key
         return authResponse;
     }
 
+    private static NameValuePair encodeNameValuePair(String name, String value) {
```
embh: Would `BasicNameValuePair` serve the same purpose? https://hc.apache.org/httpcomponents-core-4.4.x/httpcore/apidocs/org/apache/http/message/BasicNameValuePair.html

nickrobison-usds: Yes! Thank you! Not sure how I missed that the first time around.

dpc-testing/src/main/java/gov/cms/dpc/testing/APIAuthHelpers.java
```diff
@@ -120,6 +132,7 @@ public static AuthResponse jwtAuthFlow(String baseURL, String macaroon, UUID key
             builder.addParameter("client_assertion_type", CLIENT_ASSERTION_TYPE);
             builder.addParameter("client_assertion", jwt);
```
embh: With the `HttpPost` entity being the form data, are parameters still needed on the URL?

dpc-api/src/test/java/gov/cms/dpc/api/auth/jwt/JWTUnitTests.java
```diff
@@ -68,50 +70,51 @@ private JWTUnitTests() {
 
     @Nested
     @DisplayName("Query Param tests")
```
embh:
```suggestion
    @DisplayName("Form Param tests")
```

dpc-testing/src/main/java/gov/cms/dpc/testing/APIAuthHelpers.java
nickrobison-usds: They are not! Removed.

dpc-api/src/test/java/gov/cms/dpc/api/auth/jwt/JWTUnitTests.java
nickrobison-usds: Fixed!

PR #534, opened 2020-01-13
DPC-760: Add support for ECC keys

nickrobison-usds:
**Why**

In order to help push folks towards a better security posture, we'd like to add support for ECC signing keys for JWTs. This would be in addition to our existing support for RSA keys.

**What Changed**

Updated the PublicKeyHandler class to support both RSA and ECC key types.
All of the unit tests have been updated to parametrize over both key types, to ensure we don't break anything.

The primary change is to verify the OID parameters for the key types to determine whether it's an RSA key or an ECC key and route to the appropriate validation logic. The core of the class hasn't been touched as Java pretty much does the right thing out of the box.

_Note:_ This PR does NOT implement the ECC crypto functions, Java handles all that for us, and the main OID/ASN parsing is done by [BouncyCastle](https://www.bouncycastle.org). This PR is merely about ensuring the ECC keys have the correct curves and get serialized/deserialized correctly.

As part of DPC-759, I've added some sanity checks to the handler, to verify that the keys are the appropriate type and length, which is the main use for the OID verification.

We're now generating ECC keys by default for our integration tests, though we may want to change this in the future.

**Choices Made**

At this point, we're only supporting two ECC curves:

`secp256r1` and `secp384r1`, this is because we have to manually check the OIDs and it seemed simpler to only support the two most common curves.

The Smart Backend Spec says we should only support secp384, but Apple's secure enclave only supports secp256, so we're doing both. I've notified the SMART on FHIR team of our deviation from the spec, and the reasons for doing so.

**Tickets closed**:

DPC-760
DPC-759: Add sanity checking to RSA keys

**Future Work**

**Checklist**


embh:
The new algorithms suggest to me that an SIA (or at least a security checklist) is probably necessary. Has any security documentation been started for this change? cc @DeirdreHolub

DeirdreHolub:
> The new algorithms suggest to me that an SIA (or at least a security checklist) is probably necessary. Has any security documentation been started for this change? cc @DeirdreHolub

I think it will need an SIA.

StewGoin:
> > The new algorithms suggest to me that an SIA (or at least a security checklist) is probably necessary. Has any security documentation been started for this change? cc @DeirdreHolub
> 
> I think it will need an SIA.

I'd concur based on the criteria we use, it's adding security software and changing to some degree the way a core control works (improving it more than anything) -- that said I think this PR if approved by **all three** requested reviewers and a security checklist is sufficient (and updates to any software inventory / dependencies / control descriptions if any are that specific).

BouncyCastle is not new to CMS and the change seems like a reasonable adjustment of the existing security controls rather than a significant change from a control perspective. This feels closer to updating the allowed ciphers in nginx than something more invasive like changing underlying crypto _implementations_.

nickrobison-usds:
@DeirdreHolub @embh @StewGoin I've updated this branch and resolved all the conflicts and code review issues.

Security checklist is here: https://confluenceent.cms.gov/display/DAPC/ECC+Key+Support

Let me know if I missed anything, once we have 3 green checkmarks, I'll merge it!

dpc-api/src/main/java/gov/cms/dpc/api/auth/jwt/PublicKeyHandler.java
```diff
@@ -18,6 +20,13 @@
 
 public class PublicKeyHandler {
 
+    static final ASN1ObjectIdentifier RSA_PARENT = new ASN1ObjectIdentifier("1.2.840.113549");
+    static final ASN1ObjectIdentifier ECC_KEY = new ASN1ObjectIdentifier("1.2.840.10045.2.1");
+    // ECC Curve names are defined here: https://tools.ietf.org/search/rfc4492#section-5.1.1
+    // They're in two separate namespaces, certicom and ansi-x962
+    static final ASN1ObjectIdentifier SECP256_KEY = new ASN1ObjectIdentifier("1.2.840.10045.3.1.7");
+    static final ASN1ObjectIdentifier SECP384_KEY = new ASN1ObjectIdentifier("1.3.132.0.34");
```
embh: Could these use the identifiers in BouncyCastle? For example `SECObjectIdentifiers.secp384r1` for 1.3.132.0.34.

nickrobison-usds: Great find!

PR #555, opened 2020-01-22
DPC-942: Update organization in API from web app

switzersc-usds:
**Why**

When a user changes information about an organization in the web app, we should update the organization's record in the API. 

**What Changed**

* Makes 2 requests to the API when updated: one to update the org and the second to update the org's FHIR Endpoint. 
* Adds `payloadType` to Endpoint in the Organization$submit bundle (required field)
* Updated tests and mocked requests

**Choices Made**

This PR adds the Ruby FHIR client and we use it for the new update functionality. We want to use the FHIR client as much as possible to ensure our API is compatible with community tools and to test community tools ourselves. 

**Tickets closed**:

DPC-942

**Future Work**

* DPC-954 - We want to update the FHIR Endpoint model and its associations with Organization and RegisteredOrganization so that each registered organization has its own FHIR endpoint record. This will make it possible to have multiple endpoints per org and to also keep track of which endpoint belongs to which environment.
* DPC-958 - update docs to say the payloadType is required for Endpoint
* DPC-957 - implement destroy_organization in API Client

**Checklist**


dpc-web/app/services/api_client.rb
```diff
@@ -1,23 +1,53 @@
...
   def create_organization(org)
-    uri_string = base_urls[api_env] + '/Organization/$submit'
+    uri_string = base_url + '/Organization/$submit'
     json = OrganizationSubmitSerializer.new(org).to_json
     post_request(uri_string, json, fhir_headers(golden_macaroon))
     self
   end
 
-  def delete_organization(org); end
+  def get_organization(reg_org)
```
switzersc-usds: We aren't using this method in the app code but it was super helpful when messing around in console and testing with the live API.

dpc-web/app/services/organization_registrar.rb
```diff
@@ -26,18 +25,28 @@ def register_organization(api_env)
...
+  def update_existing_registered_orgs
+    existing_registered_orgs.each do |registered_org|
+      APIClient.new(registered_org.api_env).update_organization(registered_org)
+    end
   end
 
   private
 
-  def no_change?
+  def existing_registered_orgs
```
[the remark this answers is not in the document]
switzersc-usds: Took care of this

dpc-web/spec/services/api_client_spec.rb
```diff
@@ -166,6 +186,71 @@
...
+  describe '#update_organization' do
+    context 'successful request' do
+      it 'uses fhir_client to send org data to API' do
+        org = create(:organization, api_environments: [0])
+        create(:fhir_endpoint, organization: org)
+        reg_org = create(:registered_organization, organization: org, api_env: 'sandbox', api_endpoint_ref: 'Endpoint/12345')
+
+        stub_request(:put, "http://dpc.example.com/Organization/#{reg_org.api_id}").
+          with(
+            body: /#{reg_org.api_id}/,
```
switzersc-usds: Would've included the whole body but ran into an issue where I couldn't replicate the whitespace in the JSON request the FHIR Client makes, and webmock doesn't ignore whitespace. So instead I'm just making sure the body has something in it that is unique to this request.

dpc-web/app/services/fhir_resource_builder.rb
```diff
@@ -0,0 +1,65 @@
...
+      state: org.address_state,
+      country: 'US',
+      use: org.address_use,
+      type: org.address_type
+    )
+  end
+
+  def fhir_endpoint(reg_org)
+    org = reg_org.organization
+    org_endpoint = org.fhir_endpoints.first
```
switzersc-usds: This is a workaround for now. We are going to refactor this model structure soon so that we can fully support multiple FHIR endpoints for an org, rather than sort of supporting it in theory but in practicality only supporting a single one.

dpc-web/app/services/api_client.rb
```diff
@@ -1,23 +1,53 @@
...
+      Rails.logger.warn 'Unsuccessulful request to API'
+      @response_status = org_response.response[:code]
+      @response_body = { 'issue' => [{ 'details' => { 'text' => 'Request error' } }] }
+      false
+    end
+  end
+
+  def delete_organization(org)
+    # DELETE
+    # client.destroy(FHIR::Organization, org.id)
```
switzersc-usds: Will be implemented with DPC-957

dpc-web/app/services/api_client.rb
```diff
@@ -1,23 +1,53 @@
...
+    client.additional_headers = auth_header(delegated_macaroon(reg_org.api_id))
+
+    org_response = client.update(fhir_org, reg_org.api_id)
+    endpoint_response = client.update(fhir_endpoint, fhir_endpoint.id)
+
+    if org_response.response[:code] == '200' && endpoint_response.response[:code] == '200'
+      true
+    else
+      Rails.logger.warn 'Unsuccessulful request to API'
+      @response_status = org_response.response[:code]
```
nickrobison-usds: Should we also check the endpoint_response code?

dpc-web/Gemfile.lock
```diff
@@ -113,12 +118,41 @@ GEM
...
+      addressable (>= 2.3)
+      fhir_dstu2_models (>= 1.0.10)
+      fhir_models (>= 4.0.2)
+      fhir_stu3_models (>= 3.0.1)
+      nokogiri (>= 1.10.4)
+      oauth2 (~> 1.1)
+      rack (>= 1.5)
+      rest-client (~> 2.0)
+      tilt (>= 1.1)
+    fhir_dstu2_models (1.0.10)
```
nickrobison-usds: How difficult would it be to avoid bringing in the DSTU2 models? Not a huge deal, but might be a nice way to reduce the deployment size.

dpc-web/app/services/fhir_resource_builder.rb
nickrobison-usds: Do we have a ticket for this?

dpc-web/Gemfile.lock
switzersc-usds: Unavoidable if we want to use the Ruby FHIR client gem. It's required: https://github.com/fhir-crucible/fhir_client/blob/master/fhir_client.gemspec#L27

dpc-web/app/services/api_client.rb
switzersc-usds: Hm, yeah. We do check for the code in line 34 and only respond with success if both the org and the endpoint codes are 200. However, we set the status to be the org's response code. It would be better not to make two requests here in the `APIClient` but instead have the `OrganizationRegistrar` call the API Client to make both requests -- that way we don't have to muddy the semantics and with two API requests in a single method that suggests it's making just one request.

switzersc-usds: I'll refactor

dpc-web/app/services/fhir_resource_builder.rb
switzersc-usds: DPC-954

switzersc-usds:
@nickrobison-usds - updated

switzersc-usds:
Rebased; will merge on green.

PR #556, opened 2020-01-22
DPC-248: Enable distributed tracing

nickrobison-usds:
**Why**

We'd like to take advantage of the distributed tracing features in New Relic, in order to improve our ability to debug performance issues in production.

**What Changed**

Enabled the `distributed_tracing` flag in the `newrelic.yml` for both the Java and Ruby services.

Also enabled a manual `@Tracing` annotation in the aggregation service, since it's not actually being traced correctly.

Requires ops changes as part of: https://github.com/CMSgov/dpc-ops/pull/115

**Choices Made**

There are some gaps in the tracing coverage that we'll want to address in the future (e.g. the aggregation service isn't traced at all), but we can improve those areas piece-meal once we have a better understanding of useful information in production.

**Tickets closed**:

DPC-248

**Future Work**

DPC-958: Enable tracing for aggregation service
DPC-960: Improve tracing for queue database

**Checklist**


dpc-aggregation/pom.xml
```diff
@@ -120,6 +120,11 @@
             <version>${newrelic.agent.version}</version>
             <type>${newrelic.agent.type}</type>
         </dependency>
+        <dependency>
```
rnagle: Is this duplicative of the dependency defined above it? I see this definition lacks the `version` attribute.

src/main/resources/newrelic.yml
```diff
@@ -9,6 +9,8 @@ common: &default_settings
   log_limit_in_kbytes: 0
   log_daily: false
   log_file_name: STDOUT
+  circuit_breaker:
+    gc_cpu_threshold: 30
```
rnagle: Curious -- how did you arrive at this threshold? Wondering how overhead of tracing will interact with the [ASG scaling policies](https://github.com/CMSgov/dpc-ops/blob/master/terraform/modules/resources/ecs/main.tf#L189). No objection, no blocker here, just thinking out loud.

src/main/resources/server.conf
```diff
@@ -53,6 +53,7 @@ logging {
   }]
   loggers = {
     "liquibase" = WARN
+      "org.apache.http.headers" = DEBUG
```
rnagle: How necessary is this? Have you checked that this does not inadvertently result in logging of sensitive information?

dpc-aggregation/pom.xml
nickrobison-usds: It's not, we need to specify the API as a compile time dependency, otherwise the Tracing annotation isn't found, as the ZIP file is only available at runtime.

src/main/resources/newrelic.yml
nickrobison-usds: In the new relic log files, on dev, the circuit breaker was tripping at 22% GC CPU usage, so I bumped it to 30 to avoid disabling it entirely. I don't think it will interact with the ASG policies as this simply disables tracing at a much lower limit than what would trigger an auto-scale. My goal was to set it at a low threshold and then monitor for any additional tripping.

rnagle: Definitely. 👍

PR #557, opened 2020-01-23
Rename MBI identifier to BENE_ID

nickrobison-usds:
**Why**

In preparation for our migration to MBIs from BB bene_id, we need to rename the identifier enum to avoid conflicts in the future.

**What Changed**

This is a simple IntelliJ refactoring which did a find/replace across the java code.

**Choices Made**

**Tickets closed**:

**Future Work**

DPC-938: Update patient profile to use MBIs instead of bene_ids

**Checklist**


dpc-api/src/test/java/gov/cms/dpc/api/validations/ProfileTests.java
```diff
@@ -58,7 +57,7 @@ void testPatientProfile() throws IOException, URISyntaxException {
 
         // Try for a valid patient
         final Patient validPatient = invalidPatient.copy();
-        validPatient.addIdentifier().setSystem(DPCIdentifierSystem.MBI.getSystem()).setValue("test-mbi");
+        validPatient.addIdentifier().setSystem(DPCIdentifierSystem.BENE_ID.getSystem()).setValue("test-mbi");
```
embh: There's an error on this test. I'm thinking it's because the profile is expecting an MBI, but `FHIRExtractors.getPatientMPI()` is called, looking for bene_id. So we might need to set both temporarily?

nickrobison-usds: The issue was a bad commit, I pulled in the profile changes by mistake, those will come in a new commit, this is supposed to be just be creating space in the enum.

PR #564, opened 2020-01-28
adjust token ttl by 30 seconds

DeirdreHolub:
**Why**

We have evidence that auth request tokens created on the jenkins server during smoke tests are sometimes exceeding the 5 minute max TTL limit on the server. This can only happen if the clock on the jenkin's server is ahead of the clock on api server by longer than an amount of time (possibly milliseconds) that the Java time functions will calculate as >= 1 second. 

Since this is an intermittent time sync issue, Marissa checked the time sync setup of the servers for me. All servers based on the GDIT gold image are synced using NTP, and docker images use the system clock. 

**What Changed**

I changed the time TTL on token creation to be max ttl allowed (5 mins) - 30 seconds. (The subtraction value is probably much higher than required.)

**Choices Made**

Code exists in client to refresh access tokens appropriately, so auth token requests will only fail if they are created with an expiration time that is > server time + 5 mins. This is a problem that any client can experience and must handle; therefore, it made more sense to me to change the client. I like servers to be absolute, which makes for more predictable debugging.

**Tickets closed**:

DPC-973

**Future Work**

**Checklist**

None of the following apply:


StewGoin:
Do we know why we've got a >5min delta within our own systems' timekeeping?
I'm 100% ok allowing some grace time but I am curious how ours are skewed that far to be honest.

DeirdreHolub:
> Do we know why we've got a >5min delta within our own systems' timekeeping?
> I'm 100% ok allowing some grace time but I am curious how ours are skewed that far to be honest.

The clock difference only has to be 1 second for this issue to occur. I updated the notes to try to make it clearer.

DeirdreHolub:
> My one concern with this change is that we now have different behavior between the /auth and /validate endpoints.
> 
> We either need to add some drift accounting to the validate endpoint, or remove the drift compensation from the auth endpoint.

I walk through the code here, and we don't have any drift compensation anywhere in the server.

PR #572, opened 2020-01-29
DPC-966: Database migrations are not logged

ronaldheft-usds:
**Why**

We discovered during a recent deployment that there is no output from the database migrations when run during server startup.

As part of adding the log messages for database migrations, logging in general was investigated, and several improvements were made.

**What Changed**

* Fixed liquibase migrations not being logged by increasing the log level for liquibase
* Fixed an issue where `INFO` and above log messages could be dropped by Dropwizard
* Fixed several issues related to extra log messages when running CI tests
  * Added a standard `ci.application.conf` file that sets to log level to `ERROR` for running CI tests
  * Configured all instances of `DropwizardTestSupport` to use the new `ci.application.conf` file
  * Passed in the config file to all instances of CLI commands in tests (they run in a separate thread, and reload the configuration file while called)
  * Added a `LoggingInterceptor` to all HAPI FHIR testing contexts to disable request logging in tests
    * This could probably be cleaned up and refactored a bit, as our usage of creating new clients occurs in a lot of places
  * With these changes, we should only have in our CI logs:
    * Maven output
    * Logging for ERROR messages (which there are still a decent amount of, since we have a number of tests that purposely trigger errors)
  * I was able to reduce the log messages from around ~9500 lines to around ~6500 lines
* Updated the `.java-version` file to allow for newer Java 11 JDKs (had to do this when I rebuilt my Java environment from working on Blue Button)

**Choices Made**

* I discovered Dropwizard could drop log message while [reading over the docs](https://www.dropwizard.io/en/stable/manual/configuration.html)
  * Dropwizard has a setting called `discardingThreshold`
  * The default reads as: `When the blocking queue has only the capacity mentioned in discardingThreshold remaining, it will drop events of level TRACE, DEBUG and INFO, keeping only events of level WARN and ERROR. If no discarding threshold is specified (-1), then a default of queueSize / 5 (logback’s default ratio) is used. To keep all events, set discardingThreshold to 0.`
  * Since we always want to preserve auditability of our application, I changed the `discardingThreshold` to `0` per the docs
* I ended up tackling the extra log messages because adding liquibase logs to the output dramatically increased our log messages during CI

**Tickets closed**:

DPC-966: Dropwizard database migrations are not logged
DPC-897: Reduce logging during build to avoid Travis death

**Future Work**

**Checklist**


embh:
Are the HAPI FHIR debug messages useful? For example:
```20:00:00.064 [main] DEBUG ca.uhn.fhir.context.ModelScanner - Scanning datatype class: org.hl7.fhir.dstu3.model.CodeableConcept```
Wondering if we should consider setting that to `ERROR`.

ronaldheft-usds:
> Are the HAPI FHIR debug messages useful? For example:
> `20:00:00.064 [main] DEBUG ca.uhn.fhir.context.ModelScanner - Scanning datatype class: org.hl7.fhir.dstu3.model.CodeableConcept`
> Wondering if we should consider setting that to `ERROR`.

They're not, but they appear to be using a separate logger, since the log format is not in JSON. This PR targets all the log messages that were formatted in JSON from our recent logging configuration change.

If we want to do another pass on getting the log messages down, I'm all for it, but I'd prefer doing so in a separate PR as the main purpose of this PR is to fix the database changes not being logged (and we need that to verify we fix the server shutdown issue we're experiencing).

ronaldheft-usds:
Managed to reduce the logs by another ~1200 lines (for a total of ~4200 lines) by changing some log levels on Maven. The biggest areas we can cut down on log messages appear to be:

* The Maven shade JAR plugin.
  * Didn't see any quick ways to cut down on the output from that plugin.
* HAPI FHIR output
  * Seems like the library doesn't always consume our logging configuration (or it would be formatted as JSON)
  * Perhaps adding some logback.xml files in our projects would help? The HAPI documentation doesn't appear to be super helpful
* Cleaning up unit tests that output expected errors
  * This will probably have to be done on a test-by-test basis to verify the error is expected

PR #585, opened 2020-02-04
DPC-1010 Vendor Id Column

Sun-Mountain:
**Why**

Breaking up the vendor workflow ticket into smaller, more manageable sized tickets. This was just adding a column to the `organizations` table.

I also fixed a line the DPC-APP's README.md file.

**What Changed**

Added a column for vendor ids that will be created when an external user creates a vendor account. (The next ticket will be adding the generated id number to relevant accounts.)

Corrected the app's main README file to include the correct `docker-compose up` instructions.

**Choices Made**

`venid` to keep it short but recognizable and different from `npi` and `id` columns.

**Tickets closed**:

DPC 1010

**Future Work**

Will continue with the vendor workflow ticket DPC-997.

**Checklist**

### DIFF
--- a/dpc-web-test.sh
+++ b/dpc-web-test.sh
@@ -12,7 +12,7 @@ make website
 
 # Run the tests
 docker-compose -f dpc-web/docker-compose.yml up start_core_dependencies
-docker-compose -f dpc-web/docker-compose.yml run web bundle exec rails db:create db:migrate db:seed
+docker-compose -f dpc-web/docker-compose.yml run web bundle exec rails db:create db:migrate RAILS_ENV=test
 docker-compose -f dpc-web/docker-compose.yml run web bundle exec rails spec
 
 echo "┌──────────────────────────────────┐"
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ The application (along with all required dependencies) can be automatically star
 The individual services can be started (along with their dependencies) by passing the service name to the `up` command.
 
 ```bash
-docker-compose up {db,dpc-aggregation,dpc-attribution,dpc-api}
+docker-compose up {db,aggregation,attribution,api}
 ``` 
 
 By default, the Docker containers start with minimal authentication enabled, meaning that some functionality (such as extracting the organization_id from the access token) will not work as expected and always return the same value.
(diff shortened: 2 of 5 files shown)


dpc-web/db/migrate/20200203201716_add_vendor_id_to_organizations.rb
```diff
@@ -0,0 +1,5 @@
+class AddVendorIdToOrganizations < ActiveRecord::Migration[5.2]
+  def change
+    add_column :organizations, :venid, :string
```
switzersc-usds: Let's not shorten names of things unless a) the alternative is quite long and b) the shortened version is clear. In this case, `vendor_id` isn't long and it's much clearer than `venid`.

dpc-web/db/migrate/20200204152555_change_vendor_id_column_name.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeVendorIdColumnName < ActiveRecord::Migration[5.2]
```
nickrobison-usds: Do we really need a second migration to rename the column? Since the change hasn't been deployed yet, I think we can just leave this as a single column add.

dpc-web-test.sh
```diff
@@ -12,7 +12,7 @@ make website
 
 # Run the tests
 docker-compose -f dpc-web/docker-compose.yml up start_core_dependencies
-docker-compose -f dpc-web/docker-compose.yml run web bundle exec rails db:create db:migrate db:seed
+docker-compose -f dpc-web/docker-compose.yml run web bundle exec rails db:create db:migrate RAILS_ENV=test
```
nickrobison-usds: Can you explain why this change is needed now? Has it not been working previously?

Sun-Mountain: Travis didn't pass the last check.

The error it threw was 

```
** [NewRelic] FATAL : Config file at /dpc-web/config/newrelic.yml doesn't include a 'test' section!
** [NewRelic] FATAL : Config file at /dpc-web/config/newrelic.yml doesn't include a 'test' section!
Migrations are pending. To resolve this issue, run:
        bin/rails db:migrate RAILS_ENV=test
```

@switzersc-usds gave me the code to fix it. As for why, I can't answer that.

nickrobison-usds: Ok. Sounds good. Thanks!

switzersc-usds: This wasn't necessary before because the development DB and the test DB were the same. When we updated the db config to use separate DBs for dev vs test, we needed to update travis to make sure the test database gets migrated before running tests.

switzersc-usds: Also removed `db:seed` because that's not necessary for tests.

PR #613, opened 2020-02-18
DPC-1075: Catch and rethrow DataFormatException

nickrobison-usds:
**Why**

In our Splunk logs we've been seeing some `500` related to users submitting invalid FHIR resources to our endpoints. Currently, these errors get returned to the user as a generic 500 exception, rather than in a more useful format.

This is similar to the work in #582 as we're slowly purging the system of generic exceptions.

**What Changed**

Updated the FHIRHandler to catch DataFormatExceptions and rethrow them as `400` errors with a message, rather than simply leaking the raw exception as a 500 error.

We need to do this because the custom FHIR exception handlers aren't loaded at this point in the request chain so we have to handle things manually.

Also updated the unit tests to fully cover the class.

**Choices Made**

**Tickets closed**:

DPC-1075

**Future Work**

**Checklist**


nickrobison-usds:
@whuang85 Not a dumb question! It's something the FHIR community has been talking a lot about. We don't really have a way to order data loading for two reasons.

1. We don't control the order in which organizations send us their providers.
1. We don't have a sortable that we can use, provider identifiers from CMS are not ordered (at least to my knowledge).

Does that answer your question?

PR #627, opened 2020-02-20
Fix broken token creation in sandbox

ronaldheft-usds:
**Why**

The sandbox API is currently configured to use the production API URL. This is causing a number of problems with token creation.

**What Changed**

The configuration URL of the sandbox API has been changed from `https://dpc.cms.gov/api` to `https://sandbox.dpc.cms.gov/api`.

**Choices Made**

This change will break token creation for users who were using the incorrect URL, but working URL. [In our documentation](https://dpc.cms.gov/docs#authentication-and-authorization) we say to use `https://sandbox.dpc.cms.gov/api/v1/Token/auth` for the aud claim, but this was never actually working in sandbox. Users had to use `https://dpc.cms.gov/api/v1/Token/auth` as the aud claim for token authentication to work.

**This is a change we will need to communicate to our Google Group.** It's going to fix it for some users, as it matches what's in our documentation, but it's going to break it for others who were using the incorrect URL.

We need to make this change, because we have locked down the `https://dpc.cms.gov/api/` domain in preparation for actually launching production. We can no longer have users make API requests to sandbox through the production URL.

**Tickets closed**:

**Future Work**

* Communicate this on the Google Group.
* Verify the website is not also using the incorrect value for token creation in sandbox. @switzersc-usds @Sun-Mountain 

**Checklist**

### DIFF
--- a/dpc-api/src/main/resources/prod-sbx.application.conf
+++ b/dpc-api/src/main/resources/prod-sbx.application.conf
@@ -1,6 +1,6 @@
 dpc.api {
 
-    publicURL = "https://dpc.cms.gov/api" # The root URL at which the application is accessible, if necssary, include the port, do not include the application version
+    publicURL = "https://sandbox.dpc.cms.gov/api" # The root URL at which the application is accessible, if necssary, include the port, do not include the application version
 
     server {
         applicationContextPath = "/api"


dpc-api/src/main/resources/prod-sbx.application.conf
```diff
@@ -1,6 +1,6 @@
 dpc.api {
 
-    publicURL = "https://dpc.cms.gov/api" # The root URL at which the application is accessible, if necssary, include the port, do not include the application version
+    publicURL = "https://sandbox.dpc.cms.gov/api" # The root URL at which the application is accessible, if necssary, include the port, do not include the application version
```
switzersc-usds: In the website for token creation, we use this endpoint: `ENV["API_METADATA_URL_SANDBOX"] + '/Token'`. 

So, we'll just want to make sure the API_METADATA_URL_SANDBOX is set correctly. I believe it should be set to `https://sandbox.dpc.cms.gov/api/v1`, correct? Which has `v1` whereas the `publicURL` here does not.

ronaldheft-usds: `v1` is added elsewhere in the code, and `ENV["API_METADATA_URL_SANDBOX"]` was set incorrect initially (it's why users couldn't upload public keys after rolling out Akamai), but it's now using the correct URL in sandbox (and was verified yesterday).

switzersc-usds: Great, then this should be good to go from the website side of things.

PR #634, opened 2020-02-22
DPC-34 since support

RickHawesUSDS:
**Why**

Add support to the FHIR's Bulk Export _since parameter. 

**What Changed**

The export chain of components and structures had to be extended to ensure all batches use the same upper and lower bound. Consistency of these values is essential to ensure consecutive jobs do not have missing or duplicate resources. However, the overall structure of the export job has not changed. Batching and restarting a job remains the same. 

- The export API fetches the BFD's transactionTime to use as the upper bound of 
- JobBatch added since and transactionTime fields
- The AggregatorEngine and ResourceFetcher accept the new values
- The BBClient methods have a lastUpdated parameter added

One consequence of this that API now uses the BBClient module directly, whereas only the Aggregator used this module before. Again, this choice was necessary to keep transactionTime consistent between batches. About half of the file changes were made to hook this module up. 

**Choices Made**

I've added JobCompletionRecord extensions for job start and complete time. The FHIR community has encouraged experimentation with extensions. 

**Tickets closed**:

DPC-34: 

**Future Work**

CareJourney and others have pointed out that the _since parameter solves a major issue with data size, but creates a problem for new patients that are added to a roster. Future work should work on this issue. 

**Checklist**
_since has been added to our Swagger and FHIR documentation. 

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,6 +51,9 @@ services:
       - ENV=local
       - JACOCO=${REPORT_COVERAGE}
       - exportPath=/app/data
+      - BFD_HASH_ITER=${BFD_HASH_ITER}
+      - BFD_HASH_PEPPER=${BFD_HASH_PEPPER}
+      - BFD_URL=${BFD_URL}
       - JVM_FLAGS=-Ddpc.api.authenticationDisabled=${AUTH_DISABLED:-false}
     depends_on:
       - attribution
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -46,14 +46,13 @@ services:
     ports:
       - "3002:3002"
       - "9903:9900"
+    env_file:
+      - ./ops/config/decrypted/local.env
     environment:
       - attributionURL=http://attribution:8080/v1/
       - ENV=local
       - JACOCO=${REPORT_COVERAGE}
       - exportPath=/app/data
-      - BFD_HASH_ITER=${BFD_HASH_ITER}
-      - BFD_HASH_PEPPER=${BFD_HASH_PEPPER}
-      - BFD_URL=${BFD_URL}
       - JVM_FLAGS=-Ddpc.api.authenticationDisabled=${AUTH_DISABLED:-false}
     depends_on:
       - attribution
--- a/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
+++ b/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
@@ -232,7 +232,14 @@ public class AggregationEngine implements Runnable {
      */
     private Flowable<JobQueueBatchFile> completeResource(JobQueueBatch job, String patientID, ResourceType resourceType) {
         // Make this flow hot (ie. only called once) when multiple subscribers attach
-        final var fetcher = new ResourceFetcher(bbclient, job.getJobID(), job.getBatchID(), resourceType, operationsConfig);
+        final var fetcher = new ResourceFetcher(
+                bbclient,
+                job.getJobID(),
+                job.getBatchID(),
+                resourceType,
+                job.getSince().orElse(null),
+                job.getTransactionTime(),
+                operationsConfig);
         final Flowable<Resource> mixedFlow = fetcher.fetchResources(patientID);
         final var connectableMixedFlow = mixedFlow.publish().autoConnect(2);
 
--- a/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
+++ b/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
@@ -16,6 +16,7 @@ import org.slf4j.LoggerFactory;
 
 import java.security.GeneralSecurityException;
 import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
@@ -85,6 +86,7 @@ class ResourceFetcher {
      */
     private List<Resource> fetchAllBundles(Bundle firstBundle, String fetchId) {
         final var resources = new ArrayList<Resource>();
+        checkBundleTransactionTime(firstBundle);
         addResources(resources, firstBundle);
 
         // Loop until no more next bundles
@@ -92,6 +94,7 @@ class ResourceFetcher {
         while (bundle.getLink(Bundle.LINK_NEXT) != null) {
             logger.debug("Fetching next bundle {} from BlueButton for {}", resourceType.toString(), fetchId);
             bundle = blueButtonClient.requestNextBundleFromServer(bundle);
+            checkBundleTransactionTime(bundle);
             addResources(resources, bundle);
         }
 
@@ -126,13 +129,7 @@ class ResourceFetcher {
     private Bundle fetchFirst(String mbi) {
         Patient patient = fetchPatient(mbi);
         var beneId = getBeneIdFromPatient(patient);
-        // Note: FHIR bulk spec says that since is exclusive and transactionTime is inclusive
-        // It is also says that all resources should not have lastUpdated after the transactionTime.
-        // This is true for the both the since and the non-since cases.
-        // BFD will include resources that do not have a lastUpdated if there isn't a complete range.
-        final var lastUpdated = since != null ?
-                new DateRangeParam().setUpperBoundInclusive(Date.from(transactionTime.toInstant())).setLowerBoundExclusive(Date.from(since.toInstant())) :
-                new DateRangeParam().setUpperBoundInclusive(Date.from(transactionTime.toInstant()));
+        final var lastUpdated = formLastUpdatedParam();
         switch (resourceType) {
             case Patient:
                 return blueButtonClient.requestPatientFromServer(beneId, lastUpdated);
@@ -214,5 +211,41 @@ class ResourceFetcher {
                 .setDetails(new CodeableConcept().setText(details))
                 .setLocation(patientLocation);
         return outcome;
+    }
+
+    /**
+     * Form a date range for the lastUpdated parameter for this export job
+     *
+     * @return a date range for this job
+     */
+    private DateRangeParam formLastUpdatedParam() {
+        // Note: FHIR bulk spec says that since is exclusive and transactionTime is inclusive
+        // It is also says that all resources should not have lastUpdated after the transactionTime.
+        // This is true for the both the since and the non-since cases.
+        // BFD will include resources that do not have a lastUpdated if there isn't a complete range.
+        return since != null ?
+                new DateRangeParam()
+                        .setUpperBoundInclusive(Date.from(transactionTime.toInstant()))
+                        .setLowerBoundExclusive(Date.from(since.toInstant())) :
+                new DateRangeParam()
+                        .setUpperBoundInclusive(Date.from(transactionTime.toInstant()));
+    }
+
+    /**
+     * Check the transaction time of the bundle against the transaction time of the export job
+     *
+     * @param bundle to check
+     */
+    private void checkBundleTransactionTime(Bundle bundle) {
+        if (bundle.getMeta() == null || bundle.getMeta().getLastUpdated() == null) return;
+        final var bundleTransactionTime = bundle.getMeta().getLastUpdated().toInstant().atOffset(ZoneOffset.UTC);
+        if (bundleTransactionTime.isBefore(transactionTime)) {
+            logger.info("About to throw an BFD Runtime: bundle {}, job {}", bundleTransactionTime, transactionTime);
+            /**
+             * See BFD's RFC0004 for a discussion on why this type error may occur.
+             * Note: Retrying after a delay may fix this problem.
+             */
+            throw new RuntimeException("BFD's transaction time regression");
+        }
     }
 }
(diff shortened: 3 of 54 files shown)


msnook:
In the current [BCDA PR](https://github.com/CMSgov/bcda-app/pull/459) to add `_since` support, BCDA is returning a `Bad Request` in the event the `_since` parameter is set to anything prior to 2/13/2020 (to reduce confusion in expected data).  Is DPC planning to do anything similar?

RickHawesUSDS:
> In the current [BCDA PR](https://github.com/CMSgov/bcda-app/pull/459) to add `_since` support, BCDA is returning a `Bad Request` in the event the `_since` parameter is set to anything prior to 2/13/2020 (to reduce confusion in expected data). Is DPC planning to do anything similar?

I don't think so. _Since requests are before 2/13 are valid per the FHIR spec. What advantages do you see from erroring? You might consider asking the community if want to do this.

dpc-queue/src/main/java/gov/cms/dpc/queue/JobQueueCommon.java
```diff
@@ -21,19 +22,17 @@ public JobQueueCommon(int batchSize) {
     }
 
     @Override
-    public UUID createJob(UUID orgID, String providerID, List<String> patients, List<ResourceType> resourceTypes) {
+    public UUID createJob(UUID orgID, String providerID, List<String> patients, List<ResourceType> resourceTypes, OffsetDateTime since, OffsetDateTime transactionTime) {
```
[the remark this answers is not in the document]
RickHawesUSDS: Not fixing this issue.

dpc-aggregation/src/main/resources/local.application.conf
```diff
@@ -11,14 +11,21 @@ dpc.aggregation {
...
+    serverBaseUrl = "https://prod-sbx.bfdcloud.net/v1/fhir"
+  }
 
   exportPath = "/app/data"
 
   logging {
     loggers {
-      "gov.cms.dpc" = DEBUG
-      "org.hibernate.SQL" = TRACE
+      "gov.cms.dpc" = INFO
```
RickHawesUSDS: This reduces the level of noise in the aggregator logs.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
```diff
@@ -123,26 +133,26 @@
...
+                return blueButtonClient.requestCoverageFromServer(beneId, lastUpdated);
             default:
                 throw new JobQueueFailure(jobID, batchID, "Unexpected resource type: " + resourceType.toString());
         }
     }
 
     private Patient fetchPatient(String mbi) {
-        Bundle patients = null;
+        Bundle patients;
         try {
```
RickHawesUSDS: Style change. null is redundant.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
```diff
@@ -203,7 +203,14 @@ private void calculateFileMetadata(JobQueueBatch job) {
...
         // Make this flow hot (ie. only called once) when multiple subscribers attach
-        final var fetcher = new ResourceFetcher(bbclient, job.getJobID(), job.getBatchID(), resourceType, operationsConfig);
+        final var fetcher = new ResourceFetcher(
+                bbclient,
+                job.getJobID(),
+                job.getBatchID(),
+                resourceType,
+                job.getSince().orElse(null),
+                job.getTransactionTime(),
+                operationsConfig);
```
RickHawesUSDS: since and transactionTime are new.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
```diff
@@ -123,26 +133,26 @@
...
-    private Resource fetchFirst(String mbi) {
+    private Bundle fetchFirst(String mbi) {
         Patient patient = fetchPatient(mbi);
+        var beneId = getBeneIdFromPatient(patient);
+        final var lastUpdated = formLastUpdatedParam();
         switch (resourceType) {
             case Patient:
-                return patient;
+                return blueButtonClient.requestPatientFromServer(beneId, lastUpdated);
             case ExplanationOfBenefit:
```
RickHawesUSDS: `requestPatientFromServer` now does a search for so we can pass `lastUpdated`

dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/client/BlueButtonClientImpl.java
```diff
@@ -217,25 +219,29 @@ public String hashMbi(String mbi) throws GeneralSecurityException {
...
-        final Bundle bundle = query.count(config.getResourcesCount())
+        final Bundle bundle = query
+                .count(config.getResourcesCount())
+                .lastUpdated(lastUpdated)
                 .returnBundle(Bundle.class)
                 .execute();
 
         // Case where patientID does not exist at all
-        if(!bundle.hasEntry()) {
+        if(!bundle.hasEntry() && lastUpdated == null) {
```
RickHawesUSDS: This change is kind of subtle. When `lastUpdated` is specified an empty bundle does not mean that the beneId doesn't exist, rather it may mean that a resource didn't match the `lastUpdated` parameter.

msnook:
> > In the current [BCDA PR](https://github.com/CMSgov/bcda-app/pull/459) to add `_since` support, BCDA is returning a `Bad Request` in the event the `_since` parameter is set to anything prior to 2/13/2020 (to reduce confusion in expected data). Is DPC planning to do anything similar?
> 
> I don't think so. _Since requests are before 2/13 are valid per the FHIR spec. What advantages do you see from erroring? You might consider asking the community if want to do this.

Thanks.  This was a decision that we made on BCDA to reduce confusion.  Yes, dates before 2/13 are valid per FHIR spec, but we've decided to return an `OperationOutcome` indicating that our application only supports `_since` date of 2/13 or later. Still in review, so this may change, but just wanted to indicate that this is the direction we are leaning.

msnook:
> > > In the current [BCDA PR](https://github.com/CMSgov/bcda-app/pull/459) to add `_since` support, BCDA is returning a `Bad Request` in the event the `_since` parameter is set to anything prior to 2/13/2020 (to reduce confusion in expected data). Is DPC planning to do anything similar?
> > 
> > 
> > I don't think so. _Since requests are before 2/13 are valid per the FHIR spec. What advantages do you see from erroring? You might consider asking the community if want to do this.
> 
> Thanks. This was a decision that we made on BCDA to reduce confusion. Yes, dates before 2/13 are valid per FHIR spec, but we've decided to return an `OperationOutcome` indicating that our application only supports `_since` date of 2/13 or later. Still in review, so this may change, but just wanted to indicate that this is the direction we are leaning.

@RickHawesUSDS @nickrobison-usds 
From the spec:
> If the client explicitly asks for export of resources that the bulk data server doesn’t support, the server SHOULD return details via an OperationOutcome resource in an error response to the request.
I think our approach (returning error response if `_since` is specified with date prior to 2/13/2020) is in line with FHIR spec.  Thoughts?

dpc-queue/src/main/java/gov/cms/dpc/queue/JobQueueCommon.java
```diff
@@ -21,19 +22,17 @@ public JobQueueCommon(int batchSize) {
...
-                .map(patientBatch -> this.createJobBatch(jobID, orgID, providerID, patientBatch, resourceTypes))
-                .toList()
-                .blockingGet();
-
-        // Expect a single empty job when no patients passed
-        if ( jobBatches.isEmpty() && patients.isEmpty() ) {
-            jobBatches.add(this.createJobBatch(jobID, orgID, providerID, Collections.emptyList(), resourceTypes));
-        }
+        // Add a single empty job when no patients or since is less or equal to transactionTime
+        List<JobQueueBatch> jobBatches = patients.isEmpty() || (since != null && !transactionTime.isAfter(since)) ?
```
nickrobison-usds: Probably a stylistic change, but given this ternary spreads across multiple lines, maybe it makes sense to split this into an if/else statement.

nickrobison-usds:
@msnook I think the portion of the spec you linked to refers to requesting unsupported resource types (e.g. Appointment, Observation, etc), not whether or not the date range is valid. You can still have a server specific limitation but it should be noted in the capability statement and return a _422_.

RickHawesUSDS:
@msnook 

In my conversation with Dan

embh:
From the description:
> The export API fetches the BFD's transactionTime to use as the upper bound of

To use as the upper bound of the time range?

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
```diff
@@ -250,12 +259,12 @@ public Response export(@ApiParam(hidden = true)
                            @QueryParam("_type") String resourceTypes,
                            @ApiParam(value = "Output format of requested data", allowableValues = FHIR_NDJSON, defaultValue = FHIR_NDJSON)
                            @QueryParam("_outputFormat") String outputFormat,
-                           @ApiParam(value = "Request data that has been updated after the given point. (Not implemented yet)", hidden = true)
+                           @ApiParam(value = "Resources updated after this period will be included in the response. Must be a FHIR date-time.")
```
embh:
```suggestion
                           @ApiParam(value = "Resources updated after this time will be included in the response. Must be a FHIR date-time.")
```
This should probably be something that implies `instant` rather than a date range.

RickHawesUSDS: This text is actually from a version of the FHIR spec. I'm replacing it with the current version which is more precise. Also, the real change I needed to do was move from String to a InstantDt

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
```diff
@@ -65,17 +78,12 @@
...
             logger.debug("Fetching first {} from BlueButton for {}", resourceType.toString(), fetchId);
-            final Resource firstFetched = fetchFirst(mbi);
-            if (ResourceType.Coverage.equals(resourceType) || ResourceType.ExplanationOfBenefit.equals(resourceType)) {
-                return fetchAllBundles((Bundle)firstFetched, fetchId);
-            } else {
-                logger.debug("Done fetching {} for {}", resourceType.toString(), fetchId);
-                return List.of(firstFetched);
-            }
+            final Bundle firstFetched = fetchFirst(mbi);
+            return fetchAllBundles(firstFetched, fetchId);
```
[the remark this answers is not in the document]
RickHawesUSDS: Good question. Every job, including those ** without** a since parameter, is now time-bound by the transaction time. This is done for correctness. So every request to the BFD is now a search with a `lastUpdated` set. This change has the happy side-effect of making the getPatientFromServer API have the same signature as the Coverage and EOB methods. So, the special case if statements are nor longer needed.

dpc-api/src/main/java/gov/cms/dpc/api/DPCAPIModule.java
```diff
@@ -175,6 +176,7 @@ SessionFactory provideSessionFactory(DPCAuthManagedSessionFactory factory) {
 
     @Provides
     @Singleton
+    @Named("attribution")
```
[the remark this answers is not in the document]
RickHawesUSDS: I see the advantage of your suggestion. However, that is a big change affecting many files. I suggest that I do this change in a follow-on PR.

RickHawesUSDS:
@ronaldheft-usds Good catch. I'll submit the OPS PR.

docker-compose.yml
```diff
@@ -54,6 +54,9 @@ services:
       - ENV=local
       - JACOCO=${REPORT_COVERAGE}
       - exportPath=/app/data
+      - BFD_HASH_ITER=${BFD_HASH_ITER}
+      - BFD_HASH_PEPPER=${BFD_HASH_PEPPER}
+      - BFD_URL=${BFD_URL}
```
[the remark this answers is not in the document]
RickHawesUSDS: Roger.

docker-compose.yml
```diff
@@ -46,6 +46,8 @@ services:
     ports:
       - "3002:3002"
       - "9903:9900"
+    env_file:
```
ronaldheft-usds: Looks great!

dpc-api/src/main/java/gov/cms/dpc/api/DPCAPIModule.java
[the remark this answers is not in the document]
ronaldheft-usds: Sounds good to me.

dpc-common/src/main/java/gov/cms/dpc/common/models/JobCompletionModel.java
```diff
@@ -75,25 +76,32 @@ public Integer getCount() {
      * An extension field for additional information in an {@link OutputEntry}.
      */
     @JsonInclude(Include.NON_EMPTY)
-    public static class OutputEntryExtension {
+    public static class FhirExtension {
```
embh: Good call on reusing/renaming this!

PR #640, opened 2020-02-24
Switch to MBIs and use BFD prod sandbox environment

embh:
**Why**
DPC needs to accept MBIs from users when Patients are added, and use MBIs for retrieving BFD data.

**What Changed**
* BFD server configurable with `BFD_URL` environment variable. Prod sandbox is used in `local.env`, along with the correct hashing pepper and iterations
* Requests to BFD for Coverage and EOB now preceded by requests for Patient by MBI, to get bene ID
* Test data updated to use MBIs and bene IDs from BFD prod-sbx
* MBI added to IG Patient profile

**Tickets closed**:
[DPC-962](https://jiraent.cms.gov/browse/DPC-962) - Get bene ID from BFD by MBI before EOB and Coverage requests
[DPC-963](https://jiraent.cms.gov/browse/DPC-963) - BFD MBI hash pepper and iterations variables
[DPC-938](https://jiraent.cms.gov/browse/DPC-938) - MBIs in Patient profile
[DPC-993](https://jiraent.cms.gov/browse/DPC-993) - Secrets management in repository (partly completed in another PR - this portion is for the BFD JKS)

**Checklist**


dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
```diff
@@ -55,87 +57,114 @@
...
      * Based on resourceType, fetch a resource or a bundle of resources.
      *
-     * @param patientID of the resource to fetch
+     * @param mbi of the resource to fetch
      * @return either a single resource or the first bundle of resources
      */
-    private Resource fetchFirst(String patientID) {
+    private Resource fetchFirst(String mbi) {
+        Patient patient = fetchPatient(mbi);
+        String beneId;
```
nickrobison-usds: Is there a reason for declaring this outside of the switch statement? If possible, I'd prefer to do a local allocation to a final, rather than have an uninitialized variable at the top.

dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/client/BlueButtonClientImpl.java
```diff
@@ -89,6 +89,18 @@ public Patient requestPatientFromServer(String patientID) throws ResourceNotFoun
...
 
+    /**
+     * Hashes MBI and queries Blue Button server for patient data.
+     *
+     * @param mbi The MBI
+     * @return {@link Bundle} A FHIR Bundle of Patient resources
+     */
+    @Override
+    public Bundle requestPatientFromServerByMbi(String mbi) throws ResourceNotFoundException, GeneralSecurityException {
+        String mbiHash = hashMbi(mbi);
```
nickrobison-usds: I believe I ran into an issue with the secret key factory not being thread safe, were you able to determine that as well? I believe there was a commit addressing that in my original branch.

nickrobison-usds:
@embh Looking at the code coverage numbers, it looks like the coverage on the `ResourceFetcher` and `RosterUtils` classes dropped a bit, would you make seeing if we can beef up the unit tests a bit?

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
embh: The scope of a `switch` is the whole block. We can declare it under the EOB `case`, but it couldn't be `final`, unless I'm misunderstanding the suggestion?

src/test/resources/bb-test-data/patient/-20140000009893.xml
```diff
@@ -9,12 +9,16 @@
    </extension>
    <identifier>
       <system value="https://bluebutton.cms.gov/resources/variables/bene_id"/>
-      <value value="20140000009893"/>
+      <value value="-20140000009893"/>
    </identifier>
    <identifier>
-      <system value="https://bluebutton.cms.gov/resources/identifier/hicn-hash"/>
+      <system value="http://bluebutton.cms.hhs.gov/identifier#hicnHash"/>
```
RickHawesUSDS: This change to this system value doesn't look right. BTW: BFD is removing this identifier in its latest response, so the best way to fix this may be to just remove the whole HICN identifier clause.

embh: That is how we've seen it in BFD responses. Also formatted that way here: https://github.com/CMSgov/beneficiary-fhir-data/blob/a995592c40d77e3974ade09eb5017492e0865608/apps/bfd-server/dev/data-model.md

I'd propose a new ticket to remove HICN entirely.

embh:
> @embh Looking at the code coverage numbers, it looks like the coverage on the `ResourceFetcher` and `RosterUtils` classes dropped a bit, would you make seeing if we can beef up the unit tests a bit?

I'm confused about why Code Climate is saying that. If I look at the stats for the files, it says ResourceFetcher has 61/62 lines covered, and RosterUtils has 24/24. Is it expecting more tests on those same lines?

nickrobison-usds:
@embh Hmm.... If that's the case, them I'm fine merging it. Thanks for the extra info.

PR #652, opened 2020-02-26
add ability to configure practitioner limit per environment

whuang85:
**Why**

For our early pilot partners, we need to have the ability to limit the number of providers they can associate with a given organization. This will need to be configurable on a per-environment level as we only want to put in restrictions in production.

**What Changed**

Added a configuration variable in the aggregation configuration that sets a limit to determine when submitting should create a provider or not

**Choices Made**

Justify changes made and any reasons for why a given path was taken, as opposed to an alternative option.

**Tickets closed**:

https://jiraent.cms.gov/browse/DPC-1083

**Future Work**

**Checklist**


whuang85:
Not sure what the FHIR documentations are and where they are or if they need to be updated based on this PR

dpc-attribution/src/main/java/gov/cms/dpc/attribution/resources/v1/PractitionerResource.java
```diff
@@ -70,20 +73,30 @@
     @ApiOperation(value = "Register provider", notes = "FHIR endpoint to register a provider with the system." +
             "<p>Each provider must have a metadata Tag with the responsible Organization ID included." +
             "If not, we'll reject it." +
-            "<p> If a provider is already registered with the Organization, an errorr is thrown.")
+            "<p> If a provider is already registered with the Organization, an error is thrown." +
+            "<p> If a provider has already registered providers past the provider limit, then return 304")
     @ApiResponses(value = {
             @ApiResponse(code = 201, message = "New resource was created"),
+            @ApiResponse(code = 304, message = "Resource was not created")
```
nickrobison-usds: I'm thinking more about the 304 return code, I'm wondering if it would more sense to return a 429 or a 422 error, not sure 304 is very informative. Thoughts @DeirdreHolub @embh?

dpc-attribution/src/main/java/gov/cms/dpc/attribution/resources/v1/PractitionerResource.java
```diff
@@ -70,20 +73,30 @@
...
     })
     public Response submitProvider(Practitioner provider) {
 
         final ProviderEntity entity = this.converter.fromFHIR(ProviderEntity.class, provider);
-        final List<ProviderEntity> existingProviders = this.dao.getProviders(null, entity.getProviderNPI(), entity.getOrganization().getId());
-        if (existingProviders.isEmpty()) {
+        final List<ProviderEntity> existingProvidersByOrganization = this.dao.getProviders(null, null, entity.getOrganization().getId());
```
nickrobison-usds: This list could get really large in the future, I wonder if it would make sense to add a new method to the DAO to return a count, rather than all the resources themselves?

dpc-attribution/src/main/java/gov/cms/dpc/attribution/resources/v1/PractitionerResource.java
```diff
@@ -70,20 +73,30 @@
...
+        final List<ProviderEntity> existingProvidersByNPI = existingProvidersByOrganization.stream()
+                .filter(providerEntity -> providerEntity.getProviderNPI().equals(entity.getProviderNPI()))
+                .collect(Collectors.toList());
+
+        if (providerLimit != null && existingProvidersByOrganization.size() >= providerLimit) {
```
nickrobison-usds: If we add the NonNull annotation to the config object, we can avoid the null check here.

dpc-attribution/src/main/java/gov/cms/dpc/attribution/DPCAttributionConfiguration.java
```diff
@@ -43,6 +43,8 @@
     @JsonProperty("swagger")
     private SwaggerBundleConfiguration swaggerBundleConfiguration;
 
+    private Integer providerLimit;
```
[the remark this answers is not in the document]
whuang85: Wouldn’t that make it required to always define it in the configuration file?  I thought in some environments we didn’t want to have a limit.

dpc-attribution/src/main/java/gov/cms/dpc/attribution/resources/v1/PractitionerResource.java
whuang85: Yea that makes sense

dpc-attribution/src/main/java/gov/cms/dpc/attribution/resources/v1/PractitionerResource.java
whuang85: Personally, I think a 429 should only be used if you are talking about rate limits, not sure if this would qualify as a rate limiting procedure. I’d be more inclined towards 422 if 304 is not acceptable.

dpc-attribution/src/main/java/gov/cms/dpc/attribution/DPCAttributionConfiguration.java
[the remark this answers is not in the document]
nickrobison-usds: Ah yes, that makes sense. Ignore me!

dpc-attribution/src/main/java/gov/cms/dpc/attribution/resources/v1/PractitionerResource.java
nickrobison-usds: Ok, I think 422 would work. Thoughts, @embh @DeirdreHolub

embh: Agree on using 422.

PR #653, opened 2020-02-26
Correct BFD server URL env var name in configuration files

embh:
This amends https://github.com/CMSgov/dpc-app/pull/640 to set `bbclient.baseServerUrl` in the aggregation dev, test, and prod-sbx configuration files to environment variable `BFD_URL`.


ronaldheft-usds:
Related: https://github.com/CMSgov/dpc-ops/pull/134 should be merged before this PR (that build will fail, but then will pass once this gets merged in)

PR #662, opened 2020-03-03
DPC-1133: Add support for creating consent resources

nickrobison-usds:
**Why**

This is required for #659, as we need to generate some test resources in order to verify opt-out works correctly. The existing design didn't support manually creating opt-out records, which is mostly fine, but we need at least some support for generating manual data for testing.

**What Changed**

Added an additional public method to the `ConsentEntityConverter` to convert from a FHIR Consent resource to one of our Entities. This isn't part of the larger converter framework, because we need more information at conversion time than what we currently support.

Added a new POST method to the ConsentResource for creating resources. Simple tests for the conversion process, but more will come with #659.

**Choices Made**

We don't have a full test suite yet, we have basic integration tests and the actual API will get tested when we verify the opt-out process.

**Tickets closed**:

DPC-1133: This ticket

**Future Work**

DPC-1136: Add remaining CRUD operations to Consent Service

**Checklist**


dpc-consent/src/main/java/gov/cms/dpc/consent/resources/ConsentResource.java
```diff
@@ -51,7 +52,7 @@
             "<p>Must provide ONE OF Consent ID as an _id or identifier, or a patient MBI or HICN to search for.", response = Bundle.class)
     @ApiResponses(@ApiResponse(code = 400, message = "Must provide Consent or Patient id"))
     public List<Consent> search(
-            @ApiParam(value = "Consent resource _id") @QueryParam(Consent.SP_RES_ID) Optional<UUID> id,
+            @ApiParam(value = "Consent resource _id") @QueryParam(IAnyResource.SP_RES_ID) Optional<UUID> id,
             @ApiParam(value = "Consent resource identifier") @QueryParam(Consent.SP_IDENTIFIER) Optional<UUID> identifier,
             @ApiParam(value = "Patient Identifier") @QueryParam(Consent.SP_PATIENT) Optional<String> patientId) {
```
RickHawesUSDS: Question: How does search get scoped to only an organization's consent?

nickrobison-usds:
@RickHawesUSDS Good question, the design of the consent service is still in flux but the core principle is to ingest opt-out information from ACO OS and link that to a bene MBI. We don't currently support opt-out on a per-organization basis, instead, it's a global opt-out file. So when we search for a bene opt-out file we don't need the organization ID, just the MBI.

The endpoint is currently only used for testing, but in the future, we'd like to expand it to support doing more fine-grained opt-out.

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/entities/ConsentEntityConverter.java
```diff
@@ -105,4 +111,74 @@ public static Consent convert(ConsentEntity consentEntity, String orgURL, String
...
+
+    public static void addPolicyCode(ConsentEntity entity, Consent resource) {
+        final String value = resource.getPolicyRule();
+        final String code;
+        if (OPT_IN_MAGIC.equals(value)) {
+            code = OPT_IN;
+        } else if (OPT_OUT_MAGIC.equals(value)) {
+            code = OPT_OUT;
+        } else {
+            throw new IllegalArgumentException(String.format("invalid value %s; should be %s or %s", value, OPT_IN, OPT_OUT));
```
DeirdreHolub: the value should either be OPT_IN_MAGIC or OPT_OUT_MAGIC, so the error message should use those constants.

dpc-common/src/main/java/gov/cms/dpc/fhir/converters/entities/ConsentEntityConverter.java
```diff
@@ -105,4 +111,74 @@ public static Consent convert(ConsentEntity consentEntity, String orgURL, String
...
+        if (matcher.find()) {
+            handlePatientMatch(entity, matcher);
+        } else {
+            throw new IllegalArgumentException("Cannot extract identifier");
+        }
+    }
+
+    private static void handlePatientMatch(ConsentEntity entity, Matcher matcher) {
+        final String idValue = matcher.group(2);
+        // If there's no identifier, then it's an MBI
```
embh: Wouldn't the identifier system be `http://hl7.org/fhir/sid/us-mbi` for an MBI?

PR #667, opened 2020-03-06
DPC-1119: Stuck jobs fixes: change exceptions in ResourceFetcher; remove RetryTransformer

embh:
**Why**
Jobs are getting stuck in the queue when they contain MBIs that are not in BFD.

**What Changed**
* Replace some `JobQueueFailure` exceptions with `ResourceNotFoundException`s so jobs won't be marked as failed in response to BFD 404s
* Remove `.compose(RetryTransformer.of(retry))` from `fetchResources()` `Flowable`

**Tickets closed**:
[DPC-1119](https://jiraent.cms.gov/browse/DPC-1119)

**Future Work**
We may wish to revisit our approach to retries.

**Checklist**


dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
```diff
@@ -137,7 +137,7 @@ private Resource fetchFirst(String mbi) {
                 beneId = getBeneIdFromPatient(patient);
                 return blueButtonClient.requestCoverageFromServer(beneId);
             default:
-                throw new JobQueueFailure(jobID, batchID, "Unexpected resource type: " + resourceType.toString());
+                throw new ResourceNotFoundException(String.format("Job %s, batch %s: Unexpected resource type: %s", jobID, batchID, resourceType.toString()));
```
ronaldheft-usds: I don't think we want to change this behavior. A `JobQueueFailure` still makes sense if we get an unknown resource type, as this would be an indication of a problem with our code, and not something the user could accidentally request.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
```diff
@@ -146,22 +146,22 @@ private Patient fetchPatient(String mbi) {
         try {
             patients = blueButtonClient.requestPatientFromServerByMbi(mbi);
         } catch (GeneralSecurityException e) {
-            throw new JobQueueFailure(jobID, batchID, "Failed to retrieve Patient");
+            throw new ResourceNotFoundException(String.format("Job %s, batch %s: Failed to retrieve Patient", jobID, batchID));
```
ronaldheft-usds: Can we wrap the `ResourceNotFoundException` with the caught `GeneralSecurityException` so we can still debug the original exception?

Additionally, the `ResourceNotFoundException` will end up in the `OperationalOutcome` files, so it might makes sense to log out the job failure information with a `logger.error` and then remove the job and batch information from the exception message, so it doesn't become user facing.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
```diff
@@ -146,22 +146,22 @@ private Patient fetchPatient(String mbi) {
...
         }
 
         if (patients.getTotal() == 1) {
             return (Patient) patients.getEntryFirstRep().getResource();
         }
 
-        throw new JobQueueFailure(jobID, batchID, String.format("Expected 1 Patient to match MBI but found %d", patients.getTotal()));
+        throw new ResourceNotFoundException(String.format("Job %s, batch %s: Expected 1 Patient to match MBI but found %d", jobID, batchID, patients.getTotal()));
```
ronaldheft-usds: Same as above. We should use `logger.error` to log out our IDs, and update the exception message to be more user friendly. Perhaps `Patient not found matching MBI %s`?

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
```diff
@@ -146,22 +146,22 @@ private Patient fetchPatient(String mbi) {
...
     }
 
     private String getBeneIdFromPatient(Patient patient) {
         return patient.getIdentifier().stream()
                 .filter(id -> DPCIdentifierSystem.BENE_ID.getSystem().equals(id.getSystem()))
                 .findFirst()
                 .map(Identifier::getValue)
-                .orElseThrow(() -> new JobQueueFailure(jobID, batchID, "No bene_id found in Patient resource"));
+                .orElseThrow(() -> new ResourceNotFoundException(String.format("Job %s, batch %s: No bene_id found in Patient resource", jobID, batchID)));
```
ronaldheft-usds: Same suggestions as above.

dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/engine/AggregationEngineTest.java
```diff
@@ -297,7 +297,7 @@ void badJobTest() {
 
         // Look at the result
         assertAll(() -> assertTrue(queue.getJobBatches(jobID).stream().findFirst().isPresent(), "Unable to retrieve job from queue."),
-                () -> assertEquals(JobStatus.FAILED, queue.getJobBatches(jobID).stream().findFirst().get().getStatus()));
+                () -> assertEquals(JobStatus.COMPLETED, queue.getJobBatches(jobID).stream().findFirst().get().getStatus()));
```
ronaldheft-usds: Assuming this is the test for the unknown resource type, the expected result should still be `FAILED`.

I think we need a new test (with some changes to the mocked BlueButton client) to support trigger the scenario found on line 156 of `ResourceFetcher`.

Additionally, we should update the postman / smoke tests to add an invalid ID to the roster, and verify we get a success on the job and an operation outcome file (in addition to the expected resources).

docker-compose.yml
```diff
@@ -22,9 +22,6 @@ services:
     environment:
       - ENV=local
       - JACOCO=${REPORT_COVERAGE}
-      - BFD_HASH_ITER=${BFD_HASH_ITER}
```
whuang85: are these not needed anymore to start the service?

embh: They come from the decrypted local.env. With them defined here, I was finding that locally the values were blank.

src/test/EndToEndRequestTest.postman_collection.json
```diff
@@ -1,6 +1,6 @@
 {
 	"info": {
-		"_postman_id": "8ac802cc-ada3-4dbe-bc0e-66efad8c06fc",
+		"_postman_id": "4b1e96d7-5669-45a9-a10d-9c5a94234719",
```
ronaldheft-usds: Nice work reproducing the issue in a postman test!

docker-compose.yml
ronaldheft-usds: I've seen the same thing, and I've had to remove them locally. Good with removing them.

ronaldheft-usds:
@embh How goes it with getting Jenkins to pass?

embh:
> @embh How goes it with getting Jenkins to pass?

More of the same for Travis.
> [ERROR] error reading /home/travis/.m2/repository/io/netty/netty-transport/4.1.44.Final/netty-transport-4.1.44.Final.jar; zip END header not found

PR #668, opened 2020-03-06
DPC-1074: Password Required to Delete Account

Sun-Mountain:
**Why**

To improve security, external users will now be required to enter their password to delete their account.

**What Changed**

I used [devise's](https://github.com/heartcombo/devise) `destroy_with_password` method to require users to input their password before they can delete their account.

**Choices Made**

I added a password field required to delete an account.

Screen Shot 2020-02-28 at 11 48 03 AM

Screen Shot 2020-03-13 at 11 13 39 AM

Screen Shot 2020-03-13 at 11 13 48 AM

Screen Shot 2020-03-13 at 11 13 59 AM

I also added a test for the a user deleting their account.

**Tickets closed**:

DPC-1074

**Future Work**

N/A

**Checklist**


dpc-web/app/controllers/users/registrations_controller.rb
```diff
@@ -4,6 +4,20 @@ module Users
...
     include MultiModelLoginHelper
     skip_before_action :check_user, except: %i[new create]
+
+    def destroy
+      @user = User.find(current_user.id)
+      if @user.destroy_with_password(user_params[:password_to_delete])
+        redirect_to root_url, notice: 'User account deleted.'
+      else
+        redirect_to edit_user_registration_url
+        flash[:notice] = 'Could not delete account. Please enter the correct password.'
```
switzersc-usds: We use `flash[:alert]` for error messages (`:notice` being for success messages). Also, this needs to either be implemented in the same way as line 11 above, or needs to come before the redirect; otherwise I don't think it'll take effect (the redirect will be called and the block exited). I'd probably vote the former for consistency within this method.

dpc-web/app/views/devise/registrations/edit.html.erb
```diff
@@ -132,7 +132,11 @@
   <div class="ds-l-col--12 ds-l-md-col--6 ds-u-margin-bottom--1 ds-u-margin-x--auto">
     <div class="card card--border-top--red ">
       <h2 class="ds-u-color--error-dark ds-u-margin-top--0">Delete my account</h2>
-      <p><%= button_to "Delete my account", registration_path(resource_name), data: { confirm: "Are you sure you want to delete your account?" }, method: :delete, :class => "ds-c-button ds-c-button--danger" %></p>
+      <%= form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :delete }) do |f| %>
+        <%= f.label :password, :class => "ds-c-label" %> <i>Enter your password to delete your account.</i><br />
```
switzersc-usds: Let's use the modern hash with symbol key syntax rather than hash rockets -- so `class: ` instead of `:class => `

dpc-web/app/views/devise/registrations/edit.html.erb
```diff
@@ -132,7 +132,11 @@
...
+        <%= f.password_field :password_to_delete, autocomplete: "off", :class => "ds-c-field" %><br />
```
switzersc-usds: Ditto

dpc-web/spec/features/authentication/user_deletes_account_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  context 'when successful' do
+    scenario 'user inputs correct password' do
+      user_sign_in
+      fill_in 'user_password_to_delete', with: '123456'
+
+      find('[data-test="delete-user-account"]').click
+
+      expect(User.last).not_to be_present
```
switzersc-usds: Because this is a feature spec, let's try to assert on the behavior we expect to see as a user in this scenario rather than on the internals of the application. What page are we on? Do we see a message?

dpc-web/spec/features/authentication/user_deletes_account_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  context 'when unsuccessful' do
+    scenario 'user inputs incorrect password' do
+      user_sign_in
+      fill_in 'user_password_to_delete', with: '3v3ryDayPotato'
+
+      find('[data-test="delete-user-account"]').click
+
+      expect(User.last).to be_present
```
switzersc-usds: Ditto

dpc-web/spec/features/authentication/user_deletes_account_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    scenario 'user inputs no password' do
+      user_sign_in
+      fill_in 'user_password_to_delete', with: nil
+
+      find('[data-test="delete-user-account"]').click
+
+      expect(User.last).to be_present
```
switzersc-usds: Ditto

dpc-web/app/controllers/users/registrations_controller.rb
```diff
@@ -4,6 +4,20 @@ module Users
   class RegistrationsController < Devise::RegistrationsController
     include MultiModelLoginHelper
     skip_before_action :check_user, except: %i[new create]
+
+    def destroy
+      @user = User.find(current_user.id)
+      if @user.destroy_with_password(user_params[:password_to_delete])
```
switzersc-usds: Because we're in a devise controller, let's use the semantics and behavior the Devise uses to implement the behavior we want. Devise's standard [registrations#destroy](https://github.com/heartcombo/devise/blob/master/app/controllers/devise/registrations_controller.rb#L65) behavior looks like this:

    def destroy
      resource.destroy
      Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
      set_flash_message! :notice, :destroyed
      yield resource if block_given?
      respond_with_navigational(resource){ redirect_to after_sign_out_path_for(resource_name) }
    end

If we don't follow the logic and semantics used here, we risk missing something important about Devise's default implementation or creating a gotcha that trips us up later.  So let's try something like:

    def destroy
      if resource.destroy_with_password(user_params[:password_to_delete])
        Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
        set_flash_message! :notice, :destroyed
        yield resource if block_given?
        respond_with_navigational(resource){ redirect_to after_sign_out_path_for(resource_name) }
      else
        respond_with resource # this might take care of the redirect and flash message for us - should test
      end
    end

dpc-web/app/controllers/users/registrations_controller.rb
switzersc-usds: (this  comment  may not be relevant if `respond_with resource` takes care of this in my proposed refactoring)

dpc-web/spec/features/authentication/user_deletes_account_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
+
+RSpec.feature 'user deletes account' do
+  let(:user) { create :user }
+
+  def user_sign_in
```
switzersc-usds: I believe we already have a sign in helper available, so you don't need to define another one here. Instead of this, how about a `before` block that utilizes the existing method:

    before(:each) do
      sign_in user, scope: :user
      visit edit_user_registration_path
    end

dpc-web/app/controllers/users/registrations_controller.rb
```diff
@@ -4,6 +4,26 @@ module Users
   class RegistrationsController < Devise::RegistrationsController
     include MultiModelLoginHelper
     skip_before_action :check_user, except: %i[new create]
+
+    def destroy
+      @user = User.find(current_user.id)
+      if resource.destroy_with_password(user_params[:password_to_delete])
+        Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
+        flash[:notice] = 'Bye! Your account has been successfully cancelled. We hope to see you again soon.'
```
switzersc-usds: Haha. Just curious where this text came from? Not sure if it's in the DPC voice but I could be wrong!

dpc-web/app/controllers/users/registrations_controller.rb
```diff
@@ -4,6 +4,26 @@ module Users
   class RegistrationsController < Devise::RegistrationsController
     include MultiModelLoginHelper
     skip_before_action :check_user, except: %i[new create]
+
+    def destroy
+      @user = User.find(current_user.id)
```
switzersc-usds: No longer needed

dpc-web/app/controllers/users/registrations_controller.rb
```diff
@@ -4,6 +4,26 @@ module Users
...
+      if resource.destroy_with_password(user_params[:password_to_delete])
+        Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
+        flash[:notice] = 'Bye! Your account has been successfully cancelled. We hope to see you again soon.'
+        yield resource if block_given?
+        respond_with_navigational(resource) { redirect_to after_sign_out_path_for(resource_name) }
+      else
+        flash[:alert] = 'Account could not be deleted.'
```
switzersc-usds: Can we give a more informative error message? Also, did `respond_with resource` not work here as an alternative to lines 16-7 (setting the flash and doing the redirect explicitly)? Just curious -- if it does work then let's use that, but if it doesn't it'd be useful to know.

dpc-web/spec/features/authentication/user_deletes_account_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+  end
+
+  context 'when unsuccessful' do
+    scenario 'user inputs incorrect password' do
+      fill_in 'user_password_to_delete', with: '3v3ryDayPotato'
+
+      find('[data-test="delete-user-account"]').click
+
+      expect(page.body).to include('Your email or password is incorrect')
```
switzersc-usds: This text isn't what's being set in the controller (`Account could not be deleted.`). Why is there a difference?

dpc-web/app/controllers/users/registrations_controller.rb
Sun-Mountain: `respond_with resource` did not work, it threw this error:

Screen Shot 2020-03-09 at 4 40 30 PM

dpc-web/app/views/devise/registrations/edit.html.erb
```diff
@@ -7,7 +7,7 @@
       <h1>Edit your info</h1>
 
       <%= form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put }) do |f| %>
-        <%= render "devise/shared/error_messages", resource: resource %>
+        <%= devise_error_messages! %>
```
switzersc-usds: Can you explain this change?

dpc-web/app/controllers/users/registrations_controller.rb
Sun-Mountain: @switzersc-usds that's what the current message in the `devise.en.yml` file is and what we've been saying to users when they sign out. Not sure how crucial this language is. ¯\_(ツ)_/¯

dpc-web/app/views/devise/registrations/edit.html.erb
Sun-Mountain: `<%= render "devise/shared/error_messages", resource: resource %>` wasn't displaying any errors, replaced with the functional `<%= devise_error_messages! %>`

switzersc-usds: Interesting. I'm concerned about this change because the shard error messages show the errors for the registration form (e.g. if you leave out a required field) so I think they're pretty important for us to keep. Also, `devise_error_messages!` is [being deprecated](https://github.com/heartcombo/devise/blob/e01fb8bb22109cbeb00d325db8b161a67f380bb7/app/helpers/devise_helper.rb).

Instead of replacing the error messages partial, let's figure out why the message we wanted wasn't showing up and see if we can figure out another way to resolve that.

dpc-web/spec/features/authentication/user_deletes_account_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      expect(page.body).to include('Current password is invalid')
+    end
+
+    scenario 'user inputs no password' do
+      fill_in 'user_password_to_delete', with: ''
+
+      find('[data-test="delete-user-account"]').click
+
+      expect(page.body).to include('1 error prohibited this user from being saved')
```
Sun-Mountain: For some reason, the test wouldn't acknowledge `"Current password is blank"` as being present, even though that is what's being shown in addition to `1 error prohibited this user from being saved`.

PR #676, opened 2020-03-10
aggregation engine health check

whuang85:
**Why**

A Health check on the engine itself is important to give insight into knowing if the worker is actually distressed or in some other state. In combination with the queue health check, we can know if the engine is just stuck running.

**What Changed**

Added a health check based off the isRunning status. If isRunning is false, then health check will return unhealthy and if true, it'll return healthy.  IsRunning is false only when some unknown error breaks the loop

**Choices Made**

Justify changes made and any reasons for why a given path was taken, as opposed to an alternative option.

**Tickets closed**:

**Future Work**

**Checklist**


dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
```diff
@@ -179,6 +186,7 @@ protected void processJobBatch(JobQueueBatch job) {
         } catch (Exception error) {
             logger.error("FAILED job {} batch {}", job.getJobID(), job.getBatchID(), error);
             this.queue.failBatch(job, aggregatorID);
+            inError.set(true);
```
ronaldheft-usds: Marking a batch as failed is a recoverable state. How would the aggregator return to a healthy state after marking a batch as inError?

ronaldheft-usds: I'm thinking we either need to set inError on line 188, and then set it back to false in a finally block (so if the batch is marked correctly as failed and processing resumes, the aggregator is no longer inError), or we don't count this scenario as an error scenario.

dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/health/AggregationEngineHealthCheckTest.java
```diff
@@ -0,0 +1,159 @@
...
+                Collections.singletonList("1"),
+                Collections.singletonList(ResourceType.Patient)
+        );
+
+        AggregationEngineHealthCheck healthCheck = new AggregationEngineHealthCheck(aggregatorID, engine);
+        Assert.assertTrue(healthCheck.check().isHealthy());
+
+        ExecutorService executor = Executors.newCachedThreadPool();
+        executor.execute(engine);
+        executor.awaitTermination(5, TimeUnit.SECONDS);
```
ronaldheft-usds: Is this test starting the aggregator for 5 seconds, then running assertions, and stopping the aggregator?

Do we need to test the behind-the-scenes action of the AggregationEngine, or is a mock more sufficient (simulating isRunning and inError states)?

We do have an existing test suite called `AggregationEngineTest` that may be better suited to handle the verifying the `inError` state is set correct.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
whuang85: @ronaldheft-usds, if it went to inError and it's running wouldn't EC2 restart the instance and spin up a new AggregationEngine?

ronaldheft-usds: Correct, `this.queue.failBatch` is a recoverable scenario. It marks the batch as failed and then starts processing the next item. Unlike what happens in the `pollQueue` method, where an error in their stops processing, and the AggregationEngine ends up just sitting there doing nothing.

whuang85: just to put it here, so the only place where a true unrecoverable error can happen is in the error callback of the subscribe but since the processBatch catches all exceptions, it should technically never reach that callback

dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/engine/AggregationEngineTest.java
```diff
@@ -157,6 +158,26 @@ void simpleJobTest() {
...
+                orgID,
+                TEST_PROVIDER_ID,
+                Collections.singletonList(MockBlueButtonClient.TEST_PATIENT_MBIS.get(0)),
+                Collections.singletonList(ResourceType.Patient)
+        );
+
+        // Work the batch
+        queue.claimBatch(engine.getAggregatorID())
+                .ifPresent(t -> engine.onError(new RuntimeException("error")));
+
```
RickHawesUSDS: Should anything outside of the engine ever call the onError? In other words, why is this method public? I don't quite follow the logic here. It seems to be saying if engine.onError is called then engine.inError is true. Not sure if that is what you wanted to test.

dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/health/AggregationEngineHealthCheckTest.java
```diff
@@ -0,0 +1,155 @@
...
+        AggregationEngineHealthCheck healthCheck = new AggregationEngineHealthCheck(engine);
+        Assert.assertTrue(healthCheck.check().isHealthy());
+
+        ExecutorService executor = Executors.newCachedThreadPool();
+        executor.execute(engine);
+        executor.awaitTermination(5, TimeUnit.SECONDS);
+
+        Assert.assertTrue(healthCheck.check().isHealthy());
+    }
+}
```
RickHawesUSDS: Unless I'm missing something, I'm don't see a negative (eg. Assert.assertFalse(healthCheck.check().isHealth())) test. Should there be one?

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
```diff
@@ -123,17 +129,23 @@ protected void pollQueue() {
...
+                        this::onError,
+                        this::onCompleted
                 );
     }
 
+    public void onError(Throwable error) {
+        logger.error("Error processing queue. Exiting...", error);
+        queueRunning.set(false);
+        inError.set(true);
+    }
```
RickHawesUSDS: I'm wondering how a stuck job will make the aggregator unhealthy? Or is the intent to only make the aggregator unhealthy when an exception is thrown?

dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/engine/AggregationEngineTest.java
whuang85: After the changes from Ron and Emily, theoretically the `engine.onError` should never be called as the `processBatch` already catches all Exceptions and then marks the batch failed and the engine moves on. My health check is getting less relevant as Ron checks in more commits.

whuang85:
@ronaldheft-usds @RickHawesUSDS The more I poke and prod the engine to legitimately throw a non recoverable error the more I feel like this health check is irrelevant. Both the `claimBatch` and `processBatch` are now recoverable thanks to Ron's and Emily's future changes and having difficulty in crafting a test to throw the engine into an unhealthy state is leading me to believe that this check wouldn't actually do anything.. What are your thoughts?

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
ronaldheft-usds: A stuck job will be handled by the `JobQueueHealthCheck`. This is to handle the scenario we were seeing where the aggregator is exiting the loop. That has been fixed with https://github.com/CMSgov/dpc-app/pull/664, but I still think it's a great idea to have a health check that catches this issue if it gets reintroduced somehow.

dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/health/AggregationEngineHealthCheckTest.java
whuang85: I’ve left these tests to showcase that these scenarios should not cause the engine to shutdown but to continue running.

I’ve added a test in aggregationEngineTest that does have the engine exit the loop with an error and thus calls onError and sets the isRunning to false.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
```diff
@@ -54,7 +54,7 @@
     private final Meter resourceMeter;
     private final Meter operationalOutcomeMeter;
     private Disposable subscribe;
-    protected AtomicBoolean queueRunning = new AtomicBoolean(false);
+    protected AtomicBoolean queueRunning = new AtomicBoolean(true);
```
RickHawesUSDS: Suggest that you add a comment for the reason to initialize to running. There was a discussion on this and it would be helpful to capture.

dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/health/AggregationEngineHealthCheckTest.java
```diff
@@ -0,0 +1,160 @@
...
+                TEST_PROVIDER_ID,
+                Collections.singletonList("1"),
+                Collections.singletonList(ResourceType.Patient)
+        );
+
+        AggregationEngineHealthCheck healthCheck = new AggregationEngineHealthCheck(engine);
+        Assert.assertTrue(healthCheck.check().isHealthy());
+
+        ExecutorService executor = Executors.newCachedThreadPool();
+        executor.execute(engine);
```
RickHawesUSDS: Isn't it important to assert while the engine is running? Suggest that you check `isHealthy()` here.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
ronaldheft-usds: Great change! Covers us in that window during startup.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
```diff
@@ -123,17 +123,21 @@ protected void pollQueue() {
...
-                        error -> {
-                            logger.error("Error processing queue. Exiting...", error);
-                            queueRunning.set(false);
-                        },
-                        () -> {
-                            logger.info("Finished processing queue. Exiting...");
-                            queueRunning.set(false);
-                        }
+                        this::onError,
+                        this::onCompleted
```
ronaldheft-usds: Nice breaking out the methods into their own callbacks. Makes the code easier to read and easier to test.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/health/AggregationEngineHealthCheck.java
```diff
@@ -0,0 +1,27 @@
...
+    @Inject
+    public AggregationEngineHealthCheck(AggregationEngine aggregationEngine) {
+        this.aggregationEngine = aggregationEngine;
+    }
+
+    @Override
+    public Result check() {
+        Result result = Result.healthy();
+        if (!aggregationEngine.isRunning()) {
+            result = Result.unhealthy("Aggregation Engine instance: " + aggregationEngine.getAggregatorID() + " in error state");
```
ronaldheft-usds: Switching up `aggregationID` to public was a good move. Ensures we are consistent with the IDs.

dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/engine/AggregationEngineTest.java
```diff
@@ -388,6 +392,31 @@ void testBlueButtonException() throws GeneralSecurityException {
 
     }
 
+    @Test
+    public void testUnhealthyIfProcessJobBatchThrowsException() throws InterruptedException {
```
ronaldheft-usds: Great work covering this scenario. This scenario is exactly why we want this health check.

dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/health/AggregationEngineHealthCheckTest.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * These tests are here to make sure the engine is still running/polling in situations where errors are recoverable.
+ * A test to check if the engine exits out of the loop correctly when an error occurs
+ * in AggregationEngineTest#testUnhealthyIfProcessJobBatchThrowsException
+ */
+@ExtendWith(BufferedLoggerHandler.class)
+public class AggregationEngineHealthCheckTest {
```
ronaldheft-usds: I really like this test quite. You covered all the issues that have tripped up on the incident.

My only concern is the 5 second requirement per test. This adds 20+ seconds to our test suite. I don't want to hold up closing out the incident over this, but I think we should keep this tech debt in mind, that we may have to optimize our unit tests in the future to run quicker.

ronaldheft-usds:
@whuang85 Updated the branch from master. Once the tests run again, I'll merge and deploy.

dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/health/AggregationEngineHealthCheckTest.java
ronaldheft-usds: @RickHawesUSDS I think @whuang85 has that covered here: https://github.com/CMSgov/dpc-app/pull/676/files#diff-b6b2fc6f1b1f240ec7b0a995d56cf410R84

dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/health/AggregationEngineHealthCheckTest.java
whuang85: I can cut it down on some time, I just wanted to ensure it has time to process a job and the rest of the time it is sitting in the `repeatWhen` loop

PR #679, opened 2020-03-10
Additional error handling around failing batches

ronaldheft-usds:
**Why**

It's possible marking batches as failed is not succeeding, and we're swallowing the exception somewhere.

**What Changed**

* Adds a unit test to validate the queue doesn't stop processing on an exception in `processJobBatch`
* Adds logging a test around possible exceptions in `queue.failBatch`

**Choices Made**

**Tickets closed**:

**Future Work**

**Checklist**


ronaldheft-usds:
I don't think we need to consider this part of the incident, but @embh @whuang85 @DeirdreHolub  I would still appreciate a review. This gives us better logging if we ever run into a problem marking a batch as failed (like for example, if the database goes down, etc).

PR #680, opened 2020-03-10
DPC-1158: Cannot delete Thomas Jefferson University or Webster organization records in the admin portal

Sun-Mountain:
**Why**

If an organization is deleted directly in the API, the app returns an error when an internal user tries to disable or delete an organization.

**What Changed**

When the API returns the error `'Cannot find organization'`, the app will still disable the organization from the API. The user will also be able to delete the enabled organization.

**Choices Made**

N/A

**Tickets closed**:

DPC-1158

**Future Work**

N/A

**Checklist**


dpc-web/spec/models/registered_organization_spec.rb
```diff
@@ -184,18 +184,29 @@
       end
 
       context 'failed API request' do
-        it 'adds to errors and does not destroy object' do
+        before(:each) do
           stub_api_client(message: :create_organization, success: true, response: default_org_creation_response)
-          reg_org = create(:registered_organization)
+          @reg_org = create(:registered_organization)
```
switzersc-usds: In specs, it's conventional to assign variables used across examples with `let` outside of the before block. Instead of creating an instance variable here, before line 187 add this `let` block:

    let(:registered_organization) { create(:registered_organization } 

And then you'll be able to use `registered_organization` in the examples instead of `@reg_org`

dpc-web/app/models/registered_organization.rb
```diff
@@ -76,8 +76,11 @@ def update_api_endpoint
   def delete_api_organization
     api_request = APIClient.new(api_env).delete_organization(self)
     api_response = api_request.response_body
+
     return if api_request.response_successful?
 
+    return if api_response.include? 'Cannot find organization'
```
switzersc-usds: What do you think about logging that the org wasn't found in the API? I'm inclined to suggest it in case we have to trace a transaction later; logging that there was an unexpected consistency in the transaction chain might be helpful.

dpc-web/spec/models/registered_organization_spec.rb
Sun-Mountain: When I tried using `let(:registered_organization) { create(:registered_organization) }` the test responded with the following error for both tests:

```
Failure/Error:
       api_request = APIClient.new(api_env).create_organization(
         organization,
         fhir_endpoint: fhir_endpoint.attributes.slice('name', 'status', 'uri')
       )
     
       #<InstanceDouble(APIClient) (anonymous)> received unexpected message :create_organization with (#<Organization id: 1, name: "The Health Factory 1", organization_type: "primary_care_clinic", num_pro...-03-17 13:18:11", updated_at: "2020-03-17 13:18:11", npi: "test-npi-1", vendor: nil, vendor_id: nil>, {:fhir_endpoint=>{"name"=>"DPC Sandbox Test Endpoint", "status"=>"test", "uri"=>"https://dpc.cms.gov/test-endpoint"}})
```

dpc-web/app/models/registered_organization.rb
Sun-Mountain: I agree. How would you log this in the app?

switzersc-usds:
```
if  api_response.include? 'Cannot find organization'
  Rails.logger.warn "Cannot delete API organization with id #{api_id}: Organization not found."
  return
end
```

dpc-web/spec/models/registered_organization_spec.rb
switzersc-usds: Gotcha. Looks like that would be caused by `let` getting called before the API client gets stubbed (line 188). I don't think this is a blocker but I would prefer using `let` or local variables rather than instance variables, partly as a project convention and partly because ivars can end up being cruft that adds either latency or confusion later, IME.

PR #725, opened 2020-04-02
DPC-1210: Generate fake identifier for provider orgs in sandbox

Sun-Mountain:
**Why**

This solves the current workflow issue where internal admins cannot enable sandbox for a provider org without an NPI (and users do not know they are supposed to enter an NPI after they've been assigned). Furthermore, we know people are entering in fake NPIs in sandbox anyway because they may not know theirs yet.

**What Changed**

When a provider organization is created or updated, if no `provider_id` is present one will be generated.

If an NPI is not present, the `provider_id` will be used as the external identifier in the API.

Adjusted existing tests and added tests for npi/provider_id function.

**NOTE:** I'm not sure how to make this only available to [sandbox.dpc.cms.gov](https://sandbox.dpc.cms.gov/) and not production. Is this something I do on the dev end? Or is there something I need to add to let DevOps set those parameters.

**Choices Made**

I called the provider identifier `provider_id` to mirror the `vendor_id`.

**Tickets closed**:

[DPC-1210](https://jiraent.cms.gov/browse/DPC-1210)

**Future Work**

We need another ticket for differentiating sandbox and production parameters if necessary.

**Checklist**


dpc-web/app/models/organization.rb
```diff
@@ -38,21 +38,23 @@ def npi=(input)
   end
 
   def api_credentialable?
-    if health_it_vendor?
-      registered_organizations.count.positive?
-    else
-      registered_organizations.count.positive? && npi.present?
-    end
+    registered_organizations.count.positive?
```
dhgreene, line 41: Do we actually no longer care whether the NPI is present for providers?

PR #739, opened 2020-04-10
DPC-1274-create nohtml validator

whuang85:
**Why**

We don't expect scripts and/or html in any of our inputs so need to reject request when those are found in our inputs

**What Changed**

Added a NoHtml Validator with the annotation and annotated all the String fields in our entities and also the String fields in API resources.

**Choices Made**

Used a validator instead of a filter because a filter would have to be generic enough to find all string fields in the request which is a hard problem and could introduce a lot of edge cases.

**Tickets closed**:

https://jiraent.cms.gov/browse/DPC-1274

**Future Work**

**Checklist**


dpc-api/src/main/java/gov/cms/dpc/api/resources/AbstractGroupResource.java
```diff
@@ -26,7 +27,7 @@ protected AbstractGroupResource() {
     public abstract Response createRoster(OrganizationPrincipal organizationPrincipal, @Valid @Profiled(profile = AttestationProfile.PROFILE_URI) Provenance rosterAttestation, Group attributionRoster);
```
embh, line 27: Should we also be preventing HTML in the FHIR resources like Provenance and Group?

whuang85: @embh all the FHIR resources are not under our control so I can't add the annotation to them. That is why I had to validate our hibernate entity classes instead which should be fine since I believe this method hits the corresponding attribution method which just converts these FHIR resources into hibernate entity classes and saves to the database.

dpc-common/src/test/java/gov/cms/dpc/common/hibernate/validator/NoHtmlValidatorTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @MethodSource("stringSource")
+    public void noHtmlValidatorTest(String value, boolean isEmpty) {
+
+        TestObject testObject = new TestObject();
+        testObject.setA(value);
+        Set<ConstraintViolation<TestObject>> violations = validator.validate(testObject);
+        Assertions.assertEquals(isEmpty, violations.isEmpty());
+    }
+
+    private static Stream<Arguments> stringSource() {
```
dhgreene, line 37: There are a number of tricks attackers can use to fool HTML sanitizers.  Can you include a few from [this](https://owasp.org/www-community/xss-filter-evasion-cheatsheet) or another source?

PR #753, opened 2020-04-16
DPC-25 Lookback in Aggregation Engine

whuang85:
**Why**

Alternative solution where look back is in aggregation engine. Instead of preventing a patient from being added to the roster, we instead check the request if the patient is associated with the provider and org through a roster. If they are associated, we then make a call to fetch that patient's EoB to determine if they have a claim within 18 months (which determines the eligibility of the provider fetching and saving a patient's data).

**What Changed**

Aggregation engine changes to grab the EoB of the patient to be able to determine look back before processing the job. The consequence of this change is regardless of what resource types are exported, an additional call to BFD for EoB will be made for the sole purpose of look back determination. 

**Choices Made**

Refactored the observables for reusability in the look back determination. Look back only needs the EoB resource and not actually saving that data, so refactoring the methods to be broken up into more atomic methods allow look back to only grab the data without also saving the data.

**Tickets closed**:

https://jiraent.cms.gov/browse/DPC-25

**Future Work**

**Checklist**


dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/dao/RosterDAO.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.UUID;
+
+public class RosterDAO extends AbstractDAO<RosterEntity> {
+
+    @Inject
+    public RosterDAO(DPCManagedSessionFactory factory) {
+        super(factory.getSessionFactory());
+    }
+
+    public boolean withinRoster(UUID organizationID, String providerID, String patientReference) {
```
embh, line 22: Would it be more flexible to pass in the Patient and Practitioner UUIDs so they don't have to be parsed out of the reference strings?

PR #786, opened 2020-05-06
DPC-443 limit the number of members in roster

whuang85:
**Why**

We need a limit on the number of patients that a provider can have attributed to them. This is a really basic anti-fraud measure.

We'll start at 1000 for now, but that may have to be changed. It should definitely be a config option.

**What Changed**

Check roster size when creating, updating, and adding members

**Choices Made**

Check at the attribution layer for roster size.

**Tickets closed**:

https://jiraent.cms.gov/browse/DPC-443

**Future Work**

**Checklist**


dpc-attribution/src/main/resources/application.conf
```diff
@@ -48,6 +48,8 @@ dpc.attribution {
   // Disable FHIR validation
   fhir.validation.enabled = false
 
+  patientLimit = 1000
```
dhgreene, line 51: This is the kind of thing we want to be setting via an environment variable, especially since we expect it to change over time.  Adding a new one will require an associated `dpc-ops` PR request and an addition to our store of secrets.  😄

whuang85: with the 3rd party tool we use, the philosophy of the tool is to make application.conf store all the defaults and anything environment specific would be stored in prod.application.yml and etc..  So I added ${PATIENT_LIMIT} to prod.application.yml and all other environments will just use the default

dhgreene: That's great!  Let's set all the other environments (especially `prod-sbx`) to have a default of no limit, then.

whuang85: If we don’t want a limit for the lower environments then we can just take it out of application.conf as I coded this in a way where if it’s null, then there is no limit.

whuang85: If we don’t want a limit for the lower environments then we can just take it out of application.conf as I coded this in a way where if it’s null, then there is no limit.

PR #808, opened 2020-05-19
Dpc 133 use redis with async job

SMLuthi:
**Why**

Ultimately, we want our job processing to take advantage of an in memory database like Redis. Delayed job uses a database table to manage job data which is not ideal.

**What Changed**

I opted to rip out all the delayed job scaffolding and remove the gem. It was replaced by sidekiq which takes advantage of redis. 

**Choices Made**

Since nothing currently is taking advantage of the current async job processing tool we had, it was the easiest time to opt to replace it with something that falls more in line with what we are looking for. Sidekiq has solid tooling out of the box and uses Redis as its job storage db. 

**Tickets closed**:

[DPC-133](https://jira.cms.gov/browse/DPC-133)

**Future Work**

Currently, sidekiq is executed via the `-d` daemonize flag. We should move away from this as it will not bring it back up if it fails for some reason.

**Checklist**


SMLuthi:
I have confirmed that deploying this to dev and, given that `REDIS_URL` is set via the corresponding dev-ops branch, the sidekiq worker will start as a background process and be ready to consume jobs from redis. I've dropped into an ec2 instance, created an example job and verified that it was consumed and completed successfully.

PR #814, opened 2020-05-22
DPC-335: Verify public key with snippet signature

embh:
**Why**

Validation needs to be improved on public keys submitted by users.

**What Changed**

Creating a public key in the API now requires a signature. The signature is generated with the public key's corresponding private key, by signing a predefined text snippet. This applies to both the API `/Key` endpoint `POST` operation and the CLI `key upload` command. The endpoint now expects a JSON request body containing `key` and `signature`. The CLI command now must include a path to a file containing the signature.

**Tickets closed**:

[DPC-335](https://jira.cms.gov/browse/DPC-335)
[DPC-336](https://jira.cms.gov/browse/DPC-336)

**Future Work**

There is web app work in progress to go with these changes.

**Checklist**

- [x] All tests are passing via `make ci-app` (app change) and `make ci-web` (website change)
- [x] Swagger documentation has been updated
- [x] FHIR documentation has been updated
- [x] Any required dpc-ops changes have a PR submitted and mentioned in this ticket
- [x] Any manual migration steps are documented, scripts written (where applicable), and tested
- [x] Before merging, any required dpc-ops changes have been approved and merged into master of the dpc-ops repo

PR #819, opened 2020-05-27
DPC-333: Update the website to request snippets of encrypted data from users to validate public keys

Sun-Mountain:
**REVIEW AFTER [DPC-335](https://github.com/CMSgov/dpc-app/pull/814) HAS BEEN MERGED**

**Why**

**AC**: On the public keys screen, create a field for the user to submit a snippet of data encrypted by their public key. Work with the API engineer also assigned to this story to define what that snippet should include and work with the H team to add instructions to the webpage

Using Crystal's design, the page now walks users through the creation of private and public keys and the signature snippet.

**What Changed**

Paired with [DPC-335](https://github.com/CMSgov/dpc-app/pull/814), external users can enter the required signature to verify their public key.

**Choices Made**

Screen Shot 2020-06-09 at 9 16 35 AM

Screen Shot 2020-06-09 at 9 16 42 AM

**Tickets closed**:

[DPC-333](https://jira.cms.gov/browse/DPC-333) - Update the website to request snippets of encrypted data from users to validate public keys
[DPC-337](https://jira.cms.gov/browse/DPC-337) - Parse public key errors

**Future Work**

To successfully create a signature for the public key, the user will need a text file with `This is the snippet used to verify a key pair in DPC.` inside of it. <-- something to consider when talking thru the steps to create a key/signature pair. @embh suggested that we provide them with the text file. Not sure when we would do that.

**Checklist**

- [/] Swagger documentation has been updated
- [/] FHIR documentation has been updated
- [/] Any required dpc-ops changes have a PR submitted and mentioned in this ticket
- [/] Any manual migration steps are documented, scripts written (where applicable), and tested
- [/] Before merging, any required dpc-ops changes have been approved and merged into master of the dpc-ops repo


embh:
Context around my suggestion of providing a text file: the alternative would presumably be to display a string that they need to copy and paste into a file to be signed. But many text editors will automatically add a newline when a file is saved. As it a result, it would no longer be the exact string we expect, so signatures would not be correct.

dpc-web/app/controllers/public_keys_controller.rb
```diff
@@ -24,6 +29,10 @@ def create
     end
   end
 
+  def download_snippet
+    send_file 'public/snippet.txt', type: 'application/zip', status: 202
```
dhgreene, line 33: Are we using `application/zip` to avoid encoding problems?  Have we tested what that does in various browsers/OS's?

Sun-Mountain: `application/zip` is what worked in the code? but it downloads the `.txt` file.

I have not tested on different browsers but will do so immediately and report back.

Sun-Mountain: Can confirm downloads work in Chrome, Safari, Firefox, and Brave browsers.

PR #822, opened 2020-05-28
DPC-162 add auth related logging filter

whuang85:
**Why**

Since we use @FormParam for our posts that means the user can post with query strings.  One of our endpoints that takes a JWT token to get the access token allows this and to prevent this and any other endpoints from logging the query string, this filter was devised.

**What Changed**

Add a logback filter

**Choices Made**

Could not do filtering at the splunk level, so filtering at the programatic logging level here.  Since there is not a way to determine if a value is a JWT token, we can only look for the key (client_assertions).

**Tickets closed**:

https://jira.cms.gov/browse/DPC-162

**Future Work**

**Checklist**


dpc-common/src/test/java/gov/cms/dpc/common/logging/filters/AuthRelatedLoggingFilterTest.java
```diff
@@ -0,0 +1,43 @@
...
+    @MethodSource("stringSource")
+    public void testQueryString(String queryString, FilterReply expectedReply) {
+        Mockito.when(event.getQueryString()).thenReturn(queryString);
+        FilterReply reply = filter.decide(event);
+        Assertions.assertEquals(expectedReply, reply);
+    }
+
+    private static Stream<Arguments> stringSource() {
+        return Stream.of(
+                Arguments.of("client_assertions=secret_token", FilterReply.DENY),
```
embh, line 38: Could we make this `client_assertion` (singular) since that's likely the name for the parameter? Or maybe include both to show that `client_assertion` anywhere is rejected.

whuang85: ah good catch, I even copied it from the token resource so I have no idea how an `s` got appended lol

PR #859, opened 2020-06-15
Dpc 150 validate and properly escape all web model inputs

SMLuthi:
**Why**

In the penetration tests they were able to inject html into a csv file that they downloaded and then were able to fire it. This shouldn't be possible.

**What Changed**

When downloading the list of users in a csv file certain data (that does not have strong validation) is now html_escaped to prevent any malicious html being exposed. 

**Choices Made**

This is the only file I noticed that you can download via the web app. Originally, I was going to escape ALL the html displayed in any of the views that did not have strong validation. However, Rails already automatically does it (since Rails 3). Therefore it is not necessary. In fact, its good to remember that we have to use `html_safe` if we ever need to actually properly display html in a view. I checked over the validations that are already in place. From what I could find, they are about as stringent as they can be... things like city, street addresses, names, etc cannot be validated beyond presence (ie via a regex). 

**Tickets closed**:

[DPC-150](https://jira.cms.gov/browse/DPC-150)

**Checklist**


Sun-Mountain:
There is now a second file downloaded from the app (snippet.txt) on the public key page.

This LGTM but I'd want @dhgreene to sign off on this just to make sure.

PR #879, opened 2020-06-24
DPC-442 Feature: null check in hapi exception handler

whuang85:
the operation outcome is not guaranteed to be set according to HAPI base client

### Proposed Changes
Null check

### Change Details
Since HAPI Base Client does not guarantee setting of the operationOutcome, will need to add a null check and produce an empty operationOutcome with the log id

### Security Implications
N/A

### Feedback Requested

Would like to know if it's possible that PII is in our exceptions since the new code takes the message from the exception and puts it into the operation outcome

### DIFF
--- a/dpc-common/src/main/java/gov/cms/dpc/fhir/dropwizard/handlers/exceptions/HAPIExceptionHandler.java
+++ b/dpc-common/src/main/java/gov/cms/dpc/fhir/dropwizard/handlers/exceptions/HAPIExceptionHandler.java
@@ -29,8 +29,11 @@ public class HAPIExceptionHandler extends AbstractFHIRExceptionHandler<BaseServe
     @Override
     Response handleFHIRException(BaseServerResponseException exception) {
         final long exceptionID = super.logException(exception);
-        final OperationOutcome operationOutcome = (OperationOutcome) exception.getOperationOutcome();
-        operationOutcome.setId(exceptionIDtoHex(exceptionID));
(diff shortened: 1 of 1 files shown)


dpc-common/src/main/java/gov/cms/dpc/fhir/dropwizard/handlers/exceptions/HAPIExceptionHandler.java
```diff
@@ -29,8 +30,15 @@ public Response toResponse(BaseServerResponseException exception) {
...
-        final OperationOutcome operationOutcome = (OperationOutcome) exception.getOperationOutcome();
-        operationOutcome.setId(exceptionIDtoHex(exceptionID));
+        OperationOutcome operationOutcome = (OperationOutcome) exception.getOperationOutcome();
+        if (operationOutcome == null) {
+            operationOutcome = new OperationOutcome();
+            operationOutcome.addIssue()
+                    .setCode(OperationOutcome.IssueType.EXCEPTION)
+                    .setSeverity(OperationOutcome.IssueSeverity.ERROR)
+                    .setDetails(new CodeableConcept().setText(exception.getMessage()));
+            operationOutcome.setId(exceptionIDtoHex(exceptionID));
```
embh, line 40: Does this id still also need to be set when the OperationOutcome doesn't start out as null?

whuang85: good catch

PR #880, opened 2020-06-24
DPC-469 Look back corrections

whuang85:
### Proposed Changes
When getting the provider NPI from the roster, it was returning the provider ID instead of the NPI, this has been corrected.  When looking through the EOB, provider NPIs are also in the care team field.  We need to collect all the practitioner NPIs in the EOB to compare with the job.

### Change Details
Collect the practitioner NPIs from care team and provider field of EOB
Return the NPI of the practitioner instead of the database id

### Security Implications

- [x] [security checklist is completed for this change](https://confluence.cms.gov/display/DAPC/2020-07-01%3A+Claims+Lookback+Security+Checklist)


embh:
This looks good to me. Because of the H2 dependency, we might need a security checklist, but I'm not sure because it's limited to the `test` scope. @adhocEric ?

PR #881, opened 2020-06-25
DPC-148 & 362: Simplify web app for one environment

SMLuthi:
### Proposed Changes

Remove code specific to one single environment from the codebase.

### Change Details

At one point the web-app was to be designed to be run in multiple environments (and be able to switch between them at runtime). This idea was scrapped but legacy code endures that supports this implementation. This code needs to be ripped out and streamlined so that the web-app functions only within one environment at run time.

### Security Implications

None


dpc-web/spec/models/organization_user_assignment_spec.rb
```diff
@@ -3,13 +3,18 @@
 require 'rails_helper'
 
 RSpec.describe OrganizationUserAssignment, type: :model do
+  include APIClientSupport
+
   describe 'callbacks' do
     describe '#send_organization_sandbox_email' do
       it 'sends email if org is sandbox_enabled' do
```
dhgreene, line 10: Do we still need this property (and a test for it), since we're now automatically allowing all new sandbox users?  I see references to `:sandbox_enabled?` being replaced with `:api_enabled?`.

dpc-web/app/views/internal/registered_organizations/new.html.erb
```diff
@@ -3,8 +3,8 @@
 <div class="ds-l-container ds-u-padding-x--0">
   <section class="ds-u-display--flex ds-u-justify-content--between ds-u-align-items--center">
 
-    <h1 class="ds-u-font-size--title ds-u-font-weight--normal ds-u-margin--0" data-test="new-reg-org-<%= @registered_organization.api_env %>">
-      Enable <%= @api_env.capitalize %> API access for <%= @organization.name %>
+    <h1 class="ds-u-font-size--title ds-u-font-weight--normal ds-u-margin--0" data-test="new-reg-org" >
+      Enable PI access for <%= @organization.name %>
```
dhgreene, line 7:
```suggestion
      Enable API access for <%= @organization.name %>
```

dpc-web/app/views/dashboard/_assigned_content.html.erb
```diff
@@ -1,4 +1,11 @@
-<p class="ds-text--lead ds-u-measure--wide ds-u-margin-top--2">Welcome to the Data at the Point of Care Pilot. You have been assigned to the organizations below to start using the APIs in a synthetic, or test, sandbox. <strong>Remember</strong>, do NOT add any real patient information to the sandbox environment. </p>
+<p class="ds-text--lead ds-u-measure--wide ds-u-margin-top--2">
+  <% if prod_sbx? %>
+    Welcome to the Data at the Point of Care Sandbox. You have been assigned to the organizations below to start using the APIs in a synthetic, or test, sandbox. <strong>Remember</strong>, do NOT add any real patient information to the sandbox environment.
+  <% else %>
+    <%# Need to speak with Product/H Team about welcome for pilot %>
+    Welcome to the Data at the Point of Care Pilot.
```
dhgreene, line 6: We don't have a web app for pilot

dpc-web/app/services/api_client.rb
```diff
@@ -110,7 +103,7 @@ def delegated_macaroon(reg_org_api_id)
   end
 
   def golden_macaroon
-    @golden_macaroon ||= ENV.fetch("GOLDEN_MACAROON_#{api_env.upcase}")
+    @golden_macaroon ||= ENV.fetch('GOLDEN_MACAROON_SANDBOX')
```
dhgreene, line 106: Is it worth renaming this to `GOLDEN_MACAROON`?  It would require an associated ops PR, but decrease cognitive dissonance in `dev` and `test` to see `GOLDEN_MACAROON_SANDBOX`.

dpc-web/app/controllers/internal/registered_organizations_controller.rb
```diff
@@ -17,49 +16,45 @@ def new
...
 
     def update
       @organization = Organization.find(org_id_param)
-      @registered_organization = @organization.registered_organizations.find(params[:id])
-      @api_env = @registered_organization.api_env
+      @registered_organization = @organization.registered_organization
 
       if @registered_organization.update(registered_organization_params)
-        flash[:notice] = "#{@registered_organization.api_env.capitalize} access updated."
+        flash[:notice] = 'Organization access updated.'
```
dhgreene, line 41: Should "Organization" be "API" in this and the following alerts?

dpc-web/app/models/organization.rb
```diff
@@ -6,7 +6,7 @@ class Organization < ApplicationRecord
   has_one :address, as: :addressable, dependent: :destroy
   has_many :organization_user_assignments, dependent: :destroy
   has_many :users, through: :organization_user_assignments
-  has_many :registered_organizations, dependent: :destroy
+  has_one :registered_organization, dependent: :destroy
```
dhgreene, line 9: Does it still make sense to have two separate models/tables?

dpc-web/app/services/api_client.rb
SMLuthi: Yes, I've taken notes on a few things like this. I didnt include them in this PR because I figured a separate, smaller PR with an associated ops PR would be a little cleaner... rather than trying to dump it all in this one giant one

dpc-web/app/models/organization.rb
SMLuthi: Good question. I could go either way on this tbh. On one hand it contains a clean separation between an org and a registered organization but on the other hand we could just incorporate them together and allow for blank columns. Im open for suggestions.

dpc-web/spec/models/organization_user_assignment_spec.rb
```diff
@@ -3,13 +3,18 @@
 require 'rails_helper'
 
 RSpec.describe OrganizationUserAssignment, type: :model do
+  include APIClientSupport
+
   describe 'callbacks' do
     describe '#send_organization_sandbox_email' do
```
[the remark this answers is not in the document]
SMLuthi: Currently we dont have prod emailing setup at all. It just doesnt exist. Im assuming, at some point, we will introduce that. We can make changes now or wait till that work comes down the pipeline.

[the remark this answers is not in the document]
Sun-Mountain: Also right now the emails sent to sandbox have language specific _to_ the sandbox environment (i.e. naming it, reminding users not to use real patient data etc etc.

dpc-web/app/models/organization.rb
Sun-Mountain: This was something that we were definitely talking about during our work.

My understanding is a registered organization is what exists in the API (and is copied from & updated w/ the organization model).

dpc-web/app/views/dashboard/_assigned_content.html.erb
Sun-Mountain: Will it always be sandbox?

dpc-web/app/controllers/internal/registered_organizations_controller.rb
Sun-Mountain: Changed!

dpc-web/app/views/dashboard/_assigned_content.html.erb
Sun-Mountain: Reverted, but right now the language does say that they are working in Sandbox and to not upload real patient data to the API. Will the pilot users ever use the web app?

dpc-web/spec/models/organization_user_assignment_spec.rb
Sun-Mountain: Changed.

PR #912, opened 2020-07-14
dpc-502 Feature: Investigate prod gems

SMLuthi:
### Fixes [DPC-502](https://jira.cms.gov/browse/DPC-502)

A minor incident has results in an test file being pushed to a production envrionment. This came up during an audit which triggered an alarm.

### Proposed Changes

We needed to investigate whether there are any gems being installed that are either unnecessary OR did not belong in prod but, rather, only test or dev.

### Change Details

The Gemfile has been reorganized a bit. A few gems have moved around and a few gems have been removed entirely since it looks like the web-app is not using them.

### Security Implications

### Acceptance Validation

I had to update two seemingly unrelated test files that were failing for me locally. Oddly enough, I dont think this is failing any other PRs. @Sun-Mountain could maybe bring some insight into this as I believe she worked on this portion of the prod/sandbox cleanup.


SMLuthi:
@Sun-Mountain Hey, I wanted to point out that I removed both the `jbuilder` and the `jquery-rails` gems from the Gemfile. I couldn't find anywhere that these were being used and I figured we'd take em out. Let me know if I overlooked something... otherwise we can always add them back later when we DO use them. Also, I moved `brakeman` to the `dev/test` section as I dont think we use that in production (Again, let me know if im wrong).

dpc-web/Gemfile
```diff
@@ -62,15 +71,7 @@ group :development do
...
-gem 'octokit'
-gem 'kaminari'
-gem 'active_model_serializers'
-gem 'macaroons'
-gem 'newrelic_rpm', '~> 6.10'
-gem 'fhir_client'
-gem 'lograge'
-gem 'sidekiq', '~> 6.0'
-gem 'sidekiq_alive'
+group :development, :test do
```
dhgreene, line 74: Is there benefit to putting these in a separate group than the existing group above?

SMLuthi: You can only declare a gem once and if you want a gem to live in certain environments you have to create groups for each env(s). So gems that should exist in both dev and test need to be a in a group defined like that... its a little annoying

dhgreene: But there's already a group defined like that earlier in the file, correct?

SMLuthi: you are correct >_<

SMLuthi: I was moving gems around forgot to change that, one moment

PR #913, opened 2020-07-16
DPC-545 configure container to run integration tests in it

whuang85:
### Fixes [DPC-545](https://jira.cms.gov/browse/DPC-545)

In Jenkins, patient everything integration test does not have access to the /tmp folder on the host (where aggregation puts it's files) so to solve this, we will run the tests in a container where we can map the /tmp folder as a volume to the container.

### Proposed Changes

- Run api integration tests in container
- Made jobTimeoutInSeconds configurable via SSM

### Change Details

Using the official maven docker image to run the mvn test command in that container instead of directly in jenkins. This allows us to map the /tmp folder of the host to the container so when patient everything runs it can access the files that aggregation generates. Also cleaned up the jobTimeoutInSeconds configuration so it can be configured via SSM.

### Security Implications

- [x] requires more information or team discussion to evaluate security implications

### Feedback Requested

Do I need a security checklist for bringing in the new docker image although it's only ran locally (via make ci-app) and in jenkins and travis?


docker-compose.yml
```diff
@@ -4,6 +4,7 @@ services:
 
   db:
     image: postgres
+    command: postgres -c 'max_connections=250'
```
embh, line 7: Is there any downside to setting this higher than the default?

whuang85: I don't believe so, when I was testing this locally, I kept running out of connections. I believe the default is 100.

dpc-api/src/main/resources/application.conf
```diff
@@ -2,6 +2,9 @@ dpc.api {
     include "server.conf"
     include "queue.conf"
 
+    jobTimeoutInSeconds = ${?JOB_TIMEOUT_IN_SECONDS}
+    jobTimeoutInSeconds = 30
+
```
jonfulk, line 7: Is this a way of defaulting to 30 s if no value is set for `JOB_TIMEOUT_IN_SECONDS`? If there is a value set, does it just use the first one and ignore the second line?

whuang85: Yep that is exactly what those two lines do.

PR #921, opened 2020-07-21
DPC-445 check provider in provenence header and group match

whuang85:
### Fixes [DPC-445](https://jira.cms.gov/browse/DPC-445)

It was never required that the practitioner defined in provenance matches the provider in group

### Proposed Changes

- extract the practitioner uuid from provenance header and get the practitioner from attribution
- compare the practitioner NPI with the group practitioner NPI

### Change Details

Extract the UUID of the practitioner from provenance header and get the NPI by calling attribution service and then compare it with the NPI in the group.

### Security Implications


dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
```diff
@@ -431,7 +433,29 @@ private void logAttestation(Provenance provenance, UUID rosterID, Group attribut
...
+                    .execute();
+
+            Identifier provenancePractitionerNPI = FHIRExtractors.findMatchingIdentifier(practitioner.getIdentifier(), DPCIdentifierSystem.NPPES);
+            String groupPractitionerNPI = FHIRExtractors.getAttributedNPI(attributionRoster);
+
+            if (!provenancePractitionerNPI.getValue().equals(groupPractitionerNPI)) {
+                throw new WebApplicationException("Provenance header's provider does not match group provider");
+            }
+        } catch(ResourceNotFoundException e) {
+            throw new WebApplicationException("Could not find provider defined in provenance header");
```
embh, line 456: Should we set a status code on these two exceptions so it doesn't use the default of 500? My initial thought was `401 Unauthorized`, but maybe `422 Unprocessable Entity` is more accurate because it's the attestation that's a problem, not their authentication. 🤔

richbraman: I would think 422 would be most appropriate, remember this should require a swagger annotation indicating 422 as a potential API response....

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
```diff
@@ -431,7 +437,29 @@ private void logAttestation(Provenance provenance, UUID rosterID, Group attribut
...
+                    .execute();
+
+            Identifier provenancePractitionerNPI = FHIRExtractors.findMatchingIdentifier(practitioner.getIdentifier(), DPCIdentifierSystem.NPPES);
+            String groupPractitionerNPI = FHIRExtractors.getAttributedNPI(attributionRoster);
+
+            if (!provenancePractitionerNPI.getValue().equals(groupPractitionerNPI)) {
+                throw new WebApplicationException("Provenance header's provider does not match group provider", Response.status(422).build());
+            }
+        } catch(ResourceNotFoundException e) {
+            throw new WebApplicationException("Could not find provider defined in provenance header", Response.status(422).build());
```
embh, line 460: It looks like we've established a pattern of using constants for response codes, either from `Response.Status` or `HttpStatus`, so it may make sense to do that here (e.g., `HttpStatus.UNPROCESSABLE_ENTITY_422`).

PR #933, opened 2020-07-29
DPC-392: Break apart integration and unit tests

embh:
### Fixes [DPC-392](https://jira.cms.gov/browse/DPC-392)

JUnit tests in DPC include a mix of unit and integration tests. To provide the option of running them separately, we need to distinguish them in some way.

### Proposed Changes

* Annotate test classes with `@IntegrationTest` if they rely on any other service
* Create Maven `unit-tests` profile to exclude classes from the `Integration` group
  * There is an existing `integration-tests` profile that includes only the `Integration` group
* Create `dpc-unit-test.sh` script to run tests from the Maven `unit-tests` profile
* Add a handful of new unit tests in dpc-api as a starting point

### Change Details

In an attempt to change as little existing code as possible, new unit tests for API resources have been placed in classes called `*ResourceUnitTest`. In some cases, however, the original test classes (e.g., `DataResourceTest`) may already contain only unit tests. Look for an `@IntegrationTest` annotation to be sure.

Note that Code Climate reports a decrease in coverage, even though this PR adds unit tests. This is due to more tests being identified as integration tests when they were previously treated as unit tests in terms of coverage. Unit test development will continue in follow-up tasks.

### Security Implications

No security implications. Changes apply only to tests.

### Acceptance Validation

`./dpc-unit-test.sh` will run only the unit tests.

### Feedback Requested

Any


dpc-unit-test.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
whuang85, line 1: Shouldn’t this script be called in the make file somewhere?

embh: Not quite yet. The goal of this ticket was to introduce separation of the test types so we're positioned to run them separately. The tests do all still run with `make ci-app`.

PR #935, opened 2020-07-29
DPC-557 - add check to make sure _since parameter is not a future date

richbraman:
### Fixes [DPC-557](https://jira.cms.gov/browse/DPC-557)

_since parameter accepts a future date

### Proposed Changes
reject request if since_parameter is a date in the future

### Change Details
changed  function that checks /Group/$export request to enforce check on _since parameter

no further detail
### Security Implications
None

- [X ] no PHI/PII is affected by this change

### Acceptance Validation

Yes, with postman

### Feedback Requested

General


dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
```diff
@@ -337,6 +339,10 @@ private OffsetDateTime handleSinceQueryParam(String since) {
         try {
             var dt = new DateTimeDt();
             dt.setValueAsString(since);
+            LocalDate today = LocalDate.now( ZoneOffset.UTC );
```
dhgreene, line 342: Doesn't `_since` allow much more precision?  Only considering the day wouldn't allow `_since` to be set to a time earlier in the day of the query.  There might be cases where this is useful (such as a known data dump, unit tests, etc.).

richbraman: I think the precision is there.  I just named the variable "today", maybe should be called rightNow?

richbraman: ahh maybe should be LocalDateTime.now().
i am changing precision

dhgreene:
Could you add positive/negative unit tests for this?  I'd especially like to see it exercised using a _since parameter with multiple time zones, where a time zone bug would violate the check:

- _since a few seconds ago (in Puerto Rico and Guam timezones) should be allowed
- _since a few seconds from now (in Puerto Rico and Guam timezones) should not be allowed
- _since yesterday (improperly specified without timezone) should be allowed

richbraman:
I think we need this first
https://jira.cms.gov/browse/DPC-579

The current code only parses the string date into a datetime data type.  It
doesn't do the proper checks.

dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/GroupResourceUnitTest.java
```diff
@@ -157,4 +171,112 @@ public void testCreateRosterProviderNotFound() {
...
+                .named("patients")
+                .withParameters(Mockito.any(Parameters.class))
+                .returnResourceType(Bundle.class)
+                .useHttpGet()
+                .encodedJson()).thenReturn(bundleOperation);
+
+        Mockito.when(bundleOperation.execute()).thenReturn(new Bundle());
+        Meta bfdTransactionMeta = new Meta();
+        Mockito.when(mockBfdClient.requestPatientFromServer(SYNTHETIC_BENE_ID, null).getMeta()).thenReturn(bfdTransactionMeta);
+
```
jonfulk, line 207: When Will's related PR gets merged, we might want to move this stuff into the `setUp` method.

jonfulk (approved):
🥅

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
```diff
@@ -331,24 +331,19 @@ private Group executeGroupOperation(UUID rosterID, Group groupUpdate, String ope
         return resources;
     }
 
-    /**
-     * Convert the '_since' {@link QueryParam} to a Date
-     *
-     * @param since - {@link String} an instant
-     * @return - A {@link OffsetDateTime} for this since.
-     */
```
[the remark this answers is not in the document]
SalGutierrez: Yea, it's a private method and doesn't tell anything more than the signature of the method does. Thought it was distracting.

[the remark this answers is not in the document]
jonfulk: makes sense

dhgreene (approved):
✈️

PR #938, opened 2020-07-29
DPC-567 - change "kind" to "instance"

richbraman:
### Fixes [DPC-567](https://jira.cms.gov/browse/DPC-567)

- Conformance tests failing due to missing operations in Metadata (CapabilityStatement)

### Proposed Changes

The CapabilityStatement instance represents the present capabilities of a specific system instance. This is the kind returned by /metadata for a FHIR server end-point.

If kind = instance, "implementation" must be present and software may be present per rule cpb-14, so this fix adds an implementation element .

The conformance testing tool alerted that we also need to list the `patient-export` operation as well, which is the same as our `patient-everything`. Renamed that operation to match conformance tool expectations.

### Change Details

http://hl7.org/fhir/capability-statement-kind
http://hl7.org/fhir/valueset-capability-statement-kind.html (R4)
https://www.hl7.org/fhir/stu3/valueset-capability-statement-kind.html. (STU3)
http://hl7.org/fhir/codesystem-capability-statement-kind.html#capability-statement-kind-instance

### Security Implications

No

- [X ] no PHI/PII is affected by this change

### Acceptance Validation

Tested with BDT conformance tool (using version pinned to this commit: bc64ac546d309969034fca3c9f9a5a01d4d7563f)

Screen Shot 2020-10-27 at 10 45 59 AM

### Feedback Requested

General


whuang85 (dismissed):

dpc-api/src/main/resources/DPCCapabilities.json
```diff
@@ -202,8 +202,8 @@
           }
         },
         {
-        "name": "patient-everything",
-        "definition": { 
+        "name": "patient-export",
```
whuang85, line 205: shouldn't this be patient-everything?

jonfulk: Not according to the conformance tool. It was failing due to not having the `patient-export` operation listed, and it passes like this.

whuang85: is it because we place patient everything endpoint in the  `patientExportEndpoint` of the BDT tool?

jonfulk: could be... i'll switch it back and remove the `patientExportEndpoint` from the config... it may just say that `patient-export` is not supported by our API, which I suppose is technically true?

whuang85: yep, we don't support patient-export as outlined in the bulk data export, don't know if we need to add a patient-everything to our capability statement

jonfulk: Ok, you're right that the issue was that I was listing the `patient-everything` endpoint as the `patientExportEndpoint` in the config, causing the test to require it in the operations. With an update to the config and a reversion to `patient-everything`, the conformance tests still pass. I'll update the PR momentarily.
Screen Shot 2020-10-29 at 1 16 41 PM

whuang85 (approved):

PR #940, opened 2020-07-30
Silvio/dpc 163 complete akamai setup

SMLuthi:
### Proposed Changes

We are seeing 404 errors in Splunk in regards to Akamai SureRoute checks not finding the `akamai-test-object/html` page. This is used by SureRoute to find the most efficient routes to the web page. We need to add this static asset (as per these instructions https://learn.akamai.com/en-us/webhelp/ion/oca/GUID-8AF6D7C4-EF29-486F-89B3-F5037C2FF422.html).

### Change Details

Add a static page (pulled directly from Akamais test object sample they provide) to the public folder in the web app.

### Security Implications

None, its a static html test file provided by Akamai, it does not hit the back end at all.

### Feedback Requested

Note: Currently this routes to `/akamai-test-object.html` as that is where Akamai is looking for it. However, I would like to change this to something like `/akamai/akamai-test-object.html` or `/_health/akamai-test-object.html` to embed this a little more and keep it organized.

Splunk logs are registering 200s now for this check instead of 404s after deploying on dev


dhgreene:
> Note: Currently this routes to /akamai-test-object.html as that is where Akamai is looking for it. However, I would like to change this to something like /akamai/akamai-test-object.html or /_health/akamai-test-object.html 

Shall we do this now, rather than deferring the change?

dhgreene:
> Add a static page (pulled directly from Akamais test object sample they provide) to the public folder in the web app.

Given the number of comments from codeclimate on their sample page, I'd prefer we return our own custom page.

SMLuthi:
@dhgreene @Sun-Mountain This PR should be ready for review. Akamai versions have been deployed to all environments to change the SureRoute routing to `/akamai/akamai-test-object.html` so thats is embedded and a bit cleaner than just pinging it off of the root url. This has been verified by deploying it to dev and checking the logs on splunk to make sure that not only the route is up to date but that its returning a 200.

PR #953, opened 2020-08-05
logged org id in jwt key resolver

SalGutierrez:
### Security Implications


dpc-api/src/main/java/gov/cms/dpc/api/auth/jwt/JwtKeyResolver.java
```diff
@@ -50,7 +50,7 @@ public Key resolveSigningKey(JwsHeader header, Claims claims) {
         final PublicKeyEntity keyEntity;
         try {
             keyEntity = this.dao.fetchPublicKey(organizationID, UUID.fromString(keyId))
-                    .orElseThrow(() -> new WebApplicationException(String.format("Cannot find public key with id: %s", keyId), Response.Status.UNAUTHORIZED));
+                    .orElseThrow(() -> new WebApplicationException(String.format("Cannot find public key with id: %s, org id: %s", keyId,organizationID.toString()), Response.Status.UNAUTHORIZED));
```
whuang85, line 53: can we also get an actual log statement so it goes to splunk?  I don't think this exception is logged anywhere in splunk once it's thrown.

PR #954, opened 2020-08-05
DPC-234: Delete client tokens from web app

Sun-Mountain:
### Fixes [DPC-234](https://jira.cms.gov/browse/DPC-234)

As a user, I need the ability to remove one of my tokens so that I can rotate them or remove a lost token.

This will be on the website after login.

An API delete function already exists to do this but there is no UI.

### Proposed Changes

- Client token delete function added.
- UI for web app
- Response styling fixes

### Change Details

- UI
Screen Shot 2020-08-05 at 12 29 18 PM
Screen Shot 2020-08-05 at 12 29 25 PM

- Responsive styling fixes
Screen Shot 2020-08-05 at 12 30 24 PM
Screen Shot 2020-08-05 at 12 30 33 PM

### Security Implications


dpc-web/app/services/api_client.rb
```diff
@@ -71,7 +77,7 @@ def get_public_keys(reg_org_api_id)
   end
 
   def response_successful?
-    @response_status == 200
+    [200, 204].include? @response_status
```
dhgreene, line 80: Good.  Better would be to allow anything in the 20x range--201 and 202 are likely to be used, for instance.

Sun-Mountain: @SMLuthi

SMLuthi: @dhgreene @Sun-Mountain in that case we can do something like `(200..299).include? @response_status` which checks over a range

dpc-web/app/controllers/client_tokens_controller.rb
```diff
@@ -6,42 +6,60 @@ class ClientTokensController < ApplicationController
...
   def create
     @organization = current_user.organizations.find(params[:organization_id])
 
     reg_org = @organization.registered_organization
     manager = ClientTokenManager.new(registered_organization: reg_org)
-    if missing_params
-      render_error 'Label required.'
-    elsif manager.create_client_token(label: params[:label])
+
+    if params_present? && manager.create_client_token(label: params[:label])
```
SMLuthi, line 18: Would this be a little cleaner?

```
return render_error 'Label required.' unless params_present?

if manager.create_client_token(label: params[:label])
  @client_token = manager.client_token
  render :show

render_error 'Client token could not be created.'
```

Sun-Mountain: The tests did not like that >_>

SMLuthi: Hm, that shouldnt happen. I wonder if the render_errors arent being hit correctly in the way you defined them.

PR #968, opened 2020-08-11
Silvio/add simplecov to spec runs

SMLuthi:
### Fixes [DPC-624](https://jira.cms.gov/browse/DPC-624)

We are currently not running any kind of coverage tools when our test suite is run.

### Proposed Changes

SimpleCov was already installed as a gem. We should be taking advantage of this.

### Change Details

Add `SimpleCov` to spec_helper config file.
Include the following in the config:

- Group folders that aren't automatically grouped 
- Filter out files that dont need to be tested (Mainly passthrough classes)
- Set a `minimum coverage` to %90
- Set a `minimum coverage by file` to %80

Tests have been added for certain files that did not have enough coverage.
Tickets were created for files that have no coverage or were not being tested correctly.
Some code has been removed that was not being used at all.

### Security Implications

### Feedback Requested

Note #1: The above coverage values of 90%/80% are currently met but some code is not being tested. Tickets have been created to rectify this.
Note #2: SimpleCov had to be reduced by one version due to the fact that our Sonarqube cannot digest the json thats being output from version `0.18`. Hopefully, either Sonarqube or SimpleCov will fix this in the future.
Note #3: SimpleCov not only stores its data in a results.json file but it also provides an html page were you can see the results in a neat table. Heres an example...

image


dhgreene (approved):
👍

PR #974, opened 2020-08-13
DPC-582: Return HTTP 410 for jobs older than 24 hours

embh:
### Fixes [DPC-582](https://jira.cms.gov/browse/DPC-582)

To minimize the amount of time DPC stores patient data, export jobs should expire after some period of time.

### Proposed Changes

Expire export jobs after 24 hours, making them and their data inaccessible.

### Change Details

* When a job is completed, its status endpoint will include an `Expires` header to indicate when the job will no longer be available
* 24 hours after the last batch of a job completes:
  * Job status will return `410 Gone`
  * Data file requests will return `410 Gone`

Documentation updates to reflect these changes will be coordinated with our human-centered team.

### Security Implications

This change aims to improve our security posture by limiting access to data.

Export job data files, which contain PII and PHI, will now have time-limited access.

### Acceptance Validation

Tests added for job status and data file requests.

Expires header in dev:
Screen Shot 2020-08-14 at 9 36 16 AM

Response in dev when all batches in job have a `complete_time` of > 24 hours ago:
Screen Shot 2020-08-14 at 9 39 56 AM

### Feedback Requested

Any


dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/DataResource.java
```diff
@@ -130,6 +139,17 @@ public Response downloadExportFile(@ApiParam(hidden = true) @Auth OrganizationPr
 
         final FileManager.FilePointer filePointer = this.manager.getFile(organizationPrincipal.getID(), fileID);
 
+        // If job is expired, the files should no longer be accessible
+        List<JobQueueBatch> batches = queue.getJobBatches(filePointer.getJobID());
+        Set<JobStatus> jobStatusSet = batches.stream().map(JobQueueBatch::getStatus).collect(Collectors.toSet());
+        if (jobStatusSet.size() == 1 && jobStatusSet.contains(JobStatus.COMPLETED)) {
```
jonfulk, line 145: Does this mean that we will NOT return a 410 on a job if the batches are not all completed, even if expired? That sounds fine because all batches should be completed within 24 hours, but I just wanted to ask the question to be sure that there isn't a missed edge case here.

dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/JobResourceTest.java
```diff
@@ -129,6 +132,10 @@ public void testSuccessfulJob() {
         final Response response = resource.checkJobStatus(organizationPrincipal, jobID.toString());
         assertAll(() -> assertEquals(HttpStatus.OK_200, response.getStatus()));
 
+        var expires = ZonedDateTime.parse(response.getHeaderString("Expires"), JobResource.HTTP_DATE_FORMAT);
+        assertAll(() -> expires.isAfter(ZonedDateTime.now().plusHours(23)),
+                () -> expires.isAfter(ZonedDateTime.now().plusHours(25)));
```
jonfulk, line 137: why do you need both of these assertions? if the second is true, the first is as well.

jonfulk (dismissed):
A couple clarifying questions, but looks good!

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/DataResource.java
embh: Great question. The job isn't considered expired until 24 hours after *all* of its batches are completed.

dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/JobResourceTest.java
embh: You may have caught a bug; that second one should be `.isBefore()`! Taking a closer look.

embh: Fixed! Thanks for catching this.

PR #989, opened 2020-08-18
DPC-403 Feat: Run unit tests in parallel

jonfulk:
### Fixes [DPC-403](https://jira.cms.gov/browse/DPC-403)

Testing slows us down during dev and deploy. This is a step in the direction of trying to speed them up.

### Proposed Changes

- Added a make command to facilitate running just the unit tests
- Updated parent pom to run unit tests in parallel for entire project with a maximum of 10 threads (could be changed)

### Change Details

This change makes little difference on my own box, but theoretically, they could speed up deployments (TBD).
No race conditions were noticed when running the tests locally, but this could potentially be different when deployed.
Currently, the max. number of threads is set to 10, but that could be increased or decreased as needed.

### Security Implications

### Acceptance Validation

- `make ci-app` has no errors when run locally

Screen Shot 2020-08-18 at 3 26 17 PM

### Feedback Requested

Open to thoughts on whether or not this change is really necessary as well as how many maximum threads we should allocate to running unit tests in parallel.


jonfulk:
I'm leaning toward thinking that this change is not making a difference for some reason. When looking at Travis CI, there doesn't appear to be any difference in build time between this PR and others that do not attempt to run tests in parallel. Thoughts?

dhgreene:
> When looking at Travis CI, there doesn't appear to be any difference in build time between this PR and others that do not attempt to run tests in parallel.

For Travis in particular, I don't think we'll see parallel tasks without a [configuration change](https://docs.travis-ci.com/user/build-matrix/).

jonfulk:
> > When looking at Travis CI, there doesn't appear to be any difference in build time between this PR and others that do not attempt to run tests in parallel.
> 
> For Travis in particular, I don't think we'll see parallel tasks without a [configuration change](https://docs.travis-ci.com/user/build-matrix/).

That's good to know but also has me worried about deploying this to prod without first seeing the tests run in parallel. With more threads, any potential race conditions may reveal themselves at the wrong place and time.

jonfulk:
Deployed the build on dev, and no significant differences in build time were evident:

Screen Shot 2020-08-20 at 1 10 15 PM

PR #991, opened 2020-08-18
DPC-628-Reverted back to INFO log level in prod. Changed a bbclient log from DEBUG to INFO

SalGutierrez:
### Fixes [DPC-628](https://jira.cms.gov/browse/DPC-628)

We had originally intended to enabled DEBUG logging in prod in order to see a specific debug log for troubleshooting, due to PII concerns with other unrelated debug logs we have decided against enabling DEBUG logs. 

### Proposed Changes

- Revert to INFO log levels for "gov.cms.dpc" loggers (API, Attribution, and Aggregation)
- Log the required log statement in bbclient as INFO 

### Change Details

- Reverted to INFO log levels for "gov.cms.dpc" loggers (API, Attribution, and Aggregation)
- Logged the required log statement in bbclient as INFO instead of DEBUG

### Security Implications

We will be logging the MBI hash.

It was mentioned that the MBI hash is PII/psuedo-PII but only linkable not directly linked.

 logger.info("Attempting to fetch patient with MBI hash {} from baseURL: {}", mbiHash, client.getServerBase());

- [ X] requires more information or team discussion to evaluate security implications

### Acceptance Validation

Original intention was to enable dubug logging, new AC is to print the specific bbclient log as INFO

 logger.info("Attempting to fetch patient with MBI hash {} from baseURL: {}", mbiHash, client.getServerBase());
### Feedback Requested

CONFIRM MBI Hash is not PII or PHI

### DIFF
--- a/dpc-aggregation/src/main/resources/prod.application.conf
+++ b/dpc-aggregation/src/main/resources/prod.application.conf
@@ -10,9 +10,4 @@ dpc.aggregation {
 
   exportPath = "/app/data"
 
-  logging {
-    loggers {
-      "gov.cms.dpc" = DEBUG
-    }
-  }
 }
--- a/dpc-api/src/main/resources/prod.application.conf
+++ b/dpc-api/src/main/resources/prod.application.conf
@@ -19,10 +19,4 @@ dpc.api {
 
     attributionURL = "http://backend.dpc-prod.local:8080/v1/"
     exportPath = "/app/data"
-
-  logging {
-    loggers {
-      "gov.cms.dpc" = DEBUG
-    }
-  }
 }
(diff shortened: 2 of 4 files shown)


StewGoin:
An MBI Hash is the weird world of "it's PII but its hashed so it's protected PII in that form" so in many ways it's like sharing an encrypted file that has PII in it, but can safely traverse a network as an email attachment. Is PII involved? Yes. But is it at risk? In this case no -- that's the kind of explanation that helps manage risk as we do PR creation/reviews/approvals so the change and risk management is clear a week or a month from now.

Why does that matter? If the hashing approach is later found to be flawed or reversible (let's trust in math that it's not) -- we are more aware of potential exposure retrospectively of something like an MBI that's also linked to other elements like service dates, etc... but at least it's not a HICN where an SSN is also at risk.

PR #992, opened 2020-08-19
Silvio/dpc 156 rate limit emails

SMLuthi:
### Fixes [DPC-156](https://jira.cms.gov/browse/DPC-156)

Currently, a user can spam emails via the reset password request, the confirmation email request as well as the sandbox organization email request. This needs to be throttled to avoid a potential attack.

### Proposed Changes

Enable email throttling by allowing a user to only make a certain amount of email requests in a given time.

### Change Details

Use Redis to create a key value based throttling system. Increment a key (using the email) every time an email request is made and set it to expire in a given time. If the user hits an established threshold within that time with the amount of email requests they have made it will silently not queue anymore emails (as per given instructions of the ticket). Once the key has expired the user will be free to start sending emails again if need be.

Currently, the configuration for the email limit and the expiration time has been set to 5 emails and 300 seconds. This can be changed within the `application.rb` file.

### Security Implications

### DIFF
--- a/dpc-web/Gemfile
+++ b/dpc-web/Gemfile
@@ -59,6 +59,7 @@ group :test do
   gem 'climate_control'
   gem 'rails-controller-testing'
   gem 'fakefs', require: 'fakefs/safe'
+  gem 'fakeredis'
 end
 
 group :development do
--- a/dpc-web/Gemfile
+++ b/dpc-web/Gemfile
@@ -35,6 +35,7 @@ gem 'kaminari'
 gem 'active_model_serializers'
 gem 'macaroons'
 gem 'lograge'
+gem 'redis-namespace'
 
 group :development, :test do
   gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
--- a/dpc-web/app/models/organization_user_assignment.rb
+++ b/dpc-web/app/models/organization_user_assignment.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require './lib/redis_store/mail_throttle_store'
+
 class OrganizationUserAssignment < ApplicationRecord
   belongs_to :organization
   belongs_to :user
@@ -12,9 +14,11 @@ class OrganizationUserAssignment < ApplicationRecord
   def send_organization_sandbox_email
     return unless organization.prod_sbx? && organization.registered_organization.present?
 
+    mail_throttle_store = RedisStore::MailThrottleStore.new
+
     UserMailer
       .with(user: user, vendor: organization.health_it_vendor?)
       .organization_sandbox_email
-      .deliver_later
+      .deliver_later if mail_throttle_store.can_email? user.email
   end
 end
--- a/dpc-web/config/application.rb
+++ b/dpc-web/config/application.rb
@@ -53,5 +53,9 @@ module DpcWebsite
     config.action_mailer.delivery_job = "ActionMailer::MailDeliveryJob"
 
     config.to_prepare { Devise::Mailer.layout "mailer" }
+
+    # Mail throttling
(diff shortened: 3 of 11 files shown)


dpc-web/app/models/organization_user_assignment.rb
```diff
@@ -12,9 +14,13 @@ class OrganizationUserAssignment < ApplicationRecord
   def send_organization_sandbox_email
     return unless organization.prod_sbx? && organization.registered_organization.present?
 
-    UserMailer
-      .with(user: user, vendor: organization.health_it_vendor?)
-      .organization_sandbox_email
-      .deliver_later
+    mail_throttle_store = RedisStore::MailThrottleStore.new
+
+    if mail_throttle_store.can_email? user.email
```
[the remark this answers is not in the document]
SMLuthi: Ironically, this yells at me if I use a guard clause because the code, even though its a single line execution, is spaced out on multiple lines.

dpc-web/config/application.rb
```diff
@@ -53,5 +53,9 @@ class Application < Rails::Application
     config.action_mailer.delivery_job = "ActionMailer::MailDeliveryJob"
 
     config.to_prepare { Devise::Mailer.layout "mailer" }
+
+    # Mail throttling
+    config.x.mail_throttle.limit = 10 # Limit to 10 emails before hard stop
```
dhgreene, line 58: Is this worth allowing an env var to override, in case we want to dial this back during an event?  

Also, is 10 too high for a default?  Allowing for 1 email confirmation, 1 password reset, 1 email for future purposes, multiplied by a 2x "oops" factor still only gives 6 emails--and that sequence might take more than 5 minutes.

dpc-web/Gemfile
```diff
@@ -35,6 +35,7 @@ gem 'kaminari'
 gem 'active_model_serializers'
 gem 'macaroons'
 gem 'lograge'
+gem 'redis-namespace'
```
dhgreene, line 38: Can you mention the new dependencies in the PR description, with some thoughts about selection criteria?

dpc-web/config/application.rb
SMLuthi: I've thought about this as well. The issue is that once the rails web app is running it will have loaded the config already. Meaning that even if we change the env var it wont be reflected in the running app... we'd have to either run another deploy (or perhaps just killing the current containers may do it). But it may just be worth moving this to an env var (with a default) to make it easier to manage.

As for the 10 limit email, that was suggested by the ticket but I agree it may be high. Maybe we should lower this to 5?

@dhgreene

dpc-web/spec/features/confirmation/user_confirmation_email_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+  end
+
+  context 'with too many emails' do
+    before do
+      Rails.configuration.x.mail_throttle.limit = 1
+    end
+
+    after do
+      Rails.configuration.x.mail_throttle.limit = 10
```
dhgreene, line 53: Should this be setting it back to its original value, rather than 10?

dpc-web/spec/models/organization_user_assignment_spec.rb
```diff
@@ -36,6 +37,33 @@
...
         expect(UserMailer).not_to have_received(:with)
       end
+
+      context 'when mail rate limit has been reached' do
+        before do
+          Rails.configuration.x.mail_throttle.limit = 0
+        end
+
+        after do
+          Rails.configuration.x.mail_throttle.limit = 10
```
dhgreene, line 47: . . . and again

PR #993, opened 2020-08-20
DPC-389 tweak logging to one line

whuang85:
### Fixes [DPC-389](https://jira.cms.gov/browse/DPC-389)

Have the logging on one line to easily correlate the different variables used by look back easier by EoB

### Proposed Changes

* Move check of look back variables to a private method so can return after logging the different variables
* Add some more tests to look back

### Security Implications

The only new logging change here is that we can now correlate all the different eob information with the eob itself through the claim id


dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/service/LookBackServiceImpl.java
```diff
@@ -39,74 +42,71 @@ public LookBackServiceImpl(RosterDAO rosterDAO, OrganizationDAO organizationDAO,
...
 
         Set<String> eobProviderNPIs = extractPractionerNPIs(explanationOfBenefit);
 
-        LOGGER.info("billingPeriod={}", billingPeriod.orElse(null));
-        LOGGER.info("eobOrganizationID={}", eobOrganizationID.orElse(null));
-        LOGGER.info("jobOrganizationID={}", organizationID.orElse(null));
 
-        if (billingPeriod.isEmpty() || providerID.isEmpty() || organizationID.isEmpty() || eobOrganizationID.isEmpty()) {
+        boolean passLookBack = passLookBack(billingPeriod, providerNPI, organizationID, eobOrganizationID, eobProviderNPIs, withinMonth);
+        LOGGER.info("lookBackDateCompare={}-{}, providerCompare={}-{}, organizationCompare={}-{}, passLookBack={}",
```
embh, line 70: Would it be easier for us to shape the Splunk results if we give each value its own key? I'm thinking that would allow more flexibility than putting values together with `-`.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/service/LookBackServiceImpl.java
```diff
@@ -39,74 +42,68 @@ public LookBackServiceImpl(RosterDAO rosterDAO, OrganizationDAO organizationDAO,
...
+        Optional<String> optionalProviderID = Optional.ofNullable(providerID);
+        Optional<String> optionalOrganizationID = Optional.ofNullable(organizationID);
+        Optional<String> optionalEobOrganizationID = Optional.ofNullable(eobOrganizationID);
+
+        boolean result = false;
+        if (optionalBillingPeriod.isPresent() && optionalProviderID.isPresent() && optionalOrganizationID.isPresent() && optionalEobOrganizationID.isPresent()) {
+            long lookBackMonthsDifference = getMonthsDifference(optionalBillingPeriod.get(), operationsConfig.getLookBackDate());
+            boolean eobContainsProvider = eobProviderNPIs.contains(optionalProviderID.get());
+            boolean eobRelatedToOrganization = optionalOrganizationID.get().equals(optionalEobOrganizationID.get());
+            boolean eobWithinLookBackLimit = lookBackMonthsDifference < withinMonth;
```
jonfulk, line 87: For DPC-572, it would be helpful to have these 3 booleans as well as the `lookBackMonthsDifference` included in the logs. Could we return a list of these values to the logging statement and log those as well as the result of the aggregation of them?

SalGutierrez: I think we used to have them but now it prints out the value found in EOB and the Value found in the JOB, wile the match result is not provided we can perform that check in Splunk.

SalGutierrez (dismissed):

StewGoin:
I don't have any issues with this that would prevent it from being merged and pushed -- we should make sure that any documentation regarding our logging stuff reflects these changes if they prove out to be something we want to keep in the long run (which seems likely to me, at least for the pilot).

StewGoin:
I will say -- if we do intend for this to be the typical logging "behavior" in the long-term or in MVP, I think we'd be better suited to accomplish the math and aggregation info within DPC and log out the aggregated stats, etc... rather than logging out the raw PII elements -- that would be an appropriate way to minimize PII/PHI data.

I'd request that someone backlog a ticket to revisit and potentially rework how we accomplish this once we're past the current outage.

whuang85:
> I will say -- if we do intend for this to be the typical logging "behavior" in the long-term or in MVP, I think we'd be better suited to accomplish the math and aggregation info within DPC and log out the aggregated stats, etc... rather than logging out the raw PII elements -- that would be an appropriate way to minimize PII/PHI data.
> 
> I'd request that someone backlog a ticket to revisit and potentially rework how we accomplish this once we're past the current outage.

https://jira.cms.gov/browse/DPC-661

PR #997, opened 2020-08-24
DPC-640 Persist MBI as uppercase

SalGutierrez:
### Security Implications
No new data is stored. 

### Acceptance Validation

- No "bad" MBIs in DB.
  - A script (attached to ticket) will need to be ran in order to convert previously stored MBIs to uppercase.
`UPDATE patients SET beneficiary_id=UPPER(beneficiary_id);`
- No way for "bad" MBIs to get in DB.
  - Added UPPER function to MBI field.

### DIFF
--- a/dpc-common/src/main/java/gov/cms/dpc/common/entities/PatientEntity.java
+++ b/dpc-common/src/main/java/gov/cms/dpc/common/entities/PatientEntity.java
@@ -1,6 +1,7 @@
 package gov.cms.dpc.common.entities;
 
 import gov.cms.dpc.common.annotations.NoHtml;
+import org.hibernate.annotations.ColumnTransformer;
 import org.hibernate.validator.constraints.NotEmpty;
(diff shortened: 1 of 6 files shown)


dpc-common/src/main/java/gov/cms/dpc/common/entities/PatientEntity.java
```diff
@@ -118,6 +115,14 @@ public PatientEntity update(PatientEntity updated) {
         return this;
     }
 
+    @PrePersist
+    @PreUpdate
+    public void upperCaseBeneficiaryId() {
```
whuang85, line 120: Can you assert the uppercase in either the smoke test or e2e test to make sure we aren't running into any weird dropwizard transaction problems?

SalGutierrez: I had added an integration test, but also added a test to the e2e collection.

PR #1017, opened 2020-08-31
DPC-503 fix: html validator to allow `&`

jonfulk:
### Fixes [DPC-503](https://jira.cms.gov/browse/DPC-503)

When an organization name or address contained `&`, it failed validation due to a check to ensure that there
is no HTML in the form data.

### Proposed Changes

- if the String being checked for HTML contains an ampersand surrounded by spaces, the ampersand is replaced by an empty space, thereby ignoring it
- a unit test was updated to check for such a condition

### Change Details

I opted for stripping out the ` & ` because the library we are using to check for HTML (jsoup) allows whitelisting HTML tags, but not necessarily strings like this. It is important to include the spaces before and after the `&` to ensure that we are only ignoring ampersands used to mean `and` while not ignoring other codes that begin with an ampersand. The unit test includes both use cases.

### Security Implications

### Acceptance Validation

In addition to running all of the automated tests, which passed, I ran the web app locally and updated an organization to include an `&` in both the name as well as the address. These changes did not throw any errors.

Screen Shot 2020-08-31 at 2 45 13 PM
Screen Shot 2020-08-31 at 2 54 30 PM

### Feedback Requested

I would've preferred to update the jsoup Whitelist as that seemed to be the intended purpose of the Whitelist, however, after reading some docs and trying some variations, I was unable to get it to work. I am open to any ideas about how to approach this differently.


dpc-common/src/main/java/gov/cms/dpc/common/hibernate/validator/NoHtmlValidator.java
```diff
@@ -23,7 +23,9 @@ public boolean isValid(String s, ConstraintValidatorContext constraintValidatorC
         if (StringUtils.isBlank(s)) {
             return true;
         }
-        String safe = Jsoup.clean(s, "", Whitelist.none(), new Document.OutputSettings().prettyPrint(false));
-        return safe.equals(s);
+        // Ignore `&` in names and addresses
+        String s1 = s.replaceAll("(\\s&\\s)", "   ");
```
dhgreene, line 27: Would it make more sense to replace with " and "?
```suggestion
        String s1 = s.replaceAll("(\\s&\\s)", " and ");
```

jonfulk: > Would it make more sense to replace with " and "?

I considered this but thought it might be best to maintain the same character length. No one is actually going to see the modified string since the validator just returns a boolean, so I don't think the replacement matters much. I did go back and forth on it for a bit, though. I'm open to changing it.

whuang85: I think you can just change the Whitelist object to whitelist ampersand.

jonfulk: @whuang85 i tried that, as i mentioned in the description, but the Whitelist object seems to be made for whitelisting HTML tags and not random strings or regular expressions. I would prefer to use the whitelist object if you can help me figure out how to do that.

whuang85: @jonfulk oops, didn't read the description and yea it looks like it's not possible

whuang85 (approved):

PR #1018, opened 2020-09-02
DPC-54: Open redirects

Sun-Mountain:
### Fixes [DPC-54](https://jira.cms.gov/browse/DPC-54)

Answering for the pen tests finding potential weaknesses is the rails app.

[Details in keybase](keybase://team/bcda/ACT/Pen%20Test/2020%20Pen%20Test/BCDA_DPC_Open_Redirect_Via_Referrer_Header_Parameter_AR.docx)

### Proposed Changes

- Checked to make sure any `redirect` was associated with a designated path 
- Change `redirect_back` on the taggings controller

### Security Implications

### Feedback Requested

In the pen-test document in keybase, they found that `Anywhere on the site that contains a 'Cancel' button will redirect the victim to an attacker-controlled site.`

This is what a Cancel link looks like:

```
<%= link_to "Cancel", root_path, class: "ds-c-button ds-c-button--transparent" %>
```

I'm not sure how that's different from a link like this:

```
<%= link_to "Edit", edit_organization_url(organization), class: "ds-c-button ds-c-button--outline ds-c-button--small", data: { test: "edit-link" } %>
```

After some brainstorming I thought about doing something like this:

Change links to buttons:
```
= button_tag "Cancel", :type => 'button'
```

Then in the controller(s):
```
before_filter :check_for_cancel

def check_for_cancel
  if params[:commit] == "Cancel"
    redirect_to my_page_path
  end
end
```

Two questions:
1. Is this more secure than what we have now?
2. If yes, does that mean I have to do this for most links?


dhgreene (approved):
👍

PR #1019, opened 2020-09-04
DPC-613: Build a basic performance test suite for attribution

embh:
### Fixes [DPC-613](https://jira.cms.gov/browse/DPC-613)

In order to measure our current performance, and prioritize and measure improvements, we need a performance and load test suite for our API.

### Proposed Changes

Begin setting up a performance and load testing suite.

### Change Details

* Sets up a basic test suite using k6, including:
  * Two JavaScript files for testing the `/metadata` endpoint: one for performance, one for load
  * Shared `/modules/config.js` for configuring environment URL
  * Shell script for running both JS tests via Docker
  * `README.md` with instructions to run with or without Docker

### Security Implications
This is a test suite and makes no changes to our application code.

### Acceptance Validation

See README for instructions for running the tests. Sample output below.

Screen Shot 2020-09-04 at 1 38 58 PM
Screen Shot 2020-09-04 at 1 39 10 PM
Screen Shot 2020-09-04 at 1 39 23 PM

### Feedback Requested
We can of course make adjustments as our test suite matures, but wondering about things like:
* Should the performance tests and load tests run from separate shell scripts?
* How should we organize the JS files?
* Would it be helpful to have output somewhere other than stdout?

### DIFF
--- a/dpc-testing/performance/README.md
+++ b/dpc-testing/performance/README.md
@@ -1,0 +1,16 @@
+# Performance and Load Tests
+
+DPC performance and load tests use [k6](https://k6.io/).
+
+## Running the tests
+### Shell script
+The performance and load tests may be run using the script `dpc-perf-load-test.sh`. This will run k6 with Docker. Specify the environment you wish to test by setting the environment variable `ENV`. The options are `docker`, `dev`, `test`, `sandbox`, and `prod`.
+```
+ENV=docker ./dpc-perf-load-test.sh
+```
+
+### With k6
+After installing k6, you can run the tests directly. Specify the environment with the flag `-e env={environment}`, and the test script file.
+```
+k6 run -e env=local scripts/get_metadata-perf.js
+```
--- a/dpc-testing/performance/README.md
+++ b/dpc-testing/performance/README.md
@@ -10,7 +10,7 @@ ENV=docker ./dpc-perf-load-test.sh
 ```
 
 ### With k6
-After installing k6, you can run the tests directly. Specify the environment with the flag `-e env={environment}`, and the test script file.
+After installing k6, you can run the tests directly. Specify the environment with the flag `-e env={environment}`, and the test script file. The environment options are `local`, `dev`, `test`, `sandbox`, and `prod`.
 ```
 k6 run -e env=local scripts/get_metadata-perf.js
 ```
--- a/dpc-testing/performance/dpc-perf-load-test.sh
+++ b/dpc-testing/performance/dpc-perf-load-test.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+docker pull loadimpact/k6
+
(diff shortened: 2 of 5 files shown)


dpc-testing/performance/README.md
```diff
@@ -0,0 +1,16 @@
+# Performance and Load Tests
+
+DPC performance and load tests use [k6](https://k6.io/).
+
+## Running the tests
+### Shell script
+The performance and load tests may be run using the script `dpc-perf-load-test.sh`. This will run k6 with Docker. Specify the environment you wish to test by setting the environment variable `ENV`. The options are `docker`, `dev`, `test`, `sandbox`, and `prod`.
```
jonfulk, line 7: This is a pretty picky comment, so feel free to ignore. The name of that script is pretty long. I wonder if `dpc-load-test.sh` would be descriptive enough? I suppose we could eventually create a `make` command for it as well.

jonfulk: is `local` also an option for the shell script, or only with k6?

dpc-testing/performance/README.md
```diff
@@ -0,0 +1,16 @@
...
+```
+ENV=docker ./dpc-perf-load-test.sh
+```
+
+### With k6
+After installing k6, you can run the tests directly. Specify the environment with the flag `-e env={environment}`, and the test script file. The environment options are `local`, `dev`, `test`, `sandbox`, and `prod`.
```
jonfulk, line 13: is `docker` also an option here?

jonfulk (approved):
Seems pretty clear and simple to use. I like it!

dpc-testing/performance/README.md
embh: Great question! I was having some trouble with the naming. `local` will use `localhost:3002`, which won't work when running inside Docker. That's why I added the `docker` option, which uses `host.docker.internal:3002`. Would that make more sense as `local-docker` or another name?

dpc-testing/performance/README.md
embh: (Answered above: https://github.com/CMSgov/dpc-app/pull/1019#discussion_r483632338)

dpc-testing/performance/README.md
jonfulk: That makes sense. I was mainly confused because you don't list `local` in the env options above, and you don't list `docker` in the env options below even though `local` and `docker` are both listed in `config.js`

embh: Regarding the script name, that's another open question. The ticket requested performance _and_ load tests, but I don't know if the right approach is to put them in one script or split them up. I wanted to be clear that the shell script was doing both, but if we split them up, that solves the long name concern. 🤔

dhgreene (approved):
Great first step!  Regarding output: I think it would be good to integrate with [New Relic](https://k6.io/docs/results-visualization/new-relic) in a later ticket, but stdout is perfect for now.

PR #1021, opened 2020-09-04
DPC 466  Feature: Handle encrypted Redis connection

SMLuthi:
### Fixes [DPC-466](https://jira.cms.gov/browse/DPC-466)

Changing our redis database over to have at-rest encryption resulted in the web app (and the sidekiq job processor) to fail in its connection.

### Proposed Changes

The redis connection needs to be updated to allow for `ssl/tls`
- This PR corresponds with : https://github.com/CMSgov/dpc-ops/pull/200

### Change Details

All redis connection uris are switched over to include the `ssl true option`... in our case that means just changing the uri host string to be `rediss` instead of `redis`.

Also, the `hiredis` gem had to be removed. Its a wrapper for the `redis` gem that increases performance. It, however, was incompatible with `redis` gems ssl/tls features.

Lastly, the built in `Redis.current` that the `redis-rb` provides to grab the current stable connection of redis does not work in its default state anymore (most likely because its not setup to allow fr ssl/tls). An initializer has been provided that now sets up the current correctly (which also led to changing the `REDIS_URL` env var to include the redis url prefix... ie. `rediss://...`

### Security Implications


dhgreene (approved):
🔒

jaddy2726 (approved):

PR #1036, opened 2020-09-11
DPC-667: fix breaking smoke tests

jonfulk:
### Fixes [DPC-667](https://jira.cms.gov/browse/DPC-667)

Dropwizard-jackson also depends on `jackson-databind-jsr310` (v.2.9.10), and it appears that overriding this module to v.2.11.2 in `dpc-smoketest` without also upgrading `jackson-core` and `jackson-annotations` to that same version caused the following error:
```java.lang.ClassNotFoundException: com.fasterxml.jackson.databind.ser.std.ToStringSerializerBase```

### Proposed Changes
- drop `jackson-datatype-jsr310` dependency from `dpc-smoketest` pom file

### Change Details
To avoid causing more issues with `dropwizard-jackson` by upgrading other related jackson modules, drop the override and stick with the version they depend on. Note that `jackson-databind-jsr310` is currently deprecated. FasterXML recommends replacing with `jackson-modules-java8` instead. Although, that may cause other conflicts as long as we are using `dropwizard-jackson`.

### Security Implications
None

### Acceptance Validation
Smoke tests pass without the specified dependency. Also successfully ran `make ci-app` for further validation.

Screen Shot 2020-09-11 at 4 00 29 PM
Screen Shot 2020-09-11 at 4 04 46 PM

### Feedback Requested
Any


embh (approved):
LGTM! Could you update the title of the PR before merging so the ticket number is in the regular JIRA format?

PR #1041, opened 2020-09-14
DPC-19: Fix error to save updated registered org if missing required field

Sun-Mountain:
### Fixes [DPC-19](https://jira.cms.gov/browse/DPC-19)

If the internal user attempts to edit a registered organization and deletes a required field (i.e. the NPI) return an error and render the edit page again.

### Proposed Changes

- Add check in organization controller to see if `npi.nil?` if trying to enable org.
    - Responds with appropriate error for user
- Add check in registered organization controller to see if enabled and `npi.nil?` to not update the organization or registered organization
- Tests for these functions

### Security Implications

### Acceptance Validation

Screen Shot 2020-09-15 at 4 27 49 PM


dpc-web/app/controllers/internal/organizations_controller.rb
```diff
@@ -65,7 +65,10 @@ def edit
     def update
       @organization = Organization.find id_param
 
-      if @organization.update organization_params
+      if organization_enabled?(@organization) && npi_blank?
+        flash[:alert] = 'Enabled organization require an NPI.'
```
dhgreene, line 69:
```suggestion
        flash[:alert] = 'Enabled organizations require an NPI.'
```

dpc-web/app/controllers/internal/organizations_controller.rb
```diff
@@ -119,6 +122,16 @@ def delete_user_from_org
       end
     end
 
+    def npi_blank?
```
SMLuthi, line 125: This works, but I wonder if you could use rails `require` to enforce the parameter.

```
    def organization_params
      params.require(:organization).permit(
        :name, :organization_type, :num_providers, :npi, :vendor,
        address_attributes: %i[id street street_2 city state zip address_use address_type]
      ).tap do |org_params|
        org_params.require(:npi)
      end
    end
```

That may just force an error automatically

SMLuthi (dismissed):
I think this looks good. I gave a suggestion that may (or may not) work.

dpc-web/app/controllers/internal/organizations_controller.rb
Sun-Mountain: Can you also specify an error this way?

The reason I went with this method was so I could throw `flash[:alert] = 'Enabled organization require an NPI.'`

Sun-Mountain: Also this requires an npi on creation, which we may not know immediately.

dhgreene (approved):
🐎

PR #1044, opened 2020-09-15
DPC-573: Allow label creation from web app / Accept token label in body.

SalGutierrez:
### Fixes [DPC-573](https://jira.cms.gov/browse/DPC-573)

DPC Web was sending the label as a json body when creating a token, API required them as a query param.

### Proposed Changes
Instead of simply sending the request with query params it was decided to accept them in the body.
To maintain backwards compatibility query params are still accepted but should be deprecated.

### Change Details
- accept token label and expiration in body
- created model for request body
- updated integrations tests to run independently from each other (Order mattered before, now they may run in any order)
- Added more tests
- extended the formats allowed for expiration time (Kept the ability to accept ISO while also accepting FHIR)

### Security Implications
Don't see any security issues.

### Acceptance Validation

Yes, tested in Dev

Screen Shot 2020-09-16 at 11 30 03 PM
Screen Shot 2020-09-16 at 11 30 59 PM
Screen Shot 2020-09-16 at 11 31 27 PM
Screen Shot 2020-09-16 at 11 36 19 PM

### Feedback Requested

Missed anyting?


dpc-api/src/main/java/gov/cms/dpc/api/models/TokenResponse.java
```diff
@@ -1,83 +0,0 @@
-package gov.cms.dpc.api.models;
```
SalGutierrez, line 1: This file had been abandoned, deleted it since it was not being used.

dhgreene (approved):
🏷️

whuang85 (approved):

PR #1057, opened 2020-09-21
DPC-731: Basic performance test setup for /Key endpoints

embh:
### Fixes [DPC-731](https://jira.cms.gov/browse/DPC-731)

DPC is in the process of adding a basic performance test suite.

### Proposed Changes

This PR adds the API /Key endpoints to the performance test suite.

### Change Details

* POST /Key
* GET /Key
* GET /Key/{id}
* DELETE /Key/{id}

There is a follow-up ticket to address a rare public key and signature mismatch in these tests. The tests will proceed if this occurs, and the 400 response will be in the metrics written to stdout.

### Security Implications

This contains only testing code. No security impacts.

### Acceptance Validation

Screen Shot 2020-09-21 at 2 27 39 PM

### Feedback Requested

Any


dpc-testing/performance/main.go
```diff
@@ -46,6 +49,71 @@ func testMetadata() {
...
+
+		pubKeyStr, _, signature := generateKeyPairAndSignature()
+		body := fmt.Sprintf("{ \"key\": \"%s\", \"signature\": \"%s\"}", pubKeyStr, signature)
+		t.Body = []byte(body)
+
+		return nil
+	}
+}
+
+func newDELETEKeyTargeter(accessToken string, nextKeyID func() string) vegeta.Targeter {
```
dhgreene, line 106: I like the nextKeyID func!

dpc-testing/performance/setup.go
```diff
@@ -91,7 +91,7 @@ func getKeyPairAndSignature() (string, *rsa.PrivateKey, string) {
 		cleanAndPanic(err)
 	}
 
-	signature, err := privKey.Sign(rand.Reader, snippetHash.Sum(nil), crypto.SHA256)
+	signature, err := rsa.SignPKCS1v15(rand.Reader, privKey, crypto.SHA256, snippetHash.Sum(nil))
```
dhgreene, line 94: Is this change an artifact from troubleshooting the intermittent signature bug, or is it helpful in another way?

embh: It is indeed the result of troubleshooting. Turns out `rsa.SignPKCS1v15()` is what `PrivateKey.Sign()` calls if you don't specify, so it didn't make any difference, but I figured it made sense to call it directly. https://golang.org/src/crypto/rsa/rsa.go?s=4735:4794#L136

dhgreene (approved):
👍

PR #1070, opened 2020-09-29
DPC-643: fix sonar vulnerabilities in `dpc.testing.APIAuthHelpers`

jonfulk:
### Fixes [DPC-643](https://jira.cms.gov/browse/DPC-643)

Sonar scans picked up a few security vulnerabilities in the testing API Auth Helpers class that needed to be fixed.

### Proposed Changes
- Instead of always returning `true` when verifying the host name, check if the name matches the SSL session's peer host
- Specify `TLSv1.2` as suggested by SonarQube
- Add the possibility of throwing a `CertificateException` as required by the overridden methods of `X509TrustManager` See details below for more on this change.

### Change Details

The first two changes above were suggested by SonarQube, meet the guidelines, and still allow the smoke tests to pass. For the `CertificateException` change, however, this is a bit of a workaround. SonarQube wants the `X509TrustManager` to throw an exception when there is a failed certificate check, which is completely reasonable outside of testing. However, since our smoke tests do not provide any actual X509Certificates to check, these checks are not really happening. Previously, the methods were no-ops. With this change, they remain no-ops when there are no certificates being passed to the check methods, otherwise they will throw a `CertificateException`, which does not change the functionality but appeases SonarQube because the methods can now throw exceptions.

### Security Implications

Note: the SSL certificate check has been altered for testing, but it is no less secure than before as nothing was being checked previously.

### Acceptance Validation
Screen Shot 2020-09-29 at 1 04 03 PM
Screen Shot 2020-09-29 at 1 03 46 PM
Screen Shot 2020-09-29 at 10 28 04 AM

### Feedback Requested
While this change does not make the existing certificate check any less secure than it already was, it does raise the question whether we should be doing something differently here. This AuthHelper is only being used for testing. Is there any concern about the way we are essentially not checking the ssl certificate at all? Should we be passing certificates in our tests?

### DIFF
--- a/dpc-testing/src/main/java/gov/cms/dpc/testing/APIAuthHelpers.java
+++ b/dpc-testing/src/main/java/gov/cms/dpc/testing/APIAuthHelpers.java
@@ -293,22 +293,14 @@ public class APIAuthHelpers {
 
             @Override
             public void checkClientTrusted(X509Certificate[] certs, String authType) throws CertificateException {
-                if (certs.length == 0) {
-                    // Do nothing because this is only used for smoke tests
-                }
-                else {
-                    throw new CertificateException();
-                }
+                // only used for testing, so no certificates expected
+                if (certs.length != 0) { throw new CertificateException(); }
             }
 
             @Override
             public void checkServerTrusted(X509Certificate[] certs, String authType) throws CertificateException {
-                if (certs.length == 0) {
-                    // Do nothing because this is only used for smoke tests
-                }
-                else {
-                    throw new CertificateException();
-                }
+                // only used for testing, so no certificates expected
+                if (certs.length != 0) { throw new CertificateException(); }
             }
 
         }};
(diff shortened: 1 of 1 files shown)


dpc-testing/src/main/java/gov/cms/dpc/testing/APIAuthHelpers.java
```diff
@@ -291,13 +292,23 @@ private static SSLContext createTrustingSSLContext() throws KeyManagementExcepti
...
             @Override
-            public void checkClientTrusted(X509Certificate[] certs, String authType) {
-                // Do nothing
+            public void checkClientTrusted(X509Certificate[] certs, String authType) throws CertificateException {
+                if (certs.length == 0) {
+                    // Do nothing because this is only used for smoke tests
+                }
+                else {
+                    throw new CertificateException();
+                }
```
embh, line 301: Would `if (certs.length != 0) { throw new CertificateException(); }` be enough to make Sonar happy?

jonfulk: Could be! I'll try that out.

embh (approved):
LGTM!

I'd like to get @adhocEric 's thoughts on the questions raised in the Feedback Requested section. Should we create tickets?

jonfulk:
> LGTM!
> 
> I'd like to get @adhocEric 's thoughts on the questions raised in the Feedback Requested section. Should we create tickets?

Good idea. I'll go ahead and merge this (after running a build on dev to be sure it still passes the sonarqube checks) and will create a ticket for the other questions. Thanks!

PR #1074, opened 2020-09-30
DPC 724 - Added feature flags to Organization

SalGutierrez:
### Fixes [DPC-724](https://jira.cms.gov/browse/DPC-724)

During pilot we expect to enforce per-organization limits (on the number of patients per practitioner, or the total number of patients per practice, for instance). For testing purposes, we would also like per-request feature flags (to disable lookback, for instance).

In order to implement per-requests feature flags we first have to implement organization level feature flags that will allow us to send flags in each request.

If an org has Feature X then we will allow them to send in per requests flags (via headers).

### Proposed Changes

The first step (That this ticket covers) is to allow the creation and retrieval of organization level feature flags.
### Change Details

- Added new column of type JSON called "features" to attribuiton.Organization table
- Added Hibernate libraries that will allow us to work with the Json type
- Added features attribute to OrganizationEntity
- Added integration tests to prove feature flags are being stored and retrieved correctly.
- Kept structure simple; however, since we are using JSON type we can extend it if we need more complexity.

### Security Implications

`        <dependency>
            <groupId>com.vladmihalcea</groupId>
            <artifactId>hibernate-types-52</artifactId>
            <version>2.9.13</version>
        </dependency>`

In progress: https://confluence.cms.gov/pages/viewpage.action?pageId=367790846

### Acceptance Validation

- Tested locally with locally 
- Migration worked successfully
- Need to test migration on DEV

Screen Shot 2020-09-29

### Feedback Requested
Naming suggestions?
Structure changes?

### DIFF
--- a/dpc-common/src/main/java/gov/cms/dpc/common/entities/FeatureFlags.java
+++ b/dpc-common/src/main/java/gov/cms/dpc/common/entities/FeatureFlags.java
@@ -9,12 +9,10 @@ import java.util.Optional;
 public class FeatureFlags {
     private ObjectNode featuresRootNode;
 
-    public FeatureFlags(){
-        featuresRootNode = JsonNodeFactory.instance.objectNode();
-    }
+    public FeatureFlags(){}
 
-    public Optional<JsonNode> getFeaturesRootNode() {
-        return Optional.ofNullable(featuresRootNode);
+    public ObjectNode getFeaturesRootNode() {
+        return featuresRootNode;
     }
 
     public void setFeaturesRootNode(ObjectNode featuresRootNode) {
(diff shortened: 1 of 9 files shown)


dpc-common/src/main/java/gov/cms/dpc/common/entities/OrganizationEntity.java
```diff
@@ -132,6 +140,14 @@ public void setRosters(List<RosterEntity> rosters) {
         this.rosters = rosters;
     }
 
+    public FeatureFlags getFeatures() {
```
jonfulk, line 143: it might be more clear to call this and related methods `getFeatureFlags`

dpc-common/src/main/java/gov/cms/dpc/common/entities/OrganizationEntity.java
```diff
@@ -132,6 +140,14 @@ public void setRosters(List<RosterEntity> rosters) {
...
+        return features;
+    }
+
+    public void setFeatures(FeatureFlags features) {
```
jonfulk, line 147: i think `featureFlags` might be more clear than `features`

jonfulk (dismissed):
A couple naming suggestions (take em or leave em), but otherwise looks good!

dpc-common/src/main/java/gov/cms/dpc/common/entities/FeatureFlags.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public FeatureFlags(){
+        featuresRootNode = JsonNodeFactory.instance.objectNode();
+    }
+
+    public Optional<JsonNode> getFeaturesRootNode() {
+        return Optional.ofNullable(featuresRootNode);
+    }
+
+    public void setFeaturesRootNode(ObjectNode featuresRootNode) {
```
SalGutierrez, line 20: Woops, just realized I added a setter and could cause a NPE

dpc-common/src/main/java/gov/cms/dpc/common/entities/OrganizationEntity.java
SalGutierrez: done.

dpc-common/src/main/java/gov/cms/dpc/common/entities/FeatureFlags.java
SalGutierrez: fixed.

jonfulk (approved):
LGTM!

PR #1079, opened 2020-10-02
DPC-643 fix: modify cert check to avoid smoke test failures

jonfulk:
### Fixes [DPC-643](https://jira.cms.gov/browse/DPC-643)

Previous fix caused the smoke tests to fail on `test` and then later on `dev`.

### Proposed Changes

- essentially don't check certificates but add in the unreachable possibility of throwing an error so that Sonarqube approves

### Change Details

Since it is still unclear what certificates, if any, are being passed around in what environments during the smoke tests, the present fix is to do nothing regardless the number of certificates. The `CertificateException` should never be thrown, but it is currently there to make Sonarqube happy. This is really not doing anything different than the original empty methods.

### Security Implications

A ticket exists already to look into whether we should be handling certificates differently during testing.

### Acceptance Validation

- deployed to dev and smoke tests passed
- ran smoke tests locally and they passed
- built the docker image and sonarqube passed

### Feedback Requested
Any


dhgreene (approved):
👍

PR #1081, opened 2020-10-04
Silvio/dpc 697 public keys controller tests

SMLuthi:
### Fixes [DPC-697](https://jira.cms.gov/browse/DPC-xxx)

There are currently not tests written for the `public_keys_controller` and it was, therefore, failing the simplecov coverage restrictions that we put in place.

### Proposed Changes

Add tests to fully cover the controller.

### Change Details

Added tests to fully cover all methods in the controller.

### Security Implications

### Feedback Requested

Note: There was an issue were we were checking for `public_key` size. However, this actually needed to check for the label size (to avoid making an unnecessary call to the API if the label size is greater than 25). This has now been fixed and is working correctly.

### DIFF
--- a/dpc-web/app/controllers/public_keys_controller.rb
+++ b/dpc-web/app/controllers/public_keys_controller.rb
@@ -46,25 +46,21 @@ class PublicKeysController < ApplicationController
 
   private
 
-  # :nocov:
   def render_error(msg)
     flash[:alert] = msg
     render :new
   end
-  # :nocov:
 
   def missing_params
     params[:public_key].blank?
   end
 
   def public_key_length
-    params[:public_key].length <= 25
+    params[:public_key].length >= 25
   end
 
-  # :nocov:
   def unauthorized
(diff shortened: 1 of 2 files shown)


SMLuthi:
@Sun-Mountain Hey, I wanted to verify something with you. Theres a a check in the controller that verifies that the public key length is less than or equal to 25 characters. However, the flash alert says `Public key cannot be over 25 characters` if that checks returns true... So I changed the check to check if the key length is greater than or equal to 25 (so the flash message is now correct). Let me know if this is now as intended or if I got it backwards. Thanks!

Sun-Mountain:
For documentation's sake. The current app works with `params[:public_key].length <= 25` or `params[:public_key].length > 25`, catching a label that's larger than 25 characters. For ease of understanding we should change it to the latter (removing the equals, because it's okay if it's 25 characters, just not over.

Sun-Mountain (approved):
💯

PR #923, opened 2020-07-22
DPC-554 - add check to make sure Accept header is set to application/fhir+json

richbraman:
… Accept  header is not null or empty string, and a check to make sure its is set to application/fhir+json

### Fixes [DPC-554](https://jira.cms.gov/browse/DPC-554)

FHIR Bulk data spec requires the Accept parameter to be set to application/fhir+json

### Proposed Changes
add check to make sure Accept header is set to application/fhir+json
Also added swagger annotation to make sure Accept header is required and set to Application/fhir+json 

### Change Details
no further details.

### Security Implications

No; n/A
- [ ] new software dependencies
None

- [ ] security controls or supporting software altered
No

- [ ] new data stored or transmitted
No

- [ ] security checklist is completed for this change
n/A

- [ ] requires more information or team discussion to evaluate security implications
No

- [ X] no PHI/PII is affected by this change

### Acceptance Validation

Yes, via postman

### Feedback Requested

feedback on annotations

PR #1091, opened 2020-10-16
DPC-554 fix: Require `Accept` header

jonfulk:
### Fixes [DPC-554](https://jira.cms.gov/browse/DPC-554)

This PR is meant to replace https://github.com/CMSgov/dpc-app/pull/923

FHIR Bulk data spec requires the Accept parameter to be set to application/fhir+json

### Proposed Changes

- Add a unit test to check that we are correctly checking the required headers
- update header variable names to be lowercase, per convention
- split out and modify the `checkRequestHeader` method to take a `Map` of headers instead of multiple string variables

### Change Details

The decision to use a`Map` instead of a list of variables is to make the method more extensible for the future. It is better to change the implementation of the method if needed in the future instead of the interface, so a `Map` is better than a growing list of `String` parameters.

### Security Implications

### Acceptance Validation

- New unit tests pass
- smoke tests passed locally
Screen Shot 2020-10-19 at 12 54 40 PM

### Feedback Requested

Any


dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
```diff
@@ -268,15 +273,25 @@ public Response export(@ApiParam(hidden = true)
...
-                           @ApiParam(hidden = true) @HeaderParam("Prefer")  @Valid String Prefer) {
+                           @ApiParam(hidden = true) @HeaderParam("Prefer") @Valid String prefer,
+                           @ApiParam(hidden = true) @HeaderParam("Accept") @Valid String accept) {
         logger.info("Exporting data for provider: {} _since: {}", rosterID, since);
 
         // Check the parameters
-        checkExportRequest(outputFormat, Prefer);
+        Map<String, String> headers = Stream.of(Pair.of("Prefer", prefer), Pair.of("Accept", accept))
+                .peek(header -> {
+                    if (header.getRight() == null) {
```
whuang85, line 287: I feel like this check should also be in the checkRequestHeaders method

jonfulk: Initially it was, but it needs to be here in order to get rid of any `null` values before creating the `Map` since the `Map` does not allow a `null` to be set as a value. The point of this line of code is to build up a map of headers that can be passed down to the verification methods instead of multiple strings.

whuang85: ah I see.. maybe filter out the nulls instead and then do a check on an empty map or missing key in the `checkRequestHeaders`?  Sorry, just seeing if there is a way to contain the BadRequestException to the check methods.

jonfulk: I appreciate the suggestions. I'll look into that. I considered that approach as well, but I ran into the issue of losing track of which header value was for which header in the `Map`, which was necessary in order to give the proper error message.

SalGutierrez: You can add @DefaultValue("") to each header and that will allow you not worry nulls, but would still need to build the map.

It looks like there are only two headers at the moment, so we could just pass those two as arguments without a map.

Another option that we can use if we need more headers is to use `@Context HttpHeaders headers`
this will give you a nice way to manage multiple headers without having to explicitly specify each header in the method signature.

whuang85 (dismissed):

whuang85 (approved):

SalGutierrez (approved):

PR #1092, opened 2020-10-19
Silvio/dpc 710 attr performance test for orgs

SMLuthi:
### Fixes [DPC-710](https://jira.cms.gov/browse/DPC-710)

This PR has turned into a major refactor for the Go performance/load tests:

### Change Details

1. Added `/Organization` tests
2. Added `/Practitioners` tests
3. Created new package levels for `dpc` and `targeter` to diversify the file structure to make it more digestible.
4. The targeter package is now a self-contained generic targeter builder.
5. `api.go` contains all functionality having to do with the api outside of a testing context (i.e. creating tokens, refreshing access, etc...)
6. `dpc.go` contains all common helper functions/
7. All test files are prefixed with `test_*.go` to identify them.
8. Now that `main.go` has been cleaned up it can be simply run as `go run main.go`

### Security Implications


dpc-testing/performance/organization.go
```diff
@@ -0,0 +1,175 @@
...
+	go func() {
+		for {
+			for _, body := range bodies {
+				o.bodies <- body
+			}
+		}
+	}()
+}
+
+// builtTarget is the function passed to the vegeta test runner. It expects a
```
embh, line 69:
```suggestion
// buildTarget is the function passed to the vegeta test runner. It expects a
```

SMLuthi:
It says Travis build failed but they actually passed.

SMLuthi:
Please ignore the codeclimate failures... One of them is for a test file being 51/50 lines long (but this length should be irrelevant for test purposed) and the second failure is an odd request to refactor code because its importing the same thing in two different files... very strange.

dpc-testing/performance/main.go
```diff
@@ -1,99 +1,27 @@
...
-func cleanAndPanic(err error) {
-	cleanUp()
-	panic(err)
+	var api dpc.API
+
+	flag.StringVar(&api.URL, "api_url", "http://localhost:3002/v1", "Base URL of API")
+	flag.StringVar(&api.AdminURL, "admin_url", "http://localhost:9903/tasks", "Base URL of admin tasks")
+	flag.Parse()
+
+	api.CreateGoldenMacaroon()
```
embh, line 16: I think there's a risk of confusion here because the golden macaroon comes from an admin endpoint rather than an API endpoint. Maybe better in `dpc` or an `admin` package?

dpc-testing/performance/pkg/dpc/api.go
```diff
@@ -121,34 +121,42 @@ func uploadKey(key, sig, orgID string) string {
...
+	orgID       string
+	accessToken string
+	keyID       string
+	privateKey  *rsa.PrivateKey
+}
 
-	err = os.RemoveAll("tokens")
-	if err != nil {
-		fmt.Println("tokens directory could not be deleted", err)
+func (api *API) SetupOrgAuth(orgIDs ...string) orgAuth {
```
embh, line 144:
```suggestion
func (api *API) SetUpOrgAuth(orgIDs ...string) orgAuth {
```

dpc-testing/performance/pkg/dpc/dpc.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return accessTokens
+}
+
+var FHIR = "application/fhir+json"
+var JSON = "application/json"
+var PLAIN = "text/plain"
+var FORM = "application/x-www-form-urlencoded"
+var UNSET = ""
```
embh, line 61: https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

dpc-testing/performance/pkg/dpc/test_key.go
```diff
@@ -0,0 +1,54 @@
...
+	defer api.DeleteOrg(auth.orgID)
+
+	// POST /Key
+	resps := targeter.New(targeter.Config{
+		Method:      "POST",
+		BaseURL:     api.URL,
+		Endpoint:    ENDPOINT,
+		AccessToken: auth.accessToken,
+		Bodies:      generateKeyBodies(25, api.GenerateKeyPairAndSignature),
+		Headers:     Headers(JSON, UNSET),
```
embh, line 21: Could `Headers()` be changed to take a variable number of parameters so an unset one isn't required?

SMLuthi: Good question. The reason I settled on this is because, currently, most of the tests settle on default values of fhir json. This way resulted in the least amount of clutter in this field. I think, if/when more tests are included with more header parameters this will probably get reworked a bit into something thats not as minimal but more expandable... if that makes sense.

dpc-testing/performance/pkg/dpc/dpc.go
SMLuthi: Had no idea, will rename!

dpc-testing/performance/main.go
SMLuthi: I see what you mean. I kept it that way since the `GetClientToken` func was pulled directly from your code and combines both calls. But I can break them out.

Sun-Mountain (approved):
LGTM! Don't if @dhgreene wants to look at this first.

dhgreene (approved):
This works fine in `dev` and `test`.  Bonus points for a useful refactoring. 💯

PR #1097, opened 2020-10-21
DPC-814 fix: Allow multiple possible values for `Accept` header

jonfulk:
### Fixes [DPC-814](https://jira.cms.gov/browse/DPC-814)

- After adding a method to check the value of the `Accept` header, smoke tests began failing
- In smoke tests, the `Accept` header is automatically added by the HAPI-FHIR client
- The smoke tests were failing because we were checking explicitly for the value of the header to be `application/fhir+json`, which is the current standard [as shown here](https://hl7.org/fhir/STU3/http.html#mime-type), but the client is sending `application/fhir+json;q=1.0, application/json+fhir;q=0.9`

### Proposed Changes

- add more test cases to the unit test
- create a list of possible acceptable values for the `Accept` header to make it backwards compatible as the HAPI-FHIR client does
- check the contents of the `Accept` header against this list of acceptable header values

### Security Implications

### Acceptance Validation
- Smoke tests now passing locally
Screen Shot 2020-10-21 at 5 03 46 PM

### Feedback Requested
I don't love iterating over a String array inside of a stream. Is there a more straightforward way to do this that is more performant?

### DIFF
--- a/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
@@ -44,8 +44,7 @@ import java.util.*;
 import java.util.stream.Collectors;
 
 import static gov.cms.dpc.api.APIHelpers.addOrganizationTag;
-import static gov.cms.dpc.fhir.FHIRMediaTypes.FHIR_JSON;
-import static gov.cms.dpc.fhir.FHIRMediaTypes.FHIR_NDJSON;
+import static gov.cms.dpc.fhir.FHIRMediaTypes.*;
 import static gov.cms.dpc.fhir.helpers.FHIRHelpers.handleMethodOutcome;
 
 
@@ -377,7 +376,7 @@ public class GroupResource extends AbstractGroupResource {
     private static void checkRequestHeaders(List<Pair<String, String>> headers) {
         headers.stream()
                 .peek(header -> {
-                    if (header.getRight() == null) {
+                    if (header.getRight() == null || header.getRight().isBlank()) {
                         throw new BadRequestException("The " + header.getLeft() + " header is required");
                     }
                 })
@@ -385,8 +384,13 @@ public class GroupResource extends AbstractGroupResource {
                     if (pair.getLeft().equals("Prefer") && !pair.getRight().equals("respond-async")) {
                         throw new BadRequestException("The 'Prefer' header must be 'respond-async'");
                     }
-                    if (pair.getLeft().equals("Accept") && !pair.getRight().equals(FHIR_JSON)) {
-                        throw new BadRequestException("The 'Accept' header must be " + FHIR_JSON);
+                    if (pair.getLeft().equals("Accept")) {
+                        String[] values = StringUtils.split(pair.getRight(), ",");
+                        for (String value : values) {
+                            if (!ACCEPT_FHIR_JSON_VALUES.contains(value.trim())) {
+                                throw new BadRequestException("The 'Accept' header must be " + FHIR_JSON);
+                            }
+                        }
                     }
                 });
     }
--- a/dpc-common/src/main/java/gov/cms/dpc/fhir/FHIRMediaTypes.java
+++ b/dpc-common/src/main/java/gov/cms/dpc/fhir/FHIRMediaTypes.java
@@ -1,11 +1,17 @@
 package gov.cms.dpc.fhir;
 
 import javax.ws.rs.core.MediaType;
+import java.util.List;
 
 public class FHIRMediaTypes {
 
     public static final String FHIR_JSON = "application/fhir+json";
     public static final String FHIR_NDJSON = "application/fhir+ndjson";
+    public static final List<String> ACCEPT_FHIR_JSON_VALUES = List.of(
+            FHIR_JSON,
+            "application/fhir+json;q=1.0",
+            "application/json+fhir",
+            "application/json+fhir;q=0.9");
 
     private static final MediaType FHIR_JSON_MT = MediaType.valueOf(FHIR_JSON);
     private static final MediaType FHIR_NDJSON_MT = MediaType.valueOf(FHIR_NDJSON);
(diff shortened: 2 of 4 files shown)


dpc-common/src/main/java/gov/cms/dpc/fhir/FHIRMediaTypes.java
```diff
@@ -1,11 +1,17 @@
...
 import javax.ws.rs.core.MediaType;
+import java.util.List;
 
 public class FHIRMediaTypes {
 
     public static final String FHIR_JSON = "application/fhir+json";
     public static final String FHIR_NDJSON = "application/fhir+ndjson";
+    public static final List<String> ACCEPT_FHIR_JSON_VALUES = List.of(
+            FHIR_JSON,
+            "application/fhir+json;q=1.0",
```
dhgreene, line 12: Should we comment our intentions here?

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
```diff
@@ -377,16 +376,21 @@ private static void checkOutputFormat(String outputFormat) {
...
                 .forEach(pair -> {
                     if (pair.getLeft().equals("Prefer") && !pair.getRight().equals("respond-async")) {
                         throw new BadRequestException("The 'Prefer' header must be 'respond-async'");
                     }
-                    if (pair.getLeft().equals("Accept") && !pair.getRight().equals(FHIR_JSON)) {
-                        throw new BadRequestException("The 'Accept' header must be " + FHIR_JSON);
+                    if (pair.getLeft().equals("Accept")) {
+                        String[] values = StringUtils.split(pair.getRight(), ",");
+                        for (String value : values) {
+                            if (!ACCEPT_FHIR_JSON_VALUES.contains(value.trim())) {
```
dhgreene, line 390: Does this throw an error if the first value is not matched, even if we'd be happy with a later one?

dpc-common/src/main/java/gov/cms/dpc/fhir/FHIRMediaTypes.java
whuang85: This is mostly for people who use the hapi fhir client right?

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
whuang85, line 390: You can use StringUtils.equalsAnyIgnoreCase() if you want to check for equality ignoring case but you'll have to convert your `ACCEPT_FHIR_JSON_VALUES` into a String array

dpc-common/src/main/java/gov/cms/dpc/fhir/FHIRMediaTypes.java
jonfulk: Yes it is.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
jonfulk: Yes it would, but I assumed any and all header values should be valid.

SalGutierrez: I have been reading more into this and it looks like we should not assume that ONLY the current accepted values are valid.

In the same document linked (https://hl7.org/fhir/STU3/http.html#mime-type) it mentions that the char set may also be specified in the Accept parameter.

`FHIR uses UTF-8 for all request and response bodies. Since the HTTP specification (section 3.7.1) defines a default character encoding of ISO-8859-1, requests and responses SHALL explicitly set the character encoding to UTF-8 using the charset parameter of the MIME-type in the Content-Type header. Requests MAY also specify this charset parameter in the Accept header and/or use the Accept-Charset header.`

This is an older post, but shows a nice example: https://blog.fire.ly/2013/04/04/content-types-and-character-encodings/

SalGutierrez: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html mentions the use of the q accept-parameter and the semicolon.

so what our current smoke test client is sending `application/fhir+json;q=1.0, application/json+fhir;q=0.9` translates to simple english as "give me application/fhir+json if available, if not give me application/json+fhir" and that makes sense since we want to be backwards compatible in case the server does not yet support applicaton/fhir+json.

According to RFC spec, the Accept header takes in a MIME type in the format TYPE/SUBTYPE and optional parameters for the MIME Type start with a semicolon.

Based on my reading it looks like FHIR specs says that

1. The client is required to send an Accept header, and server must enforce the requirement
2. The server must support application/fhir-json (or application/json-fhir)

That means a client could send the following values for Accept

- application/xml [Rejected]
- application/json [Rejected]
- */* [Accepted, since application/fhir-json would be part of */*]
- application/* [Accepted]
- application/fhir-json;q=.5 [accepted]
- application/fhir-json;q=.2; charset=UTF-8,application/xml;q=1 [Accepted]

So, how do we handle it? 
Well it looks like we only support application/fhir-json and application/json-fhir (Same MIME type, just renamed) and charset UTF-8.

We can split the value of Accept by the comma, this will give us a list of MIME types and their respective optional parameters. We then iterate and must find a mime type that is either: [*/*, application/*, application/fhir+json, application/json+fhir], we can ignore the q parameter since we only serve one MIME type, preference does not matter.

if we do find a MIME type that matches we must then look for the optional charset param, if found validate that it is UTF-8, else move on the list.

SalGutierrez: aaaaaannnd we don't have to re-invent the wheel!

if we grab the HttpHeaders by simply adding `@Context HttpHeaders headers` to the resource method signature we have access to some fancy out of the box parsers. 

```
    private void validateAcceptHeader(){
        List<MediaType> supportedMediaTypes = Lists.newArrayList(new MediaType("application", "fhir+json"), new MediaType("Application", "json+fhir"));
        List<MediaType> acceptableMediaTypes = headers.getAcceptableMediaTypes();

        for(MediaType supportedType : supportedMediaTypes){
            for(MediaType acceptedType : acceptableMediaTypes){
                if(supportedType.isCompatible(acceptedType)) {
                    return;
                }
            }
        }
        throw new WebApplicationException("Enter error message here", Response.Status.NOT_ACCEPTABLE);
    }
```

jonfulk: Thanks for all this info, Sal! I'll give this a shot.

whuang85: @SalGutierrez  Can probably cut down the nested for using streams

jonfulk:
This PR is no longer needed. See https://github.com/CMSgov/dpc-app/pull/1098

PR #1098, opened 2020-10-22
Revert "DPC-554 fix: Require `Accept` header"

jonfulk:
Reverts CMSgov/dpc-app#1091

The `FHIRRequestFilter` is already correctly validating that the `Accept` header is a MediaType of `application/fhir+json`.
The smoke tests were passing this requirement before the more strict requirements of the PR to be reverted here.
The diagnostic tool seems to have avoided passing through the filters.


dhgreene (approved):
👍

jonfulk:
Merging without current Travis build passing since the other Travis build passed, per @dhgreene

PR #1099, opened 2020-10-22
fix: add `aud` to JWT and switch on env

jonfulk:
Same change as https://github.com/CMSgov/dpc-static-site/pull/11

### Fixes [DPC-810](https://jiraent.cms.gov/browse/DPC-810)

The JWT Tool was not adding an `aud` claim to the JWT that was generated.
In addition, the `aud` claim is environment-specific, so there needed to be a way to select which environment to use when generating the JWT.

### Proposed Changes
- added a select for the environment: `Sandbox` or `Production`
- added `aud` claim to JWT generator

### Change Details
The URL in the `aud` claim includes the environment. For example, if the env is `prod-sbx` (Sandbox), the claim will have the value `https://sandbox.dpc.cms.gov/api/v1/Token/auth`. For production, the value should be `https://dpc.cms.gov/api/v1/Token/auth`

### Security Implications

No security implications.

### Acceptance Validation

Tested by using the generator to create JWTs for various environments and then validated those JWTs using Postman.

All tests were successful.

### Feedback Requested
Any, particularly if the select should be styled differently. I would also be interested in feedback on the text added to the tooltip for the environment.

### DIFF
--- a/dpc-web/public/jwt/jwt.html
+++ b/dpc-web/public/jwt/jwt.html
@@ -21,7 +21,7 @@
       var data = {
         "iss": clientToken,
         "sub": clientToken,
-        "aud": "https://" + env + "dpc.cms.gov/api/v1/Token/auth",
+        "aud": "https://" + env + ".dpc.cms.gov/api/v1/Token/auth",
         "exp": Math.round(new Date().getTime() / 1000) + 300,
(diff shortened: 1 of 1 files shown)


dhgreene (approved):
🚀

dhgreene (approved):
🙇

PR #1101, opened 2020-10-23
DPC 570 fix silently failing smoke tests

SalGutierrez:
### Fixes [DPC-570](https://jira.cms.gov/browse/DPC-570)
Since the lookback was failing when using synthetic data we were receiving errors during export requests. As a temporary fix we had disabled checking for errors during exports, the following was commented out.
```
                .peek(jobResponse -> {
                    if (jobResponse.getError().size() > 0)
                        throw new IllegalStateException("Export job completed, but with errors");
                })
```

Now that we are capable of bypassing lookback at an org level we need to bypass lookback during smoke tests and check if there are any errors during export. (specifically checking that we are receiving bfd data.)

### Proposed Changes

- Use ids that are exempt from lookback during smoke tests (instead of randomly generating new UUIDs)
- Uncomment previously mentioned code that checks for errors.

### Change Details
- Used the same exempt orgs ids specified in aggregation config to create test organizations.
- Checked job response for errors. If errors are found tests is failed.
- Updated outdated MBIs found in patient bundle.
- Update docker-compose file and make start-app command so that the real BFD client is used during local testing.
- Removed ability to specify pre-existing org and keys. (I do not see a use case for this.)

### Security Implications

### Acceptance Validation

Tested locally, dev, test, sbx, and prod.

```
                .peek(jobResponse -> {
                    if (jobResponse.getError().size() > 0)
                        throw new IllegalStateException("Export job completed, but with errors");
                })
```

### Feedback Requested

Review logic in error checking.
Review changes related to make start-app (Enable the use of the real BFD client during local smoke testing).
Is there a use case for specifying a pre existing org and its credentials during a smoke tests?


dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/SmokeTest.java
```diff
@@ -102,85 +95,77 @@ public void teardownTest(JavaSamplerContext context) {
...
-                clientToken = FHIRHelpers.registerOrganization(adminClient, ctx.newJsonParser(), organizationID, npi, adminURL);
-                orgRegistrationResult.setSuccessful(true);
-            } catch (Exception e) {
-                orgRegistrationResult.setSuccessful(false);
-                throw new IllegalStateException("Cannot register org", e);
-            } finally {
-                orgRegistrationResult.sampleEnd();
-            }
+        String clientToken = createOrganization(smokeTestResult, adminClient, adminURL);
+        orgWasCreated = true;
```
jonfulk, line 123: Seems like we should check that the created org is what we expect before confirming that it was created, but maybe you do that in `createOrganization`

dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/SmokeTest.java
```diff
@@ -102,85 +95,77 @@ public void teardownTest(JavaSamplerContext context) {
...
-                keyTuple = Pair.of(UUID.fromString(keyID), privateKey);
-            } catch (IOException e) {
-                throw new IllegalArgumentException(String.format("Cannot read private key from: %s", privateKeyPath));
-            }
-        }
         // Create an authenticated and async client (the async part is ignored by other endpoints)
-        final IGenericClient exportClient;
+        final IGenericClient exportClient = APIAuthHelpers.buildAuthenticatedClient(ctx, apiURL, clientToken, keyTuple.getLeft(), keyTuple.getRight(), true, true);
+
+        Pair<Bundle, List<String>> practSubmitResults = submitPractitioners(javaSamplerContext, smokeTestResult, exportClient);
```
jonfulk, line 130: it's long, but i would call this `practitionerSubmitResults` for clarity.

dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/SmokeTest.java
```diff
@@ -102,85 +95,77 @@ public void teardownTest(JavaSamplerContext context) {
...
+    private Pair<UUID, PrivateKey> createPublicKey(String hostParam){
+        try {
+            return APIAuthHelpers.generateAndUploadKey(KEY_ID, organizationID, goldenMacaroon, hostParam);
+        } catch (IOException | URISyntaxException | GeneralSecurityException e) {
+            throw new IllegalStateException("Failed uploading public key", e);
+        }
+    }
+
+    private String createOrganization(SampleResult smokeTestResult, IGenericClient adminClient, String adminURL){
+        logger.info(String.format("Creating organization %s", organizationID));
```
jonfulk, line 150: you removed `String.format()` before and used `{}` for interpolation instead. we should that here too.

dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/SmokeTest.java
```diff
@@ -203,14 +188,16 @@ public SampleResult runTest(JavaSamplerContext javaSamplerContext) {
         } finally {
             practitionerSample.sampleEnd();
         }
-
         smokeTestResult.addSubResult(practitionerSample);
+        return Pair.of(providerBundle, providerNPIs);
```
jonfulk, line 192: we use `practitioner` in some places and `provider` in other places. Are they the same?

jonfulk (approved):
Some nice cleanup in here! I left a few thoughts/questions, but no concerns if you ignore them.

dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/ClientUtils.java
```diff
@@ -242,11 +246,36 @@ private static void jobResponseHandler(CloseableHttpClient client, JobCompletion
...
 
+    private static void verifyOperationOutcomeContents(File outcomeFile){
+        try (BufferedReader reader = Files.newBufferedReader(outcomeFile.toPath(), Charset.defaultCharset())){
+            String line = reader.readLine();
+            while (line != null) {
+                JsonObject outcomeEntry = new JsonParser().parse(line).getAsJsonObject();
+                JsonArray outcomeIssues = outcomeEntry.getAsJsonArray("issue");
+                String finalLine = line;
+                outcomeIssues.forEach(issue-> {
+                    if("error".equals(issue.getAsJsonObject().get("severity").getAsString())){
```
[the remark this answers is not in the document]
SalGutierrez: not sure, but good point. ill make the change.

dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/SmokeTest.java
SalGutierrez: I think that would go beyond the scope of what we are trying to cover in our smoke tests. here this flag is only used to determine if the org can be safely deleted.

dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/ClientUtils.java
```diff
@@ -242,11 +246,36 @@ private static void jobResponseHandler(CloseableHttpClient client, JobCompletion
...
         } catch (IOException e) {
             throw new RuntimeException("Cannot output file", e);
         }
     }
 
+    private static void verifyOperationOutcomeContents(File outcomeFile){
+        try (BufferedReader reader = Files.newBufferedReader(outcomeFile.toPath(), Charset.defaultCharset())){
+            String line = reader.readLine();
+            while (line != null) {
+                JsonObject outcomeEntry = new JsonParser().parse(line).getAsJsonObject();
```
whuang85, line 264: You can use the fhircontext to get a json parser that will parse it directly to an OperationOutcome object

dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/ClientUtils.java
```diff
@@ -242,11 +246,36 @@ private static void jobResponseHandler(CloseableHttpClient client, JobCompletion
         try {
             final File file = fetchExportedFiles(entry.getUrl(), client);
             System.out.println(String.format("Downloaded file to: %s", file.getPath()));
+            if(file.length() == 0){
+                throw new IllegalStateException(String.format("Downloaded file was empty. file path:  %s", file.getPath()));
+            }
+            if(ResourceType.OperationOutcome.equals(entry.getType())){
```
whuang85, line 252: I don't think this will ever get called because you already throw an exception if there are errors

SalGutierrez: My understanding is an OperationOutcome does not always have to be an error. But now that you mentioned it, I believe this would still get called but it would never fail validation since it would have thrown an exception before.

dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/ClientUtils.java
```diff
@@ -61,11 +66,10 @@ static void handleExportJob(IGenericClient exportClient, List<String> providerNP
...
                 .map(search -> (Group) search.getEntryFirstRep().getResource())
                 .map(group -> jobCompletionLambda(exportClient, httpClient, group, overrideURL))
-                //TODO: ignore until we can skip lookback per request and switch over to use real bfd client
-//                .peek(jobResponse -> {
-//                    if (jobResponse.getError().size() > 0)
-//                        throw new IllegalStateException("Export job completed, but with errors");
-//                })
+                .peek(jobResponse -> {
+                    if (jobResponse.getError().size() > 0)
+                        throw new IllegalStateException("Export job completed, but with errors");
```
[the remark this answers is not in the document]
SalGutierrez: They are found in the operation outcome files.

dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/SmokeTest.java
SalGutierrez: They are the same. We will need to revisit refactoring or rewriting these tests eventually.

dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/ClientUtils.java
SalGutierrez: Fixed: Removed this validation since it would never get reached.

whuang85 (approved):

PR #1108, opened 2020-10-29
DPC-822: Warn on endpoint registration that the domain of the URL must be valid

Sun-Mountain:
### Fixes [DPC-822](https://jira.cms.gov/browse/DPC-822)

Akamai is busy protecting our web app and API. One of the protections we discovered recently is that when we POST a URL in a form, it is validated by the Akamai's web application firewall. This can be surprising when we're working with test data, and expect a nominally-formatted URL to work. And instead of getting an error message from the web app, Akamai sends back its own unformatted "unauthorized" message.

To reduce the surprise factor, we should prompt for a semi-valid URL in any form that need them. In particular, the endpoint registration form will need this.

AC:
- Identify all places the web app is asking for URL's
- Clarify on the form that a valid domain name is required

### Proposed Changes

- Clarified on the existing error that they need a valid domain name

### Security Implications

### Acceptance Validation

Screen Shot 2020-10-29 at 3 13 51 PM


dhgreene (approved):
👍

PR #1113, opened 2020-11-03
DPC-772 Add DPC specific headers to BFD call

whuang85:
### Fixes [DPC-772](https://jira.cms.gov/browse/DPC-772)

BFD is now outlining the app specific headers it requires from each app that uses it.  DPC is now sending the applicable headers BFD is requiring.

### Proposed Changes

* Add provider id to MDC
* Use MDCs to populate the headers sent to BFD

### Security Implications

Only additional Ids that will be in the logs will be the UUID associated to the provider which is not considered sensitive information.

### Acceptance Validation

Used an existing org in dev environment, uploaded public key, created access token, created patient, provider, roster and submitted for export from local computer to postman,

Saw my VPN's ip address in the logs when the batch was being processed (won't post screen shot in case IP is sensitive info)


jonfulk (approved):
PR is missing Acceptance Validation, but otherwise LGTM

SalGutierrez (approved):
LGTM

dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/client/BlueButtonClientImpl.java
```diff
@@ -268,4 +279,23 @@ public String hashMbi(String mbi) throws GeneralSecurityException {
...
+            query.withAdditionalHeader(Constants.INCLUDE_IDENTIFIERS_HEADER, "mbi");
+            var providerId = MDC.get(MDCConstants.PROVIDER_ID);
+            var jobId = MDC.get(MDCConstants.JOB_ID);
+            if (StringUtils.isNotBlank(providerId)) {
+                query.withAdditionalHeader(Constants.BULK_CLIENT_ID_HEADER, providerId);
+            }
+            if (StringUtils.isNotBlank(jobId)) {
+                query.withAdditionalHeader(Constants.BULK_JOB_ID_HEADER, jobId);
+            }
+            query.withAdditionalHeader(HttpHeaders.X_FORWARDED_FOR, InetAddress.getLocalHost().getHostAddress());
```
dhgreene, line 295: Have we tested in `dev` from an external address, to see whether we get the actual client IP or Akamai's IP?

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
```diff
@@ -272,7 +274,8 @@ public Response export(@ApiParam(hidden = true)
                            @DefaultValue(FHIR_NDJSON) @QueryParam("_outputFormat") @NoHtml String outputFormat,
                            @ApiParam(value = "Resources will be included in the response if their state has changed after the supplied time (e.g. if Resource.meta.lastUpdated is later than the supplied _since time).")
                            @QueryParam("_since") @NoHtml String sinceParam,
-                           @ApiParam(hidden = true) @HeaderParam("Prefer")  @Valid String Prefer) {
+                           @ApiParam(hidden = true) @HeaderParam("Prefer")  @Valid String Prefer,
+                           @Context HttpServletRequest request) {
```
SalGutierrez, line 278: Not an issue, but we could just inject the HttpHeaders instead of the entire request.

whuang85: > Not an issue, but we could just inject the HttpHeaders instead of the entire request.

I don't just use the x-forwarded-for header, if it's not present then I'll call request.getRemoteAddr()

dpc-common/src/main/java/gov/cms/dpc/common/Constants.java
```diff
@@ -0,0 +1,10 @@
+package gov.cms.dpc.common;
```
SalGutierrez, line 1: Since we have an MDCConstants.java file, maybe rename this to something more specific? HeadersConstants? Unless this will be a catch all constants file and just happens to have only headers at the moment.

jonfulk (approved):
👍🏻👍🏻

SalGutierrez (approved):

dpc-common/src/main/java/gov/cms/dpc/common/Constants.java
whuang85: I was going to use this as a catch all in the future

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/GroupResource.java
SalGutierrez: oh right. good point.

SalGutierrez:
Ship it. 🚢

PR #1114, opened 2020-11-05
DPC-821: Validate NPI on org creation (admin)

Sun-Mountain:
### Fixes [DPC-821](https://jira.cms.gov/browse/DPC-821)

In prod-sandbox we're auto-generating fake NPI's for new organizations. However, in dev and test we don't. Also, on the admin page for org creation, we're not validating the NPI. This causes a problem when it's time to enable API access, because an invalid NPI will fail with a generic error.

AC:
- Validate the NPI on the admin org creation form
- Allow the NPI to begin with a number other than 1 or 2

### Proposed Changes

- Set up validation for real NPI (for both admin and external user)
- Will throw an error explaining why the organization was not able to update (re: invalid NPI)

### Change Details

- Due to users being able to add/update the org's NPI in their portal, I added the same error message if they tried to add an invalid NPI to the org.
- Moved `generate_npi` to the `OrganizationHelper` for a more DRY application between the web app and tests.

### Security Implications

### Acceptance Validation

Screen Shot 2020-11-05 at 12 20 50 PM

Screen Shot 2020-11-05 at 12 21 23 PM


whuang85 (approved):

dpc-web/spec/factories/organizations.rb
```diff
@@ -17,4 +17,11 @@
       end
     end
   end
+
+  def generate_npi
```
dhgreene, line 21: Knowledge about NPI's and how we deal with the Luhancy gem is spread _soo_ many places now.  It would be nice to encapsulate it in its own class.

Sun-Mountain: Right now it should only be in two places: OrganizationsHelper and the FactoryBot. I _can_ just remove it from the FactoryBot and `include OrganizationsHelper` in every instance I create an org? Would that be better? (Asking before implementing.)

PR #1117, opened 2020-11-09
DPC-828: Update filtering to allow filtering by tag

Sun-Mountain:
### Fixes [DPC-828](https://jira.cms.gov/browse/DPC-828)

We have tagging available for both users and organizations, but we cannot currently filter views according to those tags. We will start using tags to designate users who reply to emails with "Do Not Contact" - we need a tool to be able to find those users quickly.

### Proposed Changes

An admin user in the portal can filter views by tags.

### Change Details

- The User model is throwing an error with code climate for being one line too long
- Added UX design for tags in the filter modal (may want to consult designer for better design if number of tags becomes too many for a single view)

### Security Implications

### Acceptance Validation

Screen Shot 2020-11-10 at 8 53 03 AM

Screen Shot 2020-11-10 at 8 53 06 AM

Screen Shot 2020-11-10 at 8 53 14 AM

### DIFF
--- a/dpc-web/app/assets/javascripts/application.js
+++ b/dpc-web/app/assets/javascripts/application.js
@@ -17,6 +17,7 @@
 //= require copy-text
 //= require password-check
 //= require rails-ujs
+//= require tag-selector
 //= require activestorage
 //= require_tree ./utils
 //= require_tree .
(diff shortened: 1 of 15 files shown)


SMLuthi (dismissed):
LGTM!

PR #1123, opened 2020-11-23
DPC-902: Fix return value when adding too many providers

jonfulk:
### Fixes [DPC-902](https://jira.cms.gov/browse/DPC-902)

If a new user with no existing providers tried to do a bulk upload of more providers than allowed, they would receive a 500 due to an IndexOutOfBounds error. This is because the response was trying to include the first of a list of existing providers when there were no existing providers. The test was missing this case by first successfully uploading a provider and then trying to upload the same provider again with the same NPI.

### Proposed Changes

First, change the test so that the 2nd provider uploaded has a different NPI and therefore has no existing providers to return from. Secondly, include the entity in question in the response instead of trying to find one from a list when there possibly isn't one there.

### Change Details

- create a new practitioner with a new NPI for 2nd attempt in the test
- just return the `entity` in the response instead of trying to get a provider from a list of existing ones as there may not be any in that list

### Security Implications

### Acceptance Validation

1) Ran `make ci-app` which runs all unit and integration tests as well as the end-to-end tests
2) Ran smoke tests locally

All passed.

Screen Shot 2020-11-23 at 4 38 17 PM

Screen Shot 2020-11-23 at 5 01 14 PM

### Feedback Requested

Is there a cleaner way to change than test than what I have done here?
Anything else?


dhgreene (approved):
LGTM 👍

SalGutierrez (approved):
LGTM🚀⛴️

PR #1125, opened 2020-11-30
DPC-774  investigate patient tests

whuang85:
### Fixes [DPC-774](https://jira.cms.gov/browse/DPC-774)

Investigated the patient tests, noticed that changing the duration and/or frequency would break the tests, reason was we have a set number of sample patients and it'll send in nil for the rest. Also, org can fail to create and it would be useful to see the message and failure at that point instead of later in the process (getting a token for the non created org)

### Proposed Changes

* loop around the patients
* read the operation outcome of the org creation to determine if it should fail

### Security Implications

### Acceptance Validation

Screen Shot 2020-11-30 at 11 52 03 AM

### DIFF
--- a/dpc-testing/performance/pkg/dpc/dpc.go
+++ b/dpc-testing/performance/pkg/dpc/dpc.go
@@ -14,6 +14,7 @@ type Resource struct {
 	ID          string `json:"id"`
 	ClientToken []byte `json:"token"`
 	AccessToken string `json:"access_token"`
+	TYPE        string `json:"resourceType"`
 }
 
 // Pull `ids` out of a set of response bodies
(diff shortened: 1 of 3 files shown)


dpc-testing/performance/pkg/dpc/dpc.go
```diff
@@ -14,6 +14,7 @@ type Resource struct {
 	ID          string `json:"id"`
 	ClientToken []byte `json:"token"`
 	AccessToken string `json:"access_token"`
+	TYPE        string `json:"resourceType"`
```
dhgreene, line 17: It's more Go-like to not use all caps.  Shall we go with `Type`?

dhgreene (approved):
Good catch! 🏎️

PR #1126, opened 2020-11-30
DPC-849: Prohibited test org ids from being used when org is created via attir…

SalGutierrez:
### Fixes [DPC-849](https://jira.cms.gov/browse/DPC-849)

If no org id is provided when creating a new org a randomly generated UUID is used. While EXTREMELY unlikely, it is possible that the generated UUID matches one of the lookback exempt UUIDs used for test purposes, allowing the newly created org to bypass lookback. 
### Proposed Changes
Prohibit attribution from creating a UUID that is included in lookback exempt list.

### Change Details
Generate a random UUID that is not included in prohibited list before converting FHIR model into OrganizationEntity.

### Security Implications

### Acceptance Validation

Tested with unit tests. Smoke tests pass in dev.

### Feedback Requested
Verify logic.

### DIFF
--- a/dpc-attribution/src/main/java/gov/cms/dpc/attribution/resources/v1/OrganizationResource.java
+++ b/dpc-attribution/src/main/java/gov/cms/dpc/attribution/resources/v1/OrganizationResource.java
@@ -26,6 +26,7 @@ import javax.ws.rs.core.Response;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import static gov.cms.dpc.attribution.utils.RESTUtils.parseTokenTag;
@@ -39,6 +40,7 @@ public class OrganizationResource extends AbstractOrganizationResource {
     private final EndpointDAO endpointDAO;
     private final FHIREntityConverter converter;
     private final DPCAttributionConfiguration config;
+    private final Supplier<UUID> uuidSupplier = UUID::randomUUID;
 
     @Inject
     OrganizationResource(FHIREntityConverter converter, OrganizationDAO dao, EndpointDAO endpointDAO, DPCAttributionConfiguration config) {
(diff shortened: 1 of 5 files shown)


dpc-attribution/src/main/java/gov/cms/dpc/attribution/resources/v1/OrganizationResource.java
```diff
@@ -175,4 +181,15 @@ public Response deleteOrganization(@ApiParam(value = "Organization resource ID",
                 .map(e -> this.converter.fromFHIR(EndpointEntity.class, e))
                 .collect(Collectors.toList());
     }
+
+    private String getNewOrgId(Organization organization){
```
whuang85, line 185: maybe move this method into APIHelpers as a static method if you see this type of logic being used else where too (tests included)

SalGutierrez: I don't really see it being used anywhere else but here.

dpc-attribution/src/main/java/gov/cms/dpc/attribution/resources/v1/OrganizationResource.java
```diff
@@ -81,18 +87,22 @@
...
-        final Optional<Organization> organization = transactionBundle
+        final Optional<Organization> optOrganization = transactionBundle
                 .getEntry()
                 .stream()
                 .filter(entry -> entry.hasResource() && entry.getResource().getResourceType() == ResourceType.Organization)
                 .map(entry -> (Organization) entry.getResource())
                 .findFirst();
 
-        if (organization.isEmpty()) {
+        if (!optOrganization.isPresent()) {
```
SalGutierrez, line 97: Swapped this only because code climate was complaining about using .isPresent()

dhgreene (approved):
Thanks for closing this loophole.  Love the tests!  💯

PR #1135, opened 2020-12-06
DPC-815 Feature: Add smoke test to Github Actions workflow

jeffcorpuz:
### Fixes [DPC-815](https://jira.cms.gov/browse/DPC-815)

Run smoke test in Github Actions workflow.

### Proposed Changes

Use `make smoke` to launch smoke tests

### Security Implications

### Acceptance Validation

Unfortunately, you can only verify this with a PR so check the GH action below 
https://github.com/CMSgov/dpc-app/pull/1135/checks?check_run_id=1508340485

```
 [INFO] Replacing /home/runner/work/dpc-app/dpc-app/dpc-smoketest/target/dpc-smoketest.jar with /home/runner/work/dpc-app/dpc-app/dpc-smoketest/target/dpc-smoketest-0.4.0-SNAPSHOT-shaded.jar
[INFO] Attaching shaded artifact.
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for DPC Application 0.4.0-SNAPSHOT:
[INFO] 
[INFO] DPC Application .................................... SUCCESS [ 45.954 s]
[INFO] Common DPC testing code ............................ SUCCESS [ 22.426 s]
[INFO] DPC Common Libraries ............................... SUCCESS [ 18.093 s]
[INFO] DPC SmokeTests ..................................... SUCCESS [ 22.878 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:56 min
[INFO] Finished at: 2020-12-06T22:41:13Z
[INFO] ------------------------------------------------------------------------
```

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -25,9 +25,7 @@ jobs:
       - name: "API Build"
         run: |
           make ci-app
-      - name: "Smoke Test"
-        run: |
-          make smoke
+
   build-web:
     name: "Build and Test Website"
     runs-on: ubuntu-latest
@@ -37,3 +35,22 @@ jobs:
       - name: "Website Build"
         run: |
           make ci-web
+          
+  smoke-test:
+    name: "Run Smoke Test"
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@v1
+      - name: "Set up JDK 11" 
+        uses: actions/setup-java@v1
+        with:
+          java-version: '11.0.2'
+      - name: "Set up Python 3.8.1"
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8.1'
+      - name: "Smoke Test"
+        run: |
+          make smoke
+  
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -25,6 +25,9 @@ jobs:
       - name: "API Build"
         run: |
           make ci-app
+      - name: "Smoke Test"
+        run: |
+          make smoke
   build-web:
     name: "Build and Test Website"
     runs-on: ubuntu-latest
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -14,17 +14,20 @@ jobs:
     steps:
       - name: "Checkout code"
         uses: actions/checkout@v1
-      - name: "Set up JDK 11" 
+      - name: "Set up JDK 11"
         uses: actions/setup-java@v1
         with:
-          java-version: '11.0.2'
+          java-version: "11.0.2"
       - name: "Set up Python 3.8.1"
         uses: actions/setup-python@v2
         with:
-          python-version: '3.8.1'
+          python-version: "3.8.1"
       - name: "API Build"
         run: |
           make ci-app
+      - name: "Smoke Test"
+        run: |
+          make smoke
 
   build-web:
     name: "Build and Test Website"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-workflow.yml
```diff
@@ -34,3 +35,22 @@ jobs:
...
+        uses: actions/setup-java@v1
+        with:
+          java-version: '11.0.2'
+      - name: "Set up Python 3.8.1"
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8.1'
+      - name: "Smoke Test"
+        run: |
+          make smoke
```
msnook, line 55: What is the difference between `make smoke` and `make smoke/local`?

.github/workflows/ci-workflow.yml
```diff
@@ -34,3 +35,22 @@ jobs:
       - name: "Website Build"
         run: |
           make ci-web
+          
+  smoke-test:
```
msnook, line 39: The way this is configured, I think it will run in parallel with `build-app` and `build-web`.  Should we run it immediately after `make ci-app` since the job name is "Build and Test API".  I think that will reduce code redundancy (i.e., setting up JDK and Python multiple times) and also ensure that we're running smoke against the actual build.

msnook (commented):
Looks good.  Just a few comments.

.github/workflows/ci-workflow.yml
jeffcorpuz: The difference between this and `smoke/local` is that it uses a spun up local environment from docker and runs it against it as well as skipping the `djib` tests which need to be enabled manually.

It will fail if we utilize `smoke/local` because it requires a person's input as well. You'd have to press enter to execute it as well as using an authenticated environment.

```Running Smoke Tests against Local env

=====
The Smoke Tests require an authenticated environment!
Verify your local API environment has "authenticationDisabled = false" or these tests will fail.
=====

Press ENTER to run the tests...
```

@dhgreene @DeirdreHolub can you provide more details than what I understand from reading the code?

.github/workflows/ci-workflow.yml
jeffcorpuz: Initially I chose that serial route, however I opted for this configuration to parallelize the build process. I chose this direction because want the jobs to fail fast if one of the jobs failed.

msnook: Makes sense from a performance perspective.  But, typically the feedback loop is build, unit, integration.
It might be confusing to developers if the feedback loop results in integration tests failing earlier than the build/unit has even processed.

dhgreene (commented):
That's a _lot_ of lines of download progress--to the point where it overwhelms the actual test output.  Is there a way to cut down on the noise?

msnook:
> That's a _lot_ of lines of download progress--to the point where it overwhelms the actual test output. Is there a way to cut down on the noise?

I only see ~500 lines of output for the new `Smoke Test` phase.  (See screenshot below).  Are you perhaps referring to the existing download progress which existed previously (in Travis, too).  Not sure that can be suppress, but if so, could we do it as a separate effort since I think it might be out-of-scope for getting the smoke tests running in CI.

Screen Shot 2020-12-07 at 5 33 26 PM

msnook:
> > That's a _lot_ of lines of download progress--to the point where it overwhelms the actual test output. Is there a way to cut down on the noise?
> 
> I only see ~500 lines of output for the new `Smoke Test` phase. (See screenshot below). Are you perhaps referring to the existing download progress which existed previously (in Travis, too). Not sure that can be suppressed, but if so, could we do it as a separate effort since I think it might be out-of-scope for getting the smoke tests running in CI?
> 
> Screen Shot 2020-12-07 at 5 33 26 PM

@dhgreene Update: [here](https://github.com/CMSgov/dpc-app/pull/1136) is a separate PR to suppress the transfer progress logging.

dhgreene (approved):
👍

msnook (approved):

PR #1138, opened 2020-12-09
DPC-916 fix: Remove requirement for Patient to be in a group for Patient/$everything

jonfulk:
### Fixes [DPC-916](https://jira.cms.gov/browse/DPC-916)

The Patient/$everything endpoint was unnecessarily requiring both a Provenance header (attesting to the provider/patient relationship) and a Roster for the given patient and provider. This was redundant and confusing for users.

### Proposed Changes

It was decided to remove the requirement for a Roster when using `Patient/$everything`
- Remove Roster check from the `PatientResource`
- Update tests for this change

### Change Details

- The `PatientResourceTest` is relying heavily on existing test data and the tests are dependent on one another. As part of this PR, I refactored some of this in order to aid in readability and adjust for the change to the `PatientResource`
- Since there was no longer a Roster, `LookBackServiceImpl. getPractitionerNPIFromRoster` was failing to return a Practitioner NPI. To fix this, I added a `ProviderDAO` and fallback to checking for the provider when there is not a Roster found.
- We no longer need to limit the RosterDAO search by the provider now that we are checking for it in the `LookBackServiceImpl`

### Security Implications

LookBackService continues to check for Practitioner NPIs and Organization NPIs on the EoBs as before.

### Acceptance Validation

- all tests pass when running `make ci-app` locally
- smoke tests pass locally

Screen Shot 2020-12-09 at 4 50 43 PM

### Feedback Requested

Any

### DIFF
--- a/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/AggregationAppModule.java
+++ b/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/AggregationAppModule.java
@@ -7,6 +7,7 @@ import com.google.inject.Provides;
 import com.hubspot.dropwizard.guicier.DropwizardAwareModule;
 import com.typesafe.config.Config;
 import gov.cms.dpc.aggregation.dao.OrganizationDAO;
+import gov.cms.dpc.aggregation.dao.ProviderDAO;
 import gov.cms.dpc.aggregation.dao.RosterDAO;
 import gov.cms.dpc.aggregation.engine.AggregationEngine;
 import gov.cms.dpc.aggregation.engine.JobBatchProcessor;
@@ -35,6 +36,7 @@ public class AggregationAppModule extends DropwizardAwareModule<DPCAggregationCo
         binder.bind(AggregationEngine.class);
         binder.bind(AggregationManager.class).asEagerSingleton();
         binder.bind(JobBatchProcessor.class);
+        binder.bind(ProviderDAO.class);
         binder.bind(RosterDAO.class);
         binder.bind(OrganizationDAO.class);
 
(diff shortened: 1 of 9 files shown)


jonfulk:
Regarding CodeClimate issues with more than 4 arguments and duplicate code: the constructor requires all of those for injection by guice. The only way to avoid this would be to create another class that takes a subset of those as a constructor, which is unnecessary.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/AggregationAppModule.java
```diff
@@ -94,13 +96,13 @@ public int provideJobTimeoutInSeconds() {
     }
 
     @Provides
-    LookBackService provideLookBackService(DPCManagedSessionFactory sessionFactory, RosterDAO rosterDAO, OrganizationDAO organizationDAO, OperationsConfig operationsConfig) {
+    LookBackService provideLookBackService(DPCManagedSessionFactory sessionFactory, ProviderDAO providerDAO, RosterDAO rosterDAO, OrganizationDAO organizationDAO, OperationsConfig operationsConfig) {
```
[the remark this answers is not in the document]
jonfulk: the constructor requires all of those for injection by guice. The only way to avoid this would be to create another class that takes a subset of those as a constructor, which is unnecessary.

dpc-queue/src/main/java/gov/cms/dpc/queue/service/DataService.java
```diff
@@ -53,17 +53,17 @@ public Resource retrieveData(UUID organizationId, UUID providerId, List<String>
      * Retrieves data from BFD
      * @param organizationID UUID of organization
      * @param providerID UUID of provider
-     * @param patientIDs List of patient String UUIDs
+     * @param patientMBIs List of patient String UUIDs
```
whuang85, line 56: could you change the description too?  Something like `List of patient MBIs`

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
```diff
@@ -198,16 +197,16 @@ public Bundle everything(@ApiParam(hidden = true) @Auth OrganizationPrincipal or
...
-
         final String requestingIP = APIHelpers.fetchRequestingIP(request);
         Resource result = dataService.retrieveData(orgId, practitionerId, List.of(patientMbi), APIHelpers.fetchTransactionTime(bfdClient),
                 requestingIP, ResourceType.Patient, ResourceType.ExplanationOfBenefit, ResourceType.Coverage);
         if (ResourceType.Bundle.equals(result.getResourceType())) {
             return (Bundle) result;
         }
+        if (ResourceType.OperationOutcome.equals(result.getResourceType())) {
+            OperationOutcome resultOp = (OperationOutcome) result;
+            throw new NotAuthorizedException(resultOp.getIssueFirstRep().getDetails().getText());
```
whuang85, line 208: not sure about the name of this exception as an operation outcome could be returned due to BFD being down or the MBI was in a bad format or something else.

dpc-queue/src/main/java/gov/cms/dpc/queue/service/DataService.java
jonfulk: good point. i'll do that!

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
jonfulk: @whuang85 Are you saying that it should be a different exception type, that we should check something else before throwing an exception, and/or that the message could fail if there isn't an issue with details and text? All of the above is an acceptable answer.

whuang85: @jonfulk just not sure if it should be returning a 401 when the reason data could not be returned is because either BFD is down, mbi is malformed, or look back failed

whuang85 (approved):

SalGutierrez (approved):

PR #1139, opened 2020-12-11
DPC-952 Do not allow an org to view another org's Group

SalGutierrez:
### Fixes [DPC-952](https://jira.cms.gov/browse/DPC-952)

Currently, if you have API access and can guess the UUID of any existing Group resource, you can show the UUID's of the patients in that group. This is too broad of an access; each organization should only be able to see its own resources, including Groups.

During the implementation of this fix it was found that an organization could also delete or update another's group. This patch will fix theses issues as well.

### Proposed Changes

How path authentication is handled:

1. Path authentication (triggered by PathAuthorizer annotation on a resource) is handled by the "MacaroonsAuthorizer" before a request is sent to a resource. 

2. The MacaroonsAuthorizer uses an attribution client to search for resources (Resource searched is based on path) filtered by (organization or tag) and resource UUID (specified in path) and allows request to proceed if a resource is returned.

Cause of issue being reported:

DPC-Attribution's Group search did not support resource UUID as a filter, so it was being ignored. Since the UUID of the resource in the path was being ignored the search simply returned all groups belonging to the organization, this would allow access to any group (as long as there was a result in the search)

### Change Details

Add support for the _id filter in group search (dpc-attribution)

### Security Implications

### Acceptance Validation
Tested Locally - All tests pass
Pending manual test on dev

### Feedback Requested

Any

### DIFF
--- a/dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/GroupResourceTest.java
+++ b/dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/GroupResourceTest.java
@@ -4,10 +4,12 @@ import ca.uhn.fhir.parser.IParser;
 import ca.uhn.fhir.rest.api.MethodOutcome;
 import ca.uhn.fhir.rest.client.api.IGenericClient;
 import ca.uhn.fhir.rest.gclient.ICreateTyped;
+import ca.uhn.fhir.rest.server.exceptions.AuthenticationException;
 import ca.uhn.fhir.rest.server.exceptions.InvalidRequestException;
 import ca.uhn.fhir.rest.server.exceptions.UnprocessableEntityException;
 import gov.cms.dpc.api.APITestHelpers;
 import gov.cms.dpc.api.AbstractSecureApplicationTest;
+import gov.cms.dpc.api.TestOrganizationContext;
 import gov.cms.dpc.common.utils.NPIUtil;
 import gov.cms.dpc.common.utils.SeedProcessor;
 import gov.cms.dpc.fhir.DPCIdentifierSystem;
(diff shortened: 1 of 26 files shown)


dpc-attribution/src/main/java/gov/cms/dpc/attribution/resources/v1/GroupResource.java
```diff
@@ -108,7 +108,9 @@ public Response createRoster(Group attributionRoster) {
             "<p> You can search for Groups associated to a given provider (via the Provider NPI) and groups for which a patient is a member of (by the Patient resource ID)",
             response = Bundle.class)
     @Override
-    public List<Group> rosterSearch(@ApiParam(value = "Organization ID")
+    public List<Group> rosterSearch(@ApiParam(value = "Group resource UUID")
+                                        @QueryParam(Group.SP_RES_ID) UUID rosterID,
```
SalGutierrez, line 112: I tried hard to format this indentation, but failed and gave up. 😞

dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/GroupResourceTest.java
```diff
@@ -395,4 +400,157 @@ public void testProvenanceHeaderAndGroupProviderMatch() {
                 .encodedJson()
                 .execute();
     }
+
+    @Test
+    public void testGroupCanOnlyBeRetrievedByOwner() throws GeneralSecurityException, IOException, URISyntaxException {
```
whuang85, line 405: only going by the names of the methods but did you test that a org cannot add another org's patients via UUID

SalGutierrez: Not in this PR, that issue and fix is captured in another ticket I'm working on.

whuang85 (approved):

PR #1142, opened 2020-12-15
DPC-780 Have major Snyk finding fail builds

dhgreene:
### Fixes [DPC-780](https://jira.cms.gov/browse/DPC-780)
We would like to avoid pushing out software with a vulnerable dependency.  Our Snyk integration with Jenkins allows this, and a [separate ticket in the ops repo](https://github.com/CMSgov/dpc-ops/pull/279) will check the correct box (literally, _un_check the box by default, actually).  

However, we currently have a few known vulnerabilities in software that we only use in test environments, and for which we don't have a good upgrade path (we've [already upgraded to the latest available library](https://github.com/CMSgov/dpc-app/pull/1115)).  This PR gives us 30 days of silence from Snyk on those particular issues, so that we can proceed to reduce the changes of publishing new, _unknown_ issues.

### Proposed Changes
- Add a .snyk file 

### Change Details
The file was produced with the `snyk wizard` CLI tool

### Security Implications

By itself, this PR does nothing.  There's no effect on CI/CD, because in Jenkins the Snyk findings are ignored by default.  When paired with the associated ops PR, it allows us to focus on new issues not limited to our testing environment.

#### Build job passes with .snyk file (it currently fails if SNYK_OVERRIDE is unchecked)
Screen Shot 2020-12-15 at 4 24 15 PM
Screen Shot 2020-12-15 at 4 23 56 PM

### Feedback Requested
- Are we ignoring too many issues?
- Is there a better way to approach this topic?

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,131 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.14.1
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  SNYK-JS-LODASH-590103:
+    - newman > lodash:
+        reason: Affects tests only
+        expires: '2021-01-14T20:54:26.490Z'
+    - newman > postman-collection > lodash:
+        reason: Affects tests only
+        expires: '2021-01-14T20:54:26.490Z'
+    - newman > postman-runtime > lodash:
+        reason: Affects tests only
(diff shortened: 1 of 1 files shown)


jonfulk (approved):
This makes sense to me as there is nothing we can do until our dependency's dependencies get upgraded, especially since it's just for testing.

PR #1144, opened 2020-12-16
DPC-953 Prohibit orgs from adding another orgs patient to their group.

SalGutierrez:
### Fixes [DPC-953](https://jira.cms.gov/browse/DPC-953)

Issue: Organizations could add another org's patient to their group during group creation (if they guessed their patient's UUID), update and with the $add operation

During implementation it was also discovered that a method using @Auth AND @PathAuthorizer would ignore the @PathAuthorizer and not perform path authentication (just regular auth). This new issue was fixed in this PR since it was required in order to implement the original fix.

Cause of org being able to add another org's patient:
- Attribution only checked for a patients existence, it did not verify that the patient belonged to the org.

Cause of @PathAuthorizer being ignored when used with @Auth:

- We have 2 Authorizers:  the StandardAuthorizer and PathAuthorizer
- DPCAuthDynamicFeature.java determined what authorizer to use, if method contained @Auth the StandardAuthorizer was used, if it contained @PathAuthorizer the PathAuthorizer was used.
- @Auth came first in the list, if it was found it would use its authorizer and return, and never checked for @PathAuthorizer.

### Proposed Changes

- In attribution: check if a patient belongs to the org before adding them to group.
- @Auth should be used to inject the authenticated principal regardless of what authorizer is used; it should not be used to determine what authorizer to use.
- Introduce new security annotation "Authorizer" to specify the use of the regular authorizer, allowing the @Auth param to be used solely for its intended purpose (Injecting the principal).
Resolution

### Change Details

- Verified patients in group by searching by UUID and Org id instead of just looking up by UUID.
- Added org tag to group before making request to attribution (required to extract org id in attribution)
- Created new @Authorizer annotation, and added the annotation to any method requiring standard authentication
- re-arranged order of authorizer selection in DPCAuthDynamicFeature.java from most to least restrictive.
- Added integration tests

### Security Implications

### Acceptance Validation

Locally make-ci app and smoke tests

### Feedback Requested
thoughts on new auth changes?

### DIFF
--- a/dpc-api/src/test/java/gov/cms/dpc/api/APIResourceAnnotationTest.java
+++ b/dpc-api/src/test/java/gov/cms/dpc/api/APIResourceAnnotationTest.java
@@ -4,7 +4,8 @@ import com.codahale.metrics.annotation.ExceptionMetered;
 import com.codahale.metrics.annotation.Timed;
 import com.google.common.collect.ImmutableSet;
 import gov.cms.dpc.api.auth.annotations.AdminOperation;
-import gov.cms.dpc.api.auth.annotations.PathAuthorizer;
+import gov.cms.dpc.api.auth.annotations.Authorized;
+import gov.cms.dpc.api.auth.annotations.PathAuthorized;
 import gov.cms.dpc.api.auth.annotations.Public;
 import gov.cms.dpc.api.resources.v1.BaseResource;
 import gov.cms.dpc.testing.BufferedLoggerHandler;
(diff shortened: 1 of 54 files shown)


whuang85 (approved):

dpc-api/src/test/java/gov/cms/dpc/api/APIResourceAnnotationTest.java
```diff
@@ -82,28 +81,13 @@ void allResourcesHaveSecurityAnnotations() {
 
     /**
      * Asserts that the method has valid auth-annotations
-     * To pass, the method must either have a parameter with an Auth or a PathAuthorizer or Public annotation on the method
+     * To pass, the method must either have a parameter with an Authorized, PathAuthorized, or Public annotation on the method
```
jonfulk, line 84: Should this be `Authorizer` and `PathAuthorizer`?

jonfulk (approved):
👍

dpc-attribution/src/test/java/gov/cms/dpc/attribution/resources/v1/GroupResourceUnitTest.java
```diff
@@ -0,0 +1,149 @@
...
+    private Map<UUID,Patient> makeTestPatients(int count, UUID orgId){
+        if(count>88){
+            throw new IllegalStateException("Don't support building more than 88 patients..yet (need a better mbi generator)");
+        }
+        final Map<UUID,Patient> patients = Maps.newHashMap();
+        while(count>0){
+            UUID id = UUID.randomUUID();
+            Patient patient = FHIRPatientBuilder
+                    .newBuild()
+                    .withMbi("4S41C00AA"+(count+10)) //makes MBI in range 4S41C00AA10 -> 4S41C00AA99
```
whuang85, line 139: You can try one of those regex to string generators https://github.com/curious-odd-man/RgxGen

whuang85 (approved):

PR #1147, opened 2020-12-18
Added suport for id filter in Endpoint resource search

SalGutierrez:
### Fixes [DPC-974](https://jira.cms.gov/browse/DPC-974)

During the audit/risk assessment of DPC API it was discovered that path authorization for the /Endpoint resource was not functioning correctly.

A fully authenticated client could have access to another organization's Endpoint resource and its operations (get,update,delete).

Affected endpoints:

GET /Endpoint 
PUT /Endpoint 
Delete /Endpoint

#### Cause:

MacaroonsAuthenticator was attempting to authorize requests to Endpoint resource by using DPC-attribution's Endpoint search API. The MacaroonsAuthenticator attempted to search for an EndpointResource filtering by "_id" (id of the resource specified in the path by the client) and "organization" (Extracted from authenticated principal/token).

 If the search returned a result it meant there was a resource with that specified id and belonging to the organization, so the request would pass the path filter and allow access to the resource.

Since dpc-attribution did not support the "_id" filter param the actual search was only using "organization", therefore all resources for the organization were being returned from dpc-attribution to the filter, and the request was allowed to proceed.

### Proposed Changes

- Update the Endpoint search in DPC-Attribution to support the filtering by resource id (_id query parameter).
- Write integration tests to verify that the MacaroonsAuthenticator filter is working as intended.

### Change Details

- Added optional _id filter parameter in EndpointResource search API
- Added integration tests and updated existing tests.

Currently if a client attempts to retrieve, modify, or delete an Endpoint resource that does not exist the client is presented with a 404 Not Found response. After this patch the client will return a 401 Not Authorized regardless due to the filter's inability to distinguish between a resource not existing with the specified UUID or the resource existing but not belonging to the organization making the request.

Returning a 401 for a resource not found is consistent with the rest of the resources using the PathAuthorizer, but we would need to discuss whether or not it's considered a breaking change.

### Security Implications

### Acceptance Validation

- Tested locally using integration tests
- make ci-app and smoke-tests were successful locally

### Feedback Requested

- Mostly on the integration tests, missing any? Changes? Holes?
- Is returning a 401 instead of a 404 considered a breaking change? If so, how should we communicate this?

### DIFF
--- a/dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/EndpointDAO.java
+++ b/dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/EndpointDAO.java
@@ -29,7 +29,7 @@ public class EndpointDAO extends AbstractDAO<EndpointEntity> {
         return Optional.ofNullable(get(endpointID));
     }
 
-    public List<EndpointEntity> search(UUID organizationID, UUID endpointId) {
+    public List<EndpointEntity> endpointSearch(UUID organizationID, UUID endpointId) {
         // Build a selection query to get records from the database
         final CriteriaBuilder builder = currentSession().getCriteriaBuilder();
         final CriteriaQuery<EndpointEntity> query = builder.createQuery(EndpointEntity.class);
--- a/dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/EndpointDAO.java
+++ b/dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/EndpointDAO.java
@@ -1,5 +1,6 @@
 package gov.cms.dpc.attribution.jdbi;
 
+import com.google.common.collect.Lists;
 import gov.cms.dpc.common.entities.EndpointEntity;
 import gov.cms.dpc.common.hibernate.attribution.DPCManagedSessionFactory;
 import io.dropwizard.hibernate.AbstractDAO;
@@ -7,6 +8,7 @@ import io.dropwizard.hibernate.AbstractDAO;
 import javax.inject.Inject;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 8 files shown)


dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/EndpointDAO.java
```diff
@@ -27,15 +29,22 @@ public EndpointEntity persistEndpoint(EndpointEntity endpointEntity) {
...
         // Build a selection query to get records from the database
         final CriteriaBuilder builder = currentSession().getCriteriaBuilder();
         final CriteriaQuery<EndpointEntity> query = builder.createQuery(EndpointEntity.class);
         final Root<EndpointEntity> root = query.from(EndpointEntity.class);
         query.select(root);
 
+        final List<Predicate> predicates = Lists.newArrayList();
 
-        query.where(builder.equal(root.get("organization").get("id"), organizationID));
+        if(organizationID!=null){
```
jonfulk, line 41: in the case where the organizationID and/or the endpointID is/are null, do we want to throw some sort of error? Otherwise, wouldn't we end up just not filtering the results by those predicates?

dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/EndpointDAO.java
```diff
@@ -27,15 +29,22 @@ public EndpointEntity persistEndpoint(EndpointEntity endpointEntity) {
         return Optional.ofNullable(get(endpointID));
     }
 
-    public List<EndpointEntity> findByOrganization(UUID organizationID) {
+    public List<EndpointEntity> search(UUID organizationID, UUID endpointId) {
```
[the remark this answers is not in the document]
SalGutierrez: changed for consistency.

dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/EndpointDAO.java
SalGutierrez: Thats right, no predicates would just translate to a select.
In the rest of the DAOs we do the same. Not allowing a search without filters is more of a business requirement and probably does not belong in the DAOs.

jonfulk (approved):

PR #1151, opened 2020-12-21
DPC-979 Fix Group & Practitioner resource request body forgery issue.

SalGutierrez:
### Fixes [DPC-979](https://jira.cms.gov/browse/DPC-979)

During the audit/risk assessment of DPC API it was discovered that for certain resources are not validating values passed in by the client.

Practitioner: 
Specifying an org id in the meta tag allows an authenticated clients to register a practitioners under any org.

Group:
Specifying an org ID in the met tag and NPI allows the retrieval of another org's roster during creation.

Cause:
- Organization tag is added using the authenticated principal in API.
- Attribution retrieves the value of the FIRST tag with DPC System code. 
- A client is capable of adding their own tag with DPC SYSTEM code and specifying another org's id for the value.
- API would append the authenticated org's id to the tag, but it wold no longer be the first in the list so it would be ignored by attribution and use the client specified tag instead.

### Proposed Changes

Sanitize the organization tags before sending the request to attribution.

### Change Details

- Removed any tags with the DPC SYSTEM before setting our own. 
- Added integration tests.

### Security Implications

### Acceptance Validation

created integration tests.
make ci-app is successful locally.

### DIFF
--- a/dpc-api/src/main/java/gov/cms/dpc/api/APIHelpers.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/APIHelpers.java
@@ -64,6 +64,7 @@ public class APIHelpers {
             newMeta.addTag(orgTag);
             resource.setMeta(newMeta);
         } else {
+            meta.getTag().removeIf(coding -> DPCIdentifierSystem.DPC.getSystem().equalsIgnoreCase(coding.getSystem().trim()));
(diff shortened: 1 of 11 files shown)


whuang85 (approved):

jonfulk (approved):
looks good! Thanks for finding this, Sal!

PR #1156, opened 2021-01-05
DPC-935 fix volume mapping for local testing

jonfulk:
### Fixes [DPC-935](https://jira.cms.gov/browse/DPC-935)

When running all containers except api in the terminal and running the `Patient/{id}/$everything` tests on api in IntelliJ, there were a few 500 errors. To solve, you have to run aggregation in IntelliJ as well. This is because output files are being written to `/tmp` inside the container and not locally, meaning the volume mapping was not working correctly.

### Proposed Changes

- Created a `docker-compose-local.yml` file to override volume settings when running locally
- Add a `make` command called `make start-local` to run everything but `dpc-api` using the custom volume settings.
- Add a `make start-local-api` command that starts up the entire app using `docker-compose-local.yml` 

### Change Details

I chose not to start up `api` in the `make start-local` command since the regular practice when testing locally is to start up all containers using `make start-app` and then shut down `api` in order to test it in IntelliJ. This allows for easy api testing in IntelliJ using a single command to start the app locally,  and it also stores all of the generated output in the local `/tmp` directory for easier debugging.

I also created a `make start-local-api` command which can be used to start the entire app using `docker-compose-local.yml` or it can be run after `make start-local` to add on the `api` as well. The only caveat is that the custom local settings will not work for smoke tests as the local version of these commands use the mock BFD.

### Security Implications

### Acceptance Validation

No actual code changes were made, so smoke tests and rebuild were unnecessary. However, for the record, `make ci-app` did complete successfully. Smoke tests will fail when running local version as they require interaction with BFD and not just the mock. 

Most importantly, the `Patient/{id}/$everything` tests that revealed this issue in the first place are now all passing when running everything except the api in the docker containers using the new `make` command. 

Screen Shot 2021-01-05 at 3 40 17 PM

### Feedback Requested

Is `make start-local-api` really useful? Should we set the `USE_BFD_MOCK` variable to `false` when starting up the api, but not when starting up the other containers locally? Anything else?


dhgreene (approved):
Good improvement!  💯

PatKelsh (approved):
looks good to me

PR #1167, opened 2021-01-07
DPC-950 fix Error Prone warnings

jonfulk:
### Fixes [DPC-950](https://jira.cms.gov/browse/DPC-950)

After upgrading Error Prone, a list of 7 warnings was generated. In reality, there were 32 warnings in all. The task of this ticket was to fix all of those warnings and show that no more warnings exist in order to eventually re-enable error prone.

### Proposed Changes
- Ignore usage of the obsolete `Date` class when it was only being used for FHIR stu3 object conversion
- Replace `Date` with `YearDate` for LookBack when only the month mattered
- Replace `Date` with `LocalDate` or `Instant` when possible, as appropriate
- Replace usage of obsolete `SortedMap` with the newer and supported `NavigableMap`
- Add an ID to the test Group entity to avoid a StackOverflow error that was occurring when running tests due to trying to turn a missing ID into a String
- Other minor fixes due to warnings indicated by IntelliJ during development such as making variables `final`

### Change Details

Most of the issues in these warnings were related to the obsolete `Date` class. In almost all of the cases (other than `LookBackAnswer`), we are using `Date` only because the HAPI FHIR stu3 resource requires it. As it comes in, we immediately convert it to a `LocalDate` in most cases. Because of this, in many of the places where we were using `Date`, we are now just going to suppress that warning as the third party code is beyond our control. Otherwise, it has been updated with an appropriate type from `java.time.*`.

`LookBackAnswer` is one area where we were unnecessarily using `Date` for both `lookBackMonth` and `billingPeriodEndDate`. In both cases, we only care about the month as we are comparing the months difference between the two in order to determine if it is within the LookBack timeframe. For this reason, I chose to use `YearMonth` for both of these variables. This required changes to configuration and multiple tests.

### Security Implications

### Acceptance Validation
✔️  no `errorprone` warnings found: to confirm, see most recent api build and search for `https://errorprone` (should be 0 results)
✔️  `make ci-app`: pass
✔️  local smoke tests: pass
✔️  full build and deploy to dev: pass

Screen Shot 2021-01-08 at 2 16 02 PM

Screen Shot 2021-01-08 at 4 43 22 PM

### Feedback Requested

Any


dpc-queue/src/main/java/gov/cms/dpc/queue/JobQueueCommon.java
```diff
@@ -69,8 +69,4 @@ protected JobQueueBatch createJobBatch(UUID jobID,
                 createJobBatch(jobID, orgID, providerID, Collections.emptyList(), resourceTypes, since, transactionTime, null, true)
         );
     }
-
-    public int getBatchSize() {
```
jonfulk, line 73: This method was not used anywhere and was throwing a warning.

dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/client/MockBlueButtonClient.java
```diff
@@ -131,7 +130,8 @@ public CapabilityStatement requestCapabilityStatement() throws ResourceNotFoundE
     }
 
     @Override
-    public String hashMbi(String mbi) throws GeneralSecurityException {
```
jonfulk, line 134: One warning said that the `GeneralSecurityException` was never thrown and wanted us to remove this.

dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/client/MockBlueButtonClient.java
```diff
@@ -56,7 +55,6 @@
     public static final String MULTIPLE_RESULTS_MBI = "0SW4N00AA00";
     public static final OffsetDateTime BFD_TRANSACTION_TIME = OffsetDateTime.ofInstant(Instant.now().truncatedTo(ChronoUnit.MILLIS), ZoneOffset.UTC);
     public static final OffsetDateTime TEST_LAST_UPDATED = OffsetDateTime.parse("2020-01-01T00:00:00-05:00");
-    public static final DateRangeParam BFD_LAST_UPDATE_RANGE = new DateRangeParam().setUpperBoundInclusive(new Date());
```
jonfulk, line 59: This offending constant was never used, so I removed it.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/TokenResource.java
```diff
@@ -61,7 +61,6 @@
     // This will be removed as part of DPC-747
     private static final String DEFAULT_ACCESS_SCOPE = "system/*.*";
     private static final Logger logger = LoggerFactory.getLogger(TokenResource.class);
-    private static final String ORG_NOT_FOUND = "Cannot find Organization: %s";
```
jonfulk, line 64: This was only used in this file once and was throwing an Error Prone warning because inline string formatting is prefered.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
```diff
@@ -60,7 +60,7 @@
      * a OperationOutcome resource is used.
      *
      * @param mbi to use
-     * @param headers
```
jonfulk, line 63: this was throwing a warning because the description was missing. Added generic description.

dpc-queue/src/main/java/gov/cms/dpc/queue/JobQueueCommon.java
```diff
@@ -69,8 +69,4 @@ protected JobQueueBatch createJobBatch(UUID jobID,
...
-        return batchSize;
-    }
```
jonfulk, line 75: This method was not used anywhere and was throwing a warning.

dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/client/MockBlueButtonClient.java
adhocEric: Any clue why it wasn't thrown?

jonfulk: This method is in the MockBlueButtonClient, which is only used for testing. It simply returns the value of an `mbi` from a Map containing test `mbi`s. For this reason, there was never a path in which an exception could be thrown and the code was unnecessary.

adhocEric: Gotcha

dpc-testing/src/main/java/gov/cms/dpc/testing/factories/FHIRGroupBuilder.java
```diff
@@ -77,6 +77,12 @@ public FHIRGroupBuilder withOrgTag(UUID orgId){
         theGroup.setMeta(meta);
         return this;
     }
+
+    public FHIRGroupBuilder withUUID(){
+        theGroup.setId(UUID.randomUUID().toString());
+        return this;
+    }
```
jonfulk, line 84: This method is needed to add a UUID to a Group when testing in order to avoid the `StackOverflow` error we were seeing [here](https://github.com/CMSgov/dpc-app/runs/1658948619?check_suite_focus=true#step:5:4935).

dpc-attribution/src/test/java/gov/cms/dpc/attribution/resources/v1/GroupResourceUnitTest.java
```diff
@@ -72,6 +72,7 @@ public void testCreateRosterHappyCase(){
 
         final Group group = FHIRGroupBuilder
                 .newBuild()
+                .withUUID()
```
jonfulk, line 75: see [comment below](https://github.com/CMSgov/dpc-app/pull/1167/files#r554191505)

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/service/LookBackServiceImpl.java
```diff
@@ -52,11 +57,13 @@ public String getPractitionerNPIFromRoster(UUID orgUUID, String providerOrRoster
...
     @Override
     @UnitOfWork(readOnly = true)
+    @SuppressWarnings("JdkObsolete") // Date class used by FHIR stu3 Period model
     public LookBackAnswer getLookBackAnswer(ExplanationOfBenefit explanationOfBenefit, UUID organizationUUID, String practitionerNPI, long withinMonth) {
         MDC.put(EOB_ID, explanationOfBenefit.getId());
-        Date billingPeriod = Optional.of(explanationOfBenefit)
+        YearMonth billingPeriod = Optional.of(explanationOfBenefit)
                 .map(ExplanationOfBenefit::getBillablePeriod)
                 .map(Period::getEnd)
+                .map(date -> YearMonth.from(date.toInstant().atZone(ZoneId.systemDefault()).toLocalDate()))
```
SalGutierrez, line 66: Can date be null?

jonfulk: It could be an empty Optional, which is why we have `.orElse(null)` on line 67. Do you see any way for that to cause an NPE still? I might be missing it.

SalGutierrez: Got it. 👍

SalGutierrez (approved):
LGTM

PR #1178, opened 2021-02-02
DPC-1069: Do not enumerate account info on password reset

Sun-Mountain:
### Fixes [DPC-1069](https://jira.cms.gov/browse/DPC-1069)

In our pre-ACT penetration test, the testers were able to tell whether a given email address was registered in the web app by trying to reset a password.

We should be able to overcome this data leakage by setting the Devise configuration to use paranoid mode.

### Proposed Changes

- [Use paranoid mode](https://github.com/heartcombo/devise/wiki/How-To:-Using-paranoid-mode,-avoid-user-enumeration-on-registerable)
- Create an automated test to prove that account enumeration is not happening in the future
- Retest all account capabilities, especially those sending emails. Make sure 1) everything still works and 2) there's no difference between reseting a password for an account that exists and one that doesn't exist

### Change Details

Turned on `config.paranoid = true` so that it gives a vague notice to the user that if the email exists in our system, we'll send them either a password reset email or another confirmation email.

As of right now, the code coverage alert is failing, but I don't think it's because of the changes I made. Currently the rails app has 94.66% coverage rating.

Screen Shot 2021-02-02 at 10 44 09 AM

### Security Implications

### Acceptance Validation

**Before**

Screen Shot 2021-02-02 at 10 17 52 AM

**After**

Screen Shot 2021-02-02 at 10 16 41 AM

### DIFF
--- a/dpc-web/config/initializers/devise.rb
+++ b/dpc-web/config/initializers/devise.rb
@@ -81,7 +81,7 @@ Devise.setup do |config|
   # It will change confirmation, password recovery and other workflows
   # to behave the same regardless if the e-mail provided was right or wrong.
   # Does not affect registerable.
-  # config.paranoid = true
+  config.paranoid = true
 
   # By default Devise will store the user in session. You can skip storage for
   # particular strategies by setting this option.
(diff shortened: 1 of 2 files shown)


dhgreene (approved):
⚡

PR #1181, opened 2021-02-05
DPC-1029 Clear Artifacts from Org at Beginning of Smoke Test

SalGutierrez:
### Fixes [DPC-1029](https://jira.cms.gov/browse/DPC-1029)

In a recent troubleshooting session we found that some failure modes for our smoke test leave artifacts in the database for our test org. When this happens, the next smoke test will also fail, because it's unable to insert objects with the same identifiers.

To prevent this from happening, at the beginning of the smoke test we should list all artifacts in the test org and delete them before continuing with the test.

### Proposed Changes
- Delete an organization before the smoke test starts. Deleting an organization will delete all of the artifacts associated with the org.

### Change Details

- Added pre-test step that will check if the test organization exists, if it does it will be deleted.
- Refactored and cleaned up smoke tests.

### Security Implications

### Acceptance Validation

Facts:
1. If an organization is not deleted due to a test failure the key will still exist.
2. A subsequent run would fail if a key with the same label is uploaded.

How issue was reproduced:
1. Ran a script that registers an org and uploads a key to API.
2. Ran smoke tests (before changes) and verified tests failed due to a duplicate key exception (simulating an org was not cleaned up).

How fix was verified:
1. Ran a script that registers an org and uploads a key to API.
2. Verified that test checks for existing org, if it does it gets deleted.

#### Script that registered and uploaded key
Screen Shot 2021-02-04 at 6 42 17 PM

#### Error reproduced
Screen Shot 2021-02-04 at 7 58 55 PM

#### Test startup (Org does not exists)
Screen Shot 2021-02-04 at 7 53 15 PM

#### Test startup (Org exists and gets deleted)
Screen Shot 2021-02-04 at 7 45 28 PM


jonfulk (approved):
LGTM!

PR #1186, opened 2021-02-10
DPC-1044 feat: Add endpoint to return org associated with the principle

jonfulk:
### Fixes [DPC-1044](https://jira.cms.gov/browse/DPC-1044)

Our documentation indicates that an endpoint exists `/Organization` that will return a user's related organization. This did not exist, and our only related endpoint required the org's internal id as a parameter, which many users do not necessarily have access to.

### Proposed Changes

- Create new endpoint to return the user's org at `/Organization` as indicated in the documentation
- Leave the existing `/Organization/{UUID}` endpoint in place for backwards compatibility
- Get the organization's UUID from the authenticated token to ensure that only the org of the authenticated user is returned

### Change Details

For now, there is only 1 org / user, so this is not technically a search endpoint. However, to stay consistent with the other similar endpoints, `/Organization` should return a `Bundle` of organizations as is the case for `/Provider`, `/Patient` , `/Group` and so on. It is not yet necessary to add a search method to `attribution/.../OrganizationResource` as we are currently only expecting a single Org ID to be in the token, meaning that org can be retrieved directly using the existing `fetchOrganization` method. That said, we still need to create and return a Bundle in the api method.

*NOTE: This will require an update to documentation which currently only shows a Resource being returned from this endpoint as opposed to a Bundle of Resources.*

### Security Implications

### Acceptance Validation

- `make ci-app` build passes
Screen Shot 2021-02-10 at 5 00 41 PM

- local smoke tests pass
Screen Shot 2021-02-10 at 5 11 41 PM

- request via postman returns what is expected
Screen Shot 2021-02-10 at 7 48 10 PM

### Feedback Requested

Any


SalGutierrez (approved):
LGTM 🚢 it

dhgreene (approved):
💯 Approved pending pipeline checks.

PR #1187, opened 2021-02-11
DPC-1071: Admin site separation: create new admin site

Sun-Mountain:
### Fixes [DPC-1071](https://jira.cms.gov/browse/DPC-1071)

This ticket kicks off the separation of the `internal_user` dashboard from the web app. We are creating a second Ruby on Rails app in the DPC repository in `dpc-admin`. This sets up the basic rails app structure and allows users to sign in via github.

### Proposed Changes

- Create `dpc-admin` in root `dpc-app`
- Create main layout and structure for admin portal
- Connect to existing web app database
- Users can log into a separate admin portal via github
- Tests for `internal_user` model and sign in.

### Change Details

- Had to change default `localhost` of `dpc-web` to 3500 so that one can run both admin dashboard and the web app simultaneously.
- Included instructions in both admin and web app readmes.

### Security Implications

### Acceptance Validation

- Users can log into a separate admin portal via github
- Create main layout and structure for admin portal

Screen Shot 2021-02-17 at 11 25 10 AM

Screen Shot 2021-02-17 at 11 24 49 AM

- Have tests pass for internal user sign in and internal user model

Screen Shot 2021-02-17 at 11 23 39 AM

- Create README to set up local admin dashboard

### Feedback Requested

Any


PatKelsh (changes requested):
We should install rubocop and run "rubocop --auto-gen-config" or address the issues it raises. Also the ruby version for this is 2.6.3 but in dpc-web it is 2.7.1, should these be the same version to keep things consistent?
I don't see too many tests in the spec directory, there probably should be a bunch of test, especially for the base search service.

Sun-Mountain:
> We should install rubocop and run "rubocop --auto-gen-config" or address the issues it raises. Also the ruby version for this is 2.6.3 but in dpc-web it is 2.7.1, should these be the same version to keep things consistent?
> I don't see too many tests in the spec directory, there probably should be a bunch of test, especially for the base search service.

Good catch on the ruby version! And I've updated that and added rubocop.

The spec tests for base search will come when I work on the user/org dashboards. Right now the modals (that use the base search) aren't operational because I was struggling with the javascript.

PatKelsh (approved):
Looks good to me

PR #1196, opened 2021-02-22
DPC-1112 Allow default expiration time for Client token creation

SalGutierrez:
### Fixes [DPC-1112](https://jira.cms.gov/browse/DPC-1112)

There is a small bug in GenerateClientTokens task where an empty expiration time is not being handled correctly. Passing in the expiration with in empty value results in a 400. Task should use the default expiration time when a blank value is passed but it tries to parse it.
### Proposed Changes
Ensure the GenerateClientToken task in API is handling empty values for expiration correctly.

### Change Details
The GenerateClientToken task was checking for the expiration parameter and using the default value if it was missing; however, the admin task was sending the expiration parameter, but with a blank value.

Added an extra check for the expiration value.

### Security Implications

### Acceptance Validation
I attempted to create a client token with an empty expiration before the fix (Generate Client Token Build #52) and that resulted in an error (notice the size of s3 file)
Screen Shot 2021-02-22 at 2 40 47 PM

After deploying this branch to DEV I ran the same admin task with the same parameters (Generate Client Token Build #53) and token was successfully generated and uploaded to s3.
Screen Shot 2021-02-22 at 3 52 48 PM


jonfulk (approved):
Thanks for catching that! LGTM

PR #1199, opened 2021-02-24
DPC-817 fix: Enforce explicit fhir+json Accept header, reject nulls and wildcards

jonfulk:
### Fixes [DPC-817](https://jira.cms.gov/browse/DPC-817)

Currently, for export requests, if the `Accept` header is not provided, the FHIRRequestFilter will default it to `*/*`, which is not compliant.

### Proposed Changes

For export requests:
- if the `Accept` header is missing or the value is null, return a 400
- if the `Accept` header is not of an acceptable FHIR type, return a 415

### Change Details

- update tests to check for these conditions: all of the FHIRRequestFilterTests needed to have content and accept headers set on them since both checks can return a 415. There were tests passing before because they passed the accept header check (which was not expected), but then they failed the content header check, throwing the expected error. By adding the correct headers that are not under test, we can control for these hidden failures.
- add a check in the FHIRRequestFilter to check for the absence of the `Accept` header when the uri includes `$export` and return a 400 if it is not there.
- add a condition to return a 415 if the `Accept` header value is a wildcard `*/*`.

### Security Implications

### Acceptance Validation

- `make ci-app` builds successfully and tests pass locally
Screen Shot 2021-02-24 at 1 15 43 PM

- smoke tests pass locally
Screen Shot 2021-02-24 at 8 24 16 PM

- deploy and testing on dev: Using Postman and cURL, the `$export` requests require the FHIR Accept header (and return 415 when not valid or when `*/*` is passed in as the value), but the other requests do not have this requirement. See the following screenshots for examples

✅  Not `$export` and `Accept: */*`
Screen Shot 2021-02-24 at 9 59 42 PM
------------
❌ `$export` and `Accept: */*`
Screen Shot 2021-02-24 at 9 50 08 PM
------------
✅ `$export` and `Accept: application/fhir+json
Screen Shot 2021-02-24 at 9 56 35 PM
------------

### Feedback Requested

Any

Note: both Postman and cURL appear to be automatically adding `Accept: */*` when an accept header is not explicitly added to the request. This makes it impossible to demo the 400 error where the accept header is missing or the value is null, but the FHIRRequestFilterTest tests do check for this case.


dhgreene (approved):
LGTM.  👏 for closer adherence to spec.

PR #1200, opened 2021-02-24
DPC-988 reduce BFD health checks.

SalGutierrez:
### Fixes [DPC-988](https://jira.cms.gov/browse/DPC-988)

The BFD team reports

... seeing about 280K requests to the BFD metadata endpoint from DPC in PROD and PROD-SBX, every 24 hours.

This is due to API and Aggregation's health endpoints performing deep health checks and hitting BFD during each request.

- DPC-API and DPC-Attribution have the BlueButtonClientModule registered and that contains the BlueButton Health Check.

- When we hit the /healthcheck endpoint on API or Aggregation we trigger a health check of the dependencies, BlueButton being one of them.
- We do not log the actual call to BlueButton but we do log the /healthcheck request that API or Aggregation receives, so that can be used to track the number of calls made to BlueButton's /metadata

Based on Splunk queries on PROD:
- DPC-API receives ~8,650 /healthcheck requests/hour or ~305,474 requests/day (Query)
- DPC-Aggregation receives ~2,896 /healthcheck requests/hour or ~102,512 requests/day (Query)
- Thats about 407,986 api/aggregation /healthcheck requests/day -> 407,986 BlueButton /metadata requests per day.

### Proposed Changes
3 solutions were proposed:

1. Reduce the number of health checks to API and Aggregation that the load balancers perform, this would then reduce the BFD health checks.
2. Have API and Aggregation check BFD health based on a schedule instead of being triggered by their /healthcheck endpoints and log any unhealthy checks. Splunk alerts would be created to catch those logs.
3. Completely remove the BFD health checks from API/Aggregation and create a new lambda that will poll BFD's /metadata on a schedule and log any unhealthy checks. Splunk alerts would be created to catch those logs.

We went with option 3 and will be implement it in 3 steps:

1. Remove the BFD health checks from API and leave them in Aggregation; this is a short-term solution that will satisfy this ticket by reducing the /metadata calls.
2. Work to create the BFD health check lambda has been planned.
3. Once the lambda is complete we will remove the BFD health checks from Aggregation

### Change Details

- Made the BFD health check registration configurable/optional in the BlueButtonClientModule
- By default it will be enabled for Aggregation and disabled for API

### Security Implications

### Acceptance Validation

Screen Shot 2021-02-24 at 10 15 54 AM


dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/BlueButtonClientModule.java
```diff
@@ -59,8 +59,11 @@ public void configure(Binder binder) {
             this.bbClientConfiguration = getConfiguration().getBlueButtonConfiguration();
         }
 
-        binder.bind(BlueButtonHealthCheck.class);
-
+        final boolean healthCheckEnabled = this.bbClientConfiguration.isRegisterHealthCheck();
+        if(healthCheckEnabled){
+            binder.bind(BlueButtonHealthCheck.class);
+        }
+        logger.info("Blue Button health checks are {}.", healthCheckEnabled ? "enabled" : "disabled");
```
dhgreene, line 66: _That's_ going to be a common message!

dhgreene (approved):
📉

dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/BlueButtonClientModule.java
SalGutierrez: It only get printed when the module is registered, so only on startup.

PR #1213, opened 2021-03-03
DPC-1136 add config management

whuang85:
### Fixes [DPC-1136](https://jira.cms.gov/browse/DPC-1136)

Need a way to manage configurations in go project.  

### Proposed Changes

* moved the code one level down into a `src` directory
* created a config directory at the root of the project (dpc-attribution or dpc-api) that holds yml config files
* create conf package that sets up viper and the config by first reading the base.yml then any env.yml (local.yml, test.yml, etc..) then a configuration file that can be defined via env variables
* add a missing tests for logger.go (bonus)

### Change Details

Will be using the approach of a base.yml config file and then if you define an env variable keyed `ENV`, the configuration will merge in the `${ENV}.yml` configs and if you define an env variable keyed `ADDITIONAL_CONFIG` (a env file path) then will merge those variables into the final config. The config also takes into consideration system variables that start with `DPC`

### Security Implications

### Feedback Requested

I'd like to use the init() function and grab additional directories to search for config files from an environment variable but during testing, the init() function gets called before the os.SetEnv so I can't get it to work.  Therefore I created a NewConfig func that takes in a list of directories to search for.  Please let me know if there is different way to approach this.


dhgreene (approved):
I'm loving the config tests, especially at this stage of the code!

dhgreene (approved):

SalGutierrez (approved):
To the moon. 🚀

PR #1219, opened 2021-03-04
DPC-1167 filter out unneeded fields

whuang85:
### Fixes [DPC-1167](https://jira.cms.gov/browse/DPC-1167)

When clients are passing in FHIR data to the api, we don't want to store all the various FHIR fields as it does not concern us therefore it could be a security risk if we did store data that we did not need. This PR is to address that by filtering (removing) the various data fields in a FHIR request that we don't care about.

### Proposed Changes

* Create a fhir filter middleware
* create a filter for org, practitioner, patient that only brings in data we do want
* added docker-compose files and docker file for convenience to start up either attribution or api locally

### Change Details

Added a new middleware that is used on create/update operations that will filter the fhir request by removing data that is not needed by the api

### Security Implications

### Feedback Requested

If a filter is not found for a resourceType, should we return an error or allow it thru? (Assuming downstream we have validators for the data)

If an error is thrown during unmarshalling, should we return an error or allow it thru?

Should this middleware be on the attribution router or api router? I argue that it should be on api since we don't want attribution to care or know about fhir


dpc-go/dpc-attribution/migrator/README.md
```diff
@@ -25,10 +25,10 @@ Assuming the last migration in the `/migrations` folder has an id of 41 a new mi
 *Note: The `Dockerfile` copies the files under /migrations to the image's own volume.*
 
 2) Run the migrations: 
- `docker run  -network host migrator -database "postgresql://postgres:dpc-safe@localhost"  up`
+ `docker run --network host migrator -path=/migrations/ -database=postgres://postgres:dpc-safe@localhost:5432/dpc_attribution_v2?sslmode=disable up`
  
 3) Rollback a migration: 
-  `docker run  -network host migrator -database "postgresql://postgres:dpc-safe@localhost"  down 4`
+  `docker run --network host migrator -path=/migrations/ -database=postgres://postgres:dpc-safe@localhost:5432/dpc_attribution_v2?sslmode=disable down 4`
```
SalGutierrez, line 31: Thanks. 🌮

SalGutierrez (approved):
LGTM 🌮 🌮 🌮

PR #1239, opened 2021-03-17
DPC-1198 Fixed job status percentage bug

SalGutierrez:
### Fixes [DPC-1198](https://jira.cms.gov/browse/DPC-1198)

During onboarding it was mentioned that the job status percentage would bounce around when checking the status. (Example 50% -> 30% -> 70%).

Possible cause:
In the Job Resource (API) we: * Add up all of the patient indexes (number of patients processed) for all batches.

Divide the previous result by the total number of patients.
Pretty straight forward and should work just fine assuming that the current patient index will never decrease for any job batch.But it looks like the assumption is incorrect and the patient index for a batch is set to NULL (treated as 0) once that batch is marked as completed.Example:
Current state of batches | Job Status request #1 # B1: (Patient Index = 9, Total = 10)

B2: (Patient Index = 3, Total = 10)
B3: (Patient Index = 2, Total = 5)
Status: 9+3+2 / 25 = 56% <-Correct statusCurrent state of batches | Job Status request #2 # B1: (Patient Index = NULL, Total = 10) COMPLETED

B2: (Patient Index = 5, Total = 10)
B3: (Patient Index = 3, Total = 5)
Status: 0+5+3 / 25 = 32% <-Incorrect, should be 10 + 5 + 3 / 25 -> 72%

### Proposed Changes

When computing the job status percentage use the batch size instead of the patient index for batches that are COMPLETE.

During implementation I also noticed that if there are no RUNNING batches in a job the job status will output "QUEUED: 0.00%" else it will compute the percentage and output "RUNNING X.XX%".

This means that if there is ever a job with batches that have been COMPLETED, batches that are QUEUED, but no RUNNING batches the output will be "QUEUED: 0.00%", Ideally it should say "RUNNING X.XX%" or at least "QUEUED X.XX%"

### Change Details

- Updated logic to compute percentage
- Returned "RUNNING: X.XX%" for jobs that are not complete, have no running batches, and have at least 1 complete batch.

### Security Implications

### Acceptance Validation

Unit tests.

### Feedback Requested
Thoughts on the logic and the response message for the special case mentioned.


dpc-queue/src/main/java/gov/cms/dpc/queue/models/JobQueueBatch.java
```diff
@@ -447,6 +447,16 @@ public void setUpdateTime() {
...
     }
 
+    /**
+     * Retrieve the number of patients that have been processed.
+     */
+    public int getPatientsProcessed() {
+        if(JobStatus.COMPLETED.equals(status)){
+            return getPatients() == null ? 0 : getPatients().size();
+        }
+        return getPatientIndex().orElse(-1)+1;
```
whuang85, line 457: doesn't this make it so it will never return 0 which seems valid since you can have zero patients processed?

SalGutierrez: if you have 0 patients processed then the patient_index will be null or -1, then this will return 0

whuang85 (approved):

PR #1251, opened 2021-03-30
DPC-1225 remove unchecked exception for bfd time regression

whuang85:
### Fixes [DPC-1225](https://jira.cms.gov/browse/DPC-1225)

We have a check that make sure that when request data from BFD, the time that BFD returns on subsequent requests are always after the initial request. Sometimes that doesn't always stay true because their servers could have a time drift issue. 

### Proposed Changes

Continue logging the issue but not kill our job

### Change Details

Continue logging the issue but not kill our job

### Security Implications


dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
```diff
@@ -248,7 +248,7 @@ private void checkBundleTransactionTime(Bundle bundle) {
             logger.error("Failing the job for a BFD transaction time regression: BFD time {}, Job time {}",
                     bfdTransactionTime,
                     transactionTime);
-            throw new JobQueueFailure("BFD's transaction time regression");
+//            throw new JobQueueFailure("BFD's transaction time regression");
```
jonfulk, line 251: I think we should just remove this line instead of commenting it out unless we plan to put it back in there pretty quickly, I guess.

jonfulk (dismissed):

DeirdreHolub (commented):
Please do these things:
1) Create (or ask Megan to create) an automatic monthly JIRA ticket to review this issue. The ticket can be assigned to Eric.
2) Refer to the JIRA ticket in the code in a comment with a brief comment (in case the JIRA ticket gets lost)
You can remove the throw statement entirely, as Jon suggested, and just leave the comment.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcher.java
```diff
@@ -248,7 +248,6 @@ private void checkBundleTransactionTime(Bundle bundle) {
             logger.error("Failing the job for a BFD transaction time regression: BFD time {}, Job time {}",
```
SalGutierrez, line 248: Maybe change the wording here, since we're not failing it anymore

SalGutierrez (approved):

megmcf:
Scheduled ticket has been created

PR #1264, opened 2021-04-14
DRAFT: DPC-1233 use npis

jonfulk:
### Security Implications


dhgreene (commented):
LGTM!  How careful do we need to be in our production environments to not deploy this change while jobs are running?

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/AggregationAppModule.java
```diff
@@ -101,14 +99,14 @@ public int provideJobTimeoutInSeconds() {
     }
 
     @Provides
-    LookBackService provideLookBackService(DPCManagedSessionFactory sessionFactory, ProviderDAO providerDAO, RosterDAO rosterDAO, OrganizationDAO organizationDAO, OperationsConfig operationsConfig) {
+    LookBackService provideLookBackService(DPCManagedSessionFactory sessionFactory, OrganizationDAO organizationDAO, OperationsConfig operationsConfig) {
         //Configuring to skip look back when look back months is less than 0
         if (operationsConfig.getLookBackMonths() < 0) {
             return new EveryoneGetsDataLookBackServiceImpl();
         }
         return new UnitOfWorkAwareProxyFactory("roster", sessionFactory.getSessionFactory()).create(LookBackServiceImpl.class,
```
whuang85, line 107: since we don't need the DAOs anymore, we can also delete this UnitOfWorkAwareProxyFactory and just return a new LookBackServiceImpl and passing in the OperationsConfig

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/dao/RosterDAO.java
```diff
@@ -22,6 +22,7 @@ public RosterDAO(DPCManagedSessionFactory factory) {
 
     /**
```
whuang85, line 23: I don't think any of the DAOs in the aggregation is being used anymore so it should be safe to delete them all  under aggregation and their associated tests

jonfulk:
> LGTM! How careful do we need to be in our production environments to not deploy this change while jobs are running?

That's a good question... I'm pretty sure this would interrupt jobs in progress.

dhgreene:
> I'm pretty sure this would interrupt jobs in progress.

And cause the remaining batches to fail, right?

jonfulk:
> > I'm pretty sure this would interrupt jobs in progress.
> 
> 
> 
> And cause the remaining batches to fail, right?
> 
> 

Correct. We would need all jobs to be completed before deploying in order to avoid errors.

jonfulk:
Note: `JobResource` uses the job batch's `providerID`, so that issue will still need to be addressed.

dhgreene (approved):
Thanks for the thorough testing in `dev`!  🚀

MrBilnon (approved):

PR #1268, opened 2021-04-20
DPC-1326: Add columns to JobQueueBatch to hold org and provider NPIs

jonfulk:
### Fixes [DPC-1326](https://jira.cms.gov/browse/DPC-1326)

This PR is phase 1 of a [refactor of the aggregation service](https://github.com/CMSgov/dpc-app/pull/1264) in order to get away from internal IDs for providers and rosters (groups) and instead to store the org NPI as well as the provider NPI on the job. This will be necessary as we move to V2 of DPC which will not require nor depend on permanent providers and groups.

In this first phase, we add the columns and begin storing the org and provider NPIs on the jobs so that, when we modify LookBackService to get these NPIs from the job instead of the objects, the values will be there on jobs in the queue. This will facilitate a smooth deployment of phase 2 without any downtime. (hat tip to @SalGutierrez for this clever approach).

We need to be able to use NPIs for providers and organizations instead of UUIDs as we transition to v2 of DPC

### Proposed Changes

- add columns for organization_npi and provider_npi
- add column for request_url as this is currently being reconstructed in the JobResource using the provider_id (usually really the rosterId) which we are going to stop using
- modify all calls to createJob and related tests to use NPIs instead of the provider internal id, but continue saving the org id as this will still exist in v2

### Change Details

- many tests were affected by this change
- export methods now require fetching objects (group, org, provider) in order to find the required NPIs for storing on the job
- lots of changes to testing, added mocks for fetching org and group in unit tests where required

### Security Implications

### Acceptance Validation

- [x] `make ci-app` builds and passes locally
Screen Shot 2021-04-20 at 10 32 07 AM

- [x] smoke tests pass locally
Screen Shot 2021-04-20 at 10 45 03 AM

- [x] deploy to dev successful
Screen Shot 2021-04-26 at 6 11 54 PM

- [x] postman tests successfully export data for a group/$export and patient/$everything endpoints without modification to the API contract

### Feedback Requested

Any, particularly are we missing anything when it comes to normal job completion and deployment of this and the next phase?


SalGutierrez (approved):
LGTM🛰

dhgreene (approved):
Thanks for the thorough testing in `dev`!  🚀

dpc-api/src/main/java/gov/cms/dpc/api/APIHelpers.java
```diff
@@ -93,4 +94,11 @@ public static String fetchRequestingIP(HttpServletRequest request) {
         }
         return ipAddress;
     }
-}
+
+    public static String fetchRequestUrl(HttpServletRequest request) {
```
MrMorie, line 98: NBD and no change requested, the the name of function and parameter makes it seem like its going to make a call in order to "fetch" the url (but its just extracting it)

jonfulk: good point. i followed the model of the existing methods `fetchTransactionTime` and `fetchRequestingIP`

MrMorie (approved):
LGTM, Good stuff 🚀

PR #1265, opened 2021-04-15
DPC-1073: Updating dpc-admin to use RAILS_RELATIVE_URL_ROOT

orenfromberg:
### Related to [DPC-1073](https://jira.cms.gov/browse/DPC-1073)

This branch adds the ability to set the relative URL root for dpc-admin so that it can be served from `dpc.cms.gov/admin/`.

The techniques used were taken from this gist: https://gist.github.com/ebeigarts/5450422

### Security Implications

- [ ] new software dependencies

- [ ] security controls or supporting software altered

- [ ] new data stored or transmitted

- [ ] security checklist is completed for this change

- [ ] requires more information or team discussion to evaluate security implications

- [ ] no PHI/PII is affected by this change

PR #1271, opened 2021-04-21
DPC-1073: Serving dpc-admin from /admin

orenfromberg:
### Related to [DPC-1073](https://jira.cms.gov/browse/DPC-1073)

This branch serves the admin app from the `/admin` relative root and supersedes #1265.

### Proposed Changes

Using `scope` to serve all routes under the `/admin` relative root.

### Change Details

The use of `scope` is documented [here](https://guides.rubyonrails.org/routing.html): 
image

### Security Implications

### Acceptance Validation

- [x] The admin app serves from `localhost:3000/admin` locally.

### Feedback Requested

All feedback is welcome.


dpc-admin/config/environments/test.rb
```diff
@@ -46,4 +46,7 @@
 
   # Raises error for missing translations.
   # config.action_view.raise_on_missing_translations = true
+
+  config.assets.prefix = '/admin_assets'
```
dhgreene, line 50: Is this (with the path change to the relative links to assets) still something we need?

orenfromberg: Yes because I am not sure how to serve both our route resources as well as our static assets from `/admin`.

orenfromberg: on second thought, this configuration is redundant because it is already specified in `dpc-admin/config/application.rb`, so it can be removed.

dpc-admin/config/initializers/health_check.rb
```diff
@@ -0,0 +1,69 @@
+HealthCheck.setup do |config|
```
orenfromberg, line 1: this was mostly copied from `dpc-web/config/initializers/health_check.rb`.

dpc-admin/config/application.rb
```diff
@@ -15,6 +15,7 @@ class Application < Rails::Application
     config.load_defaults 6.0
 
     # Add fonts to asset pipeline
+    config.assets.prefix = '/admin_assets'
```
dhgreene, line 18: I'm pretty sure we want this instead
```suggestion
    config.assets.prefix = '/admin/assets'
```

And then all of the `href="/admin_assets"` changes become `href="admin/assets"`.  If we _don't_ make this change, we have to configure our network systems to forward two paths to the admin app: `/admin` and `/admin_assets`.

orenfromberg: At first I tried serving the static assets from `/admin` but I couldn't get it to work. After some struggling I decided to add the following path patterns to the listener rule to accommodate:
https://github.com/CMSgov/dpc-ops/pull/328/files#diff-3b36f2c3ef0d4771518dfb7cb618a2754e8486290026465163af6008ca02ec24R157
Do you see any issues with this method going forward?

dpc-admin/app/services/api_client.rb
```diff
@@ -134,7 +134,7 @@ def http_request(request, uri)
 
     if use_ssl?
       http.use_ssl = true
-      http.verify_mode = OpenSSL::SSL::VERIFY_NONE unless verify_ssl_cert?
+      http.verify_mode = OpenSSL::SSL::VERIFY_PEER
```
orenfromberg, line 137: I believe this is the default value, but if not then this will explicitly enforce the client checking the validity of the server TLS cert.

dpc-admin/config/application.rb
orenfromberg: @dhgreene fixed!

orenfromberg:
related to https://github.com/CMSgov/dpc-ops/pull/328

dhgreene (approved):
🥳

PR #1278, opened 2021-04-27
DPC 1329 Fix: Pin Postgres image version to 11

lostpurplecow:
### Fixes [DPC-1329](https://jira.cms.gov/browse/DPC-1329)

The current deployed major version of Postgres in AWS for DPC is Postgres 11.  However, in the docker-compose stacks for dpc-app, we are pulling from postgres:latest (see here for example).

We need to ensure all references to pulling Postgres images from Dockerhub in dpc-app align with the version of Postgres that we are using in AWS.  This should be `postgres:11` image in Dockerhub.

### Proposed Changes

- Pin Postgres version to 11

### Security Implications

### DIFF
--- a/docker-compose.portals.yml
+++ b/docker-compose.portals.yml
@@ -9,7 +9,7 @@ services:
     image: redis:latest
 
   db:
-    image: postgres
+    image: postgres:11
     environment:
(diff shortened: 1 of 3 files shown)


dhgreene (approved):
🧊

PR #1284, opened 2021-05-03
DPC-1337 return operation outcome when no active patients

MrBilnon:
### Proposed Changes

* Check if the group retrieved contains any patient and if not, throw an error

### Change Details

* `fetchPatientMBIs` already only grabs active patients, so the only the change will be to see if the list of active patients is empty or not.  If it's empty, will return a WebApplicationException that gets converted to an OperationOutcome in our FHIR exception handlers.

### Security Implications


jonfulk (dismissed):
LGTM!

jonfulk (approved):

dhgreene (approved):
🐛 👞

PR #1285, opened 2021-05-03
DPC-1268: Invited users are able to complete account creation

Sun-Mountain:
### Fixes [DPC-1268](https://jira.cms.gov/browse/DPC-1268)

Branched off of [DPC-1265](https://jira.cms.gov/browse/DPC-1265)

As we create a new implementer portal, we want implementers to be able to invite multiple users to a single implementer. This will reduce burden for implementer turnover and improve security by making it less likely for users to share passwords for single accounts.

### Proposed Changes

- User can complete account creation by setting password and agreeing to terms of service
- Ensure bad actors cannot change `implementer` or `implementer_id` when sending a new user invitation
- Ensure if an invited user requests to resend invitation instructions, they receive the correct email

### Change Details

- Track who sends the invite to the user -- allowing to crosscheck `implementer`/`implementer_id` values before saving the account
- When a user requests to resend confirmation, if the user was invited they will receive another invitation token (via the Confirmations Controller)
- Refactored `create` in Invitation Controller
- Respective tests added or updated

### Security Implications

### Acceptance Validation

Screen Shot 2021-05-03 at 12 07 50 PM

Screen Shot 2021-05-03 at 12 07 39 PM

Screen Shot 2021-05-03 at 12 07 21 PM

Screen Shot 2021-05-03 at 12 04 29 PM

Screen Shot 2021-05-03 at 11 59 03 AM


dpc-impl/app/models/user.rb
```diff
@@ -21,6 +25,34 @@ def name
 
   private
 
+  # TODO: remove after connecting to API
+  def assign_implementer_id
+    self.implementer_id = SecureRandom.uuid if implementer_id.blank?
+  end
+
+  def check_impl
```
dhgreene, line 33: Does this run every time a user changes their name, email, or password?  The potential side effects seem heavy.

dhgreene: Also, with this logic in place we should _never_ change the implementer_id for a user, otherwise anyone they invited would also be changed.

Sun-Mountain: So would only on create make sense?

dpc-impl/db/structure.sql
```diff
@@ -0,0 +1,292 @@
...
+
+--
+-- Name: users; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.users (
+    id bigint NOT NULL,
+    first_name character varying NOT NULL,
+    last_name character varying NOT NULL,
+    implementer character varying NOT NULL,
```
dhgreene, line 107: Ok for now; any more `implementer`-related data and we'll need a new table.  😄

Sun-Mountain: I don't think I'll need any more data in the rails app.

dpc-impl/app/models/user.rb
```diff
@@ -21,6 +25,34 @@ def name
...
+    @user = self
+    @invite = User.where(id: @host).first
+    user_id = @user.implementer_id
+    invite_id = @invite.implementer_id
+    user_imp = @user.implementer
+    invite_imp = @invite.implementer
+
+    if user_id != invite_id
+      @user.implementer_id = invite_id
+      @user.save
```
dhgreene, line 47: Is this necessary in a `before_save` callback?

Sun-Mountain: Nope! Removed.

dhgreene (approved):
📨

PR #1295, opened 2021-05-13
DPC-1296 GET implementer/:id/org

MrMorie:
### Fixes [DPC-1296](https://jira.cms.gov/browse/DPC-1296)

Since we're no longer storing organization information in the webapp, we need to be able to call the API to view which organizations are associated with a given implementor.

AC:

The API has a webapp accessible /implementer endpoint
When the webapp calls the API, this endpoint returns a list of all of the organizations that have a relationship with the implementer (approved and pending)
API returns organization name, Type 2 NPI, UUID, and relationship status (approved, pending, etc)
This endpoint is not public; we will need to add this to a private service
### Proposed Changes
- Add Get /implementer/:id/org resource

### Change Details

- Added Get /implementer/:id/org resource

### Security Implications

### Acceptance Validation

Screen Shot 2021-05-26 at 5 57 49 PM


dpc-go/dpc-attribution/src/model/Implementer_org_status.go
```diff
@@ -0,0 +1,27 @@
...
+
+	// Pending relation not yet active
+	Pending
+
+	// Active Relation is active
+	Active
+)
+
+func (u ImplOrgStatus) String() string {
+	return [...]string{"Unknown", "Pending", "Active"}[u]
```
dhgreene, line 26: Tell me more about the value of an unknown state.  Do we have a point in time where we really don't know?  Would it make more sense to have just two values ("Inactive" and "Active"), with "Inactive" as the default?  At that point it could be a boolean type, perhaps.

dpc-go/dpc-attribution/migrator/migrations/4_create_implementer_org_relation_table.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE EXTENSION IF NOT EXISTS pgcrypto;
+CREATE TABLE "implementer_org_relation" (
```
dhgreene, line 2: Also plural?

dpc-go/dpc-attribution/src/v2/implementer_test.go
```diff
@@ -58,7 +58,7 @@ func (suite *ImplementerServiceTestSuite) TestPost() {
 	_ = faker.FakeData(&impl)
 	suite.repo.On("Insert", mock.Anything, mock.Anything).Return(&impl, nil)
 
-	req := httptest.NewRequest("POST", "http://example.com/foo", strings.NewReader(`{"name":"test-name"}`))
+	req := httptest.NewRequest("POST", "http://example.com/foo", strings.NewReader("{\"name\":\"test-name\"}"))
```
dhgreene, line 61: For consistency, or because it didn't work?

dpc-go/dpc-attribution/src/model/Implementer_org_status.go
MrMorie: We won't have such case.

I just recall watching some tutorials that recommend defining a "zero" value when using iota.

I believe if we were to define a variable of type Status without giving it a value it would default to the zero value (0) in this case and incorrectly assume it is Active or Pending.

MrMorie: I suppose we could also use the skip value to accomplish the same 🤔

dpc-go/dpc-attribution/src/v2/implementer_test.go
MrMorie: Don't recall, I'll give it a try without.

dpc-go/dpc-attribution/src/model/Implementer_org_status.go
dhgreene: If we have a boolean value, we don't need iota.  The column name could be a boolean `implementer_org_relations.active` and we'd write code like `if implOrgRel.Active`.

MrMorie: If we do boolean it would be much easier, I don't think we currently have a use case for another status but would we ever? like Denied?

dpc-go/dpc-attribution/src/v2/implementer_test.go
MrMorie: don't remember why i changed it, but I reverted

dpc-go/dpc-attribution/src/model/managed_org.go
```diff
@@ -0,0 +1,9 @@
+package model
```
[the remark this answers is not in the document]
MrMorie: I did! Why is it complaining!?

dhgreene (commented):
The example output in the PR description doesn't include the organization name.  Is that just a side effect of the data in the test database, or is it somehow being filtered?

MrMorie:
> The example output in the PR description doesn't include the organization name. Is that just a side effect of the data in the test database, or is it somehow being filtered?

@dhgreene the screenshot was outdated. I have updated the correct/current output.

dpc-go/dpc-attribution/src/v2/implementer_org.go
```diff
@@ -0,0 +1,256 @@
...
+			boom.Conflict(w, "relation already exists")
+			return
+		}
+	}
+
+	ior, err := ios.impOrgRepo.Insert(r.Context(), implID, org.ID, model.Active)
+	if err != nil {
+		log.Error("Failed to create Implementer org relation", zap.Error(err))
+		boom.BadData(w, err)
+		return
```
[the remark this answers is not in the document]
MrMorie: In sbx we automatically set it to active, but if we "reset" it to active it already exists (but in another state) then we would have to add extra logic to handle prod (what would the default be)?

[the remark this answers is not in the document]
MrMorie: Ignore my last comment, I confused auto creating an org with setting it to active. If I recall being active be default happens in sbx and prod.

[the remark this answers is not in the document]
MrMorie: re-activating would require adding extra logic to the repo for update operations. I propose we make that change once we have a PUT (or Patch) endpoint (or another way of "un-activating"). For now all relations will be "Active" and no way of updating the status.

dpc-go/dpc-attribution/src/v2/implementer_org.go
```diff
@@ -0,0 +1,256 @@
...
+			if err != nil {
+				log.Error("Failed to create new org", zap.Error(err))
+				boom.BadData(w, err)
+				return
+			}
+		} else {
+			log.Error("organization with provided NPI not found", zap.Error(err))
+			boom.NotFound(w, err)
+			return
+		}
```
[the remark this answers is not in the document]
MrMorie: done

dpc-go/dpc-attribution/src/v2/implementer_org.go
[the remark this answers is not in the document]
dhgreene: Fair enough. 👌

dhgreene (approved):
🏆

PR #1296, opened 2021-05-13
DPC-1254: Create a docker file for the implementer portal

Sun-Mountain:
### Fixes [DPC-1254](https://jira.cms.gov/browse/DPC-1254)

Create a docker file for implementer rails app.

### Proposed Changes

- Create Dockerfile
- `entrypoint.sh`
- docker-compose
- Update [Jenkinsfile.docker_build](https://github.com/CMSgov/dpc-ops/pull/371) in `dpc-ops`

### Change Details

- Set up `letter_opener_web` so emails can be accessed.
- Add default `ENV[STATIC_SITE_URL]` if none provided for tests in `dpc-impl`
- Changed port to `4000`

### Security Implications

### Acceptance Validation

Screen Shot 2021-05-13 at 12 51 37 PM

Screen Shot 2021-05-13 at 12 51 45 PM


Makefile
```diff
@@ -48,6 +52,7 @@ start-portals:
 	@docker-compose -p dpc-portals -f docker-compose.yml -f docker-compose.portals.yml up start_core_dependencies
 	@docker-compose -p dpc-portals -f docker-compose.yml -f docker-compose.portals.yml up start_web
 	@docker-compose -p dpc-portals -f docker-compose.yml -f docker-compose.portals.yml up start_admin
+	@docker-compose -p dpc-portals -f docker-compose.portals.yml up start_impl
```
dhgreene, line 55: Why don't we need `-f docker-compose.yml` here?  We have it in `make start-dpc`.
```suggestion
	@docker-compose -p dpc-portals -f docker-compose.yml -f docker-compose.portals.yml up start_impl
```

docker/postgres/postgres-multiple-db-init.sh
```diff
@@ -6,7 +6,7 @@ set -u
 function create_user_and_database() {
 	local database=$1
 	echo "  Creating user and database '$database'"
-	psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" <<-EOSQL
+	psql --username "$POSTGRES_USER" <<-EOSQL
```
dhgreene, line 9: Why are we not wanting to stop any more if we hit an error?

dpc-impl/config/routes.rb
```diff
@@ -24,5 +24,6 @@
   if Rails.env.development?
     require 'sidekiq/web'
     mount Sidekiq::Web, at: '/sidekiq'
+    mount LetterOpenerWeb::Engine, at: "/letter_opener" if Rails.env.development?
```
dhgreene, line 27: It's already inside an `if` block 😄 
```suggestion
    mount LetterOpenerWeb::Engine, at: "/letter_opener"
```

dpc-impl/db/structure.sql
```diff
@@ -11,7 +11,7 @@ SET row_security = off;
 
 SET default_tablespace = '';
 
-SET default_table_access_method = heap;
+-- SET default_table_access_method = heap;
```
dhgreene, line 14: Interesting that we were explicitly setting this to the default value.  If we don't want that statement, perhaps we should remove it?  Or was this just part of some troubleshooting?

dpc-portals-test.sh
```diff
@@ -32,6 +34,13 @@ echo "└───────────────────────
 docker-compose -p dpc-portals -f docker-compose.portals.yml run --entrypoint "rubocop" dpc_admin
 docker-compose -p dpc-portals -f docker-compose.portals.yml run --entrypoint "bundle exec rails spec" dpc_admin
 
+echo "┌──────────────────────────┐"
+echo "│                          │"
+echo "│  Running DPC Impl Tests  │"
+echo "│                          │"
+echo "└──────────────────────────┘"
+docker-compose -p dpc-portals -f docker-compose.portals.yml run --entrypoint "bundle exec rails spec" dpc_impl
```
dhgreene, line 42: The other Rails apps have rubocop here; is there a reason to not follow that precedent?

Makefile
orenfromberg: now we need it so we've added it back.

docker/postgres/postgres-multiple-db-init.sh
orenfromberg: we've restored this code.

dpc-impl/db/structure.sql
orenfromberg: I'm not sure how this file is generated but it seems like this SQL is from a different version than what we're running in docker, at least that was the error that we got. commenting this line out allowed this to run without error. @Sun-Mountain any ideas? If it isn't needed then we can remove it but we should know why the error happened.

dpc-portals-test.sh
orenfromberg: @Sun-Mountain should the rubocop work be done in a separate ticket?

orenfromberg (approved):
:shipit:

dpc-impl/db/structure.sql
dhgreene: Gotcha.  I remember that we were out of sync with the production database and recently fixed our development versions.  So this is a PG v12 statement that we can remove.

dhgreene (approved):
📦

PR #1299, opened 2021-05-18
DPC-1317: Create docker-compose file to run v1 aggregation with v2 API

jonfulk:
### Fixes [DPC-1317](https://jira.cms.gov/browse/DPC-1317)

In order to start a v1 job via DPC v2, some v1 services need to be running. We need an easy way to get the necessary services started.

### Proposed Changes

- add a docker-compose.v2 file in the root of the app with the necessary services
- utilize existing docker-compose files in dpc-go/dpc-api and dpc-go/dpc-attribution
- tie it all together with a `make` target to facilitate startup and shutdown 
- add a `make` target to migrate and seed the db (since i always have to lookup the commands)
- update readme with new make commands

### Change Details

- use a project name for all of the docker-compose commands in order to link them in a single network
- environment variables for the URL override `localhost` in the configs with the container names, which will probably only work for local development

### Security Implications

### Acceptance Validation

✅  `make start-v2` starts up necessary containers
Screen Shot 2021-05-20 at 9 09 52 PM

✅ `make seed-db` adds data to the database and allows a job to be created (some logs not included in screenshot for brevity)
Screen Shot 2021-05-20 at 9 10 05 PM
Screen Shot 2021-05-20 at 9 10 18 PM

✅ `make down-v2` shuts down all required containers
Screen Shot 2021-05-20 at 9 10 29 PM
Screen Shot 2021-05-20 at 9 10 39 PM

### Feedback Requested
Any


dpc-go/dpc-attribution/configs/base.yml
```diff
@@ -1,7 +1,7 @@
 db:
   url: postgresql://postgres:dpc-safe@localhost:5432/dpc_attribution_v2
-  urlV1: postgresql://postgres:dpc-safe@localhost:5432/dpc_attribution
-  queueUrl: postgresql://postgres:dpc-safe@localhost:5432/dpc_queue
+  urlV1: postgresql://postgres:dpc-safe@dpc-app_db_1:5432/dpc_attribution
```
MrBilnon, line 3: you can override this in environment variable of attribution docker-compose, I believe it will be:
`- DPC_DB_URLV1=postgresql://postgres:dpc-safe@dpc-app_db_1:5432/dpc_attribution` or maybe
`- DPC_DB_URL_V1=postgresql://postgres:dpc-safe@dpc-app_db_1:5432/dpc_attribution`

dhgreene (commented):
@orenfromberg and @msnook do you have some suggestions for the Docker networks in this PR?  Besides meaningful names ("proxy" isn't appropriate here), I'm concerned that there are multiple networks defined with the same name.  Is that actually an issue?

@jonfulk are you sure that we actually need multiple networks?  Unless we're carefully modeling public/private services I'd think it more likely we just need to explicitly specify one.

orenfromberg:
@dhgreene @jonfulk @msnook I think it's possible to avoid using a custom network by using the `-p` parameter on `docker-compose` so that all the containers have the same prefix. For example, see how `start-portals` works in the `Makefile`.

MrMorie (approved):
🐳

Makefile
```diff
@@ -74,6 +74,25 @@ down-dpc:
...
+	@docker-compose -p dpc-v2 -f docker-compose.yml -f docker-compose.v2.yml up start_core_dependencies
+	@docker-compose -p dpc-v2 -f docker-compose.yml -f docker-compose.v2.yml up start_api_dependencies
+	@docker-compose -p dpc-v2 -f docker-compose.yml -f dpc-go/dpc-attribution/docker-compose.yml up -d attribution2
+	@docker-compose -p dpc-v2 -f dpc-go/dpc-api/docker-compose.yml up -d api
+	@docker ps
+
+.PHONY: down-v2
+down-v2:
+	@docker-compose -p dpc-v2 -f docker-compose.yml -f dpc-go/dpc-attribution/docker-compose.yml down
+	@docker-compose -p dpc-v2 -f dpc-go/dpc-api/docker-compose.yml down
```
dhgreene, line 88: We can probably get away without `-f docker-compose.yml` here, since the previous line will take down the same dependencies.  I do wonder if we'll remember that when we edit this section later.  Do you mind adding it?

jonfulk: if i remove `-f docker-compose.yml` here, i get the following error: 

```╰─$ make down-v2
ERROR: Service 'migrator' depends on service 'db' which is undefined.
make: *** [down-v2] Error 1```

This is because the v2 migrator now depends on the v1 db (defined in the base docker-compose.yml) since we've moved all the db's to a single postgres instance.

jonfulk: @dhgreene or maybe you were talking about line 88 which used to have `-f docker-compose.yml` but i had removed it before you commented... sorry for the confusion!

PR #1308, opened 2021-05-27
revert change to `docker-compose.portals.yml`

jonfulk:
### Security Implications

### DIFF
--- a/docker-compose.portals.yml
+++ b/docker-compose.portals.yml
@@ -8,6 +8,14 @@ services:
   redis:
     image: redis:latest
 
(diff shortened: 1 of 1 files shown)


docker-compose.portals.yml
```diff
@@ -135,7 +143,7 @@ services:
     depends_on:
       - db
       - redis
-    command: db:5432
+    command: db:5434
```
msnook, line 146: This should be `5432` because it uses the container-exposed port and not the host-exposed port.

PR #1314, opened 2021-06-01
DPC-1249: handle export params _type and _since

jonfulk:
### Fixes [DPC-1249](https://jira.cms.gov/browse/DPC-1249)

Currently, the `_type`, and `_since` params are not being saved to the `job_queue_batch` record

### Proposed Changes

- use middleware to validate, extract, and store since and type param values to context and pass along to attribution
- pass validated values to attribution as query params
- save type and since values to the BatchRequest object when available, which is then passed to the Insert method for saving to the database
- while working on this ticket, I noticed some problems in the way error messages were being handled and returned. Added code to pass relevant error messages to api from attribution.
- refactoring for code climate acceptance after merge conflicts

### Change Details

The Since value needs to be nullable, so I changed the type of this value from `time.Time` to `sql.NullTime`. This required checking for a valid Time in NullTime object and passing that along when present to the insert statement.

### Security Implications

### Acceptance Validation
✅ Unit tests pass
Screen Shot 2021-06-09 at 9 02 57 PM
Screen Shot 2021-06-09 at 9 03 03 PM

✅ Postman request confirms since and type are saved correctly
Screen Shot 2021-06-09 at 9 06 59 PM
Screen Shot 2021-06-09 at 9 07 43 PM

### Feedback Requested

Any


dpc-go/dpc-api/src/middleware/constants.go
```diff
@@ -13,6 +13,16 @@ const (
...
+	// PatientString is the string constant for Patient
+	PatientString string = "Patient"
+	// EoBString is the string constant for ExplanationOfBenefit
+	EoBString string = "ExplanationOfBenefit"
+	// CoverageString is the string constant for Coverage
+	CoverageString string = "Coverage"
+	// AllResources is the list of allowed resources for export
+	AllResources string = "Patient,Coverage,ExplanationOfBenefit"
+	// SinceLayout is the time format for the since parameter
+	SinceLayout string = "2006-01-02T15:04:05-07:00"
```
dhgreene, line 25: What's your strategy for allowing the "+07:00" option in the spec?

jonfulk: Good question haha! Working on that now.

jonfulk: This was a good callout that warranted more tests. I have confirmed that `+07:00` will still work with this time format as long as the `+` is URL-encoded in the query params, as is the case with v1 of the API

dpc-go/dpc-api/src/middleware/context_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	req := httptest.NewRequest(http.MethodGet, "http://www.example.com/Group/some-id$export?_type=Coverage,ExplanationOfBenefit&_since=2012-01-02T12:12:12-05:00", nil)
+	res := httptest.NewRecorder()
+
+	e := ExportSinceParamCtx(nextHandler)
+	e.ServeHTTP(res, req)
+	assert.Equal(suite.T(), "2012-01-02T12:12:12-05:00", since)
+}
+
+func (suite *ContextTestSuite) TestExportSinceParamPlusTZ() {
```
jonfulk, line 78: @dhgreene when the `+` is not encoded, we will return a 400 response with a message that we could not parse the _since param. That error response will be tested in another test suite.

dpc-go/dpc-api/src/client/attribution_client.go
```diff
@@ -132,9 +129,24 @@ func (ac *AttributionClient) Export(ctx context.Context, resourceType ResourceTy
...
+	if errMsg != "" {
+		return nil, errors.Errorf(errMsg)
+	}
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		return nil, errors.Errorf("Failed to start job for %s/%s", resourceType, id)
+	}
 	return body, nil
 }
 
+func checkForErrorMsg(body []byte) string {
```
MrBilnon, line 142: Why can’t the status code be used?

MrMorie: I think the error message is being proxied because we don't know exactly what went wrong since we do not have application error/status codes. So a status code of 400 does not really give you munch info of what actually went wrong (does group not exist or is it empty).

jonfulk: Exactly. The message sent could be saying that the group was not found or that there are no patients in that group. The status code doesn't give us enough info.

MrBilnon: But that should have been logged in attribution and you can find that message using the correlation id?

dpc-go/dpc-attribution/src/service/v1/job.go
```diff
@@ -37,14 +49,11 @@ func NewJobService(pr repository.PatientRepo, jr repository.JobRepo) service.Job
...
 	log := logger.WithContext(r.Context())
-
-	groupID := util.FetchValueFromContext(r.Context(), w, middleware.ContextKeyGroup)
-	orgID := util.FetchValueFromContext(r.Context(), w, middleware.ContextKeyOrganization)
-	types := "Patient,Coverage,ExplanationOfBenefit"
-	requestIP := r.Header.Get(middleware.FwdHeader)
-	requestURL := r.Header.Get(middleware.RequestURLHeader)
-
-	groupNPIs, err := js.pr.GetGroupNPIs(r.Context(), groupID)
+	ei := buildExportInfo(w, r)
```
MrBilnon, line 52: are you doing this just to get around code climate?

jonfulk: Technically, yes, but i do prefer to split out work as much as possible so that functions have (ideally) just one job. If there's a better way to reduce complexity in this function, I'm all ears!

MrBilnon (approved):

PR #1318, opened 2021-06-03
fix go lint and go sec stuff

MrBilnon:
No ticket here, just fixing some linting and gosec stuff caught by make test


jonfulk (approved):
👍

PR #1312, opened 2021-05-28
Update DB port binding in docker-compose to simplify local development

msnook:
As a follow-up to [this PR](https://github.com/CMSgov/dpc-app/pull/1310), there is a desire to update additional docker-compose configurations to ensure all Makefile targets expose a static port for the DB (as opposed to a random, ephemeral port).  

### Change Details

Updated the docker-compose configurations so that the Makefile targets demonstrate the following behavior:

- `make ci-app` exposes 5432 to host
- `make ci-portals` exposes 5434 to host 
- `make start-app` exposes 5432 to host 
- `make start-local` exposes 5432 to host
- `make start-local-api` exposes an 5432 to host
- `make start-portals` exposes 5434 to host
- `make start-dpc` exposes 5434 to host 
- `make start-v2` exposes 5432 to host

### Security Implications

- [ ] new software dependencies
- [ ] security controls or supporting software altered
- [ ] new data stored or transmitted
- [ ] security checklist is completed for this change
- [ ] requires more information or team discussion to evaluate security implications
- [x] no PHI/PII is affected by this change

### Acceptance Validation

To ensure these changes do not cause port clashing in Jenkins, this feature branch was built successfully [here](https://management.dpc.cms.gov/job/DPC%20-%20Docker%20Build/1675/).

### Feedback Requested

Please review.

PR #1319, opened 2021-06-04
Fix dpc-go/dpc-attribution/docker-compose.yml

jonfulk:
Following changes made in [this PR](https://github.com/CMSgov/dpc-app/pull/1310), `make start-v2` was no longer exposing the port for the postgres container. According to the logs, the `dpc-go/dpc-attribution/docker-compose.yml` command was recreating the db since the migrator depended on it. This caused an error when trying to connect to the db via Postico or by using the `make seed-db` command:

```
could not connect to server: Connection refused
	Is the server running on host "localhost" (::1) and accepting
	TCP/IP connections on port 5432?
could not connect to server: Connection refused
	Is the server running on host "localhost" (127.0.0.1) and accepting
	TCP/IP connections on port 5432?
```

```
org.postgresql.util.PSQLException: Connection to localhost:5432 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.
```

### Proposed Changes

- add back in `dpc-go/dpc-attribution/docker-compose.yml` the `db` service, specifying ports `5432:5432`

### Security Implications

### Acceptance Validation

With this change, I was able to successfully start the application using `make start-v2` and to seed the db using `make seed-db`. I was also able to connect to the db in Postico and view and manipulate the data there.

Screen Shot 2021-06-04 at 3 54 17 PM

Screen Shot 2021-06-04 at 3 54 37 PM

Screen Shot 2021-06-04 at 3 54 47 PM

confirmed: Docker images still build successfully
Screen Shot 2021-06-04 at 4 24 27 PM

### Feedback Requested

Any

### DIFF
--- a/dpc-go/dpc-attribution/docker-compose.yml
+++ b/dpc-go/dpc-attribution/docker-compose.yml
@@ -1,6 +1,9 @@
 version: "3"
 
 services:
+  db:
+    ports:
+      - "5432:5432"
   migrator:
     image: migrate/migrate
     depends_on:
--- a/dpc-go/dpc-attribution/docker-compose.yml
+++ b/dpc-go/dpc-attribution/docker-compose.yml
@@ -1,9 +1,7 @@
 version: "3"
 
 services:
-  db:
-    ports:
-      - "5432:5432"
+
   migrator:
     image: migrate/migrate
     depends_on:
--- a/dpc-go/dpc-attribution/docker-compose.yml
+++ b/dpc-go/dpc-attribution/docker-compose.yml
@@ -1,7 +1,6 @@
 version: "3"
 
 services:
-
   migrator:
     image: migrate/migrate
     depends_on:
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ down-dpc:
 start-v2: secure-envs
 	@docker-compose -p dpc-v2 -f docker-compose.yml -f docker-compose.v2.yml up start_core_dependencies
 	@docker-compose -p dpc-v2 -f docker-compose.yml -f docker-compose.v2.yml up start_api_dependencies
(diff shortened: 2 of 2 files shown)


dpc-go/dpc-attribution/docker-compose.yml
```diff
@@ -1,6 +1,9 @@
 version: "3"
 
 services:
+  db:
```
msnook, line 4: We already specify the explicit DB service and port mapping for V2 [here](https://github.com/CMSgov/dpc-app/blob/master/docker-compose.v2.yml#L8-L10).
Would it make sense, to reduce redundancy, to instead chain `-f docker-compose.v2.yml` in the `Makefile` targets for V2 [here](https://github.com/CMSgov/dpc-app/blob/master/Makefile#L81) and [here](https://github.com/CMSgov/dpc-app/blob/master/Makefile#L82)?

jonfulk: I believe we tried that before but ran into issues with the context of the commands. I'll give it another try in case we have since resolved that issue.

msnook: But would the changes that were made [here](https://github.com/CMSgov/dpc-app/pull/1312/files) ensure that this strategy works?  Was your attempt performed before or after this change?

jonfulk: Before. Definitely worth looking into your suggestion. Thanks!

jonfulk: @msnook thanks again! [This](https://github.com/CMSgov/dpc-app/pull/1319/commits/e72580af5e8f0a89e75211f56465369e7f03581a) worked.

msnook (approved):
👍

PR #1332, opened 2021-06-17
DPC-1488: Add BFD Client to DPC Attribution

jonfulk:
### Fixes [DPC-1488](https://jira.cms.gov/browse/DPC-1488)

This is the first step of completing DPC-1488. In order to add in the transaction time for a job, we need to be able to make a request to BFD. This requires a BFD client. Since BCDA already has such a client written in Go, it is most efficient to use that as the starting point.

### Proposed Changes

- updated BCDA client files and tests for DPC environment
- added encrypted versions of the BFD certs in a `shared_files/encrypted` directory
- using existing `.vault_password` and `ops/scripts/secrets` script, create a `make` target to easily temporarily decrypt the  certificates for local use.

### Security Implications

While no changes are being made to the security controls, it is worth noting that this change stores an encrypted version of the certificates in the repo. There are no decrypted copies committed to the repo, and this is a process we are already using in DPC and in BCDA.

### Acceptance Validation

All unit tests pass
Screen Shot 2021-06-17 at 4 23 24 PM

### Feedback Requested

- Is there any way to make the certificate decryption process easier using existing tools?
- Is it ok to use the same test data as BCDA for the BFD unit tests, or should we customize it?
- Anything else?


dhgreene (commented):
This is a _lot_ of code we won't be using at the moment, and don't plan to use later. How would you whittle it down to the minimum necessary?

dpc-go/dpc-attribution/src/client/bfd_client.go
```diff
@@ -0,0 +1,418 @@
...
+	"github.com/cenkalti/backoff/v4"
+	"github.com/go-chi/chi/middleware"
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"golang.org/x/crypto/pbkdf2"
+)
+
+//const syntheticBeneID = "-19990000000001"
```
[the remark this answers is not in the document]
jonfulk: @dhgreene I saved that for the `TransactionTime` work, but it's not needed for this PR. I'll remove it. Thanks!

As for the unused code, I was leaning on @DeirdreHolub 's suggestion that more is better than less in this situation. I think I'd like to discuss this more "in person". Thanks!

Makefile
```diff
@@ -142,6 +142,12 @@ maven-config:
 	@: > ./.mvn/maven.config
 	@while read line;do echo "-D$${line} " >> ./.mvn/maven.config;done < ./ops/config/decrypted/local.env
 
-.PHONE: unit-tests
```
MrMorie, line 145: ☎️

dpc-go/dpc-attribution/src/conf/config.go
```diff
@@ -74,3 +75,25 @@ func GetAsInt(key string, dv ...int) int {
 func Get(key string) interface{} {
 	return config.Get(key)
 }
+
+// SetEnv is a public function that adds key values into conf. This function should only be used
+// either in this package itself or testing. Protect parameter is type *testing.T, and it's there
+// to ensure developers knowingly use it in the appropriate circumstances.
+// This function will most likely become a private function in later versions of the package.
+func SetEnv(protect *testing.T, key string, value interface{}) {
```
MrBilnon, line 83: where is this being used and what is it being used for?  Sorry, probably in one of the files that github says too big to diff.

jonfulk: That's being used in several of the tests. This came from BCDA along with the client.
Screen Shot 2021-06-22 at 10 42 12 AM

dhgreene (approved):
Thanks for being so faithful to the original code.

PR #1338, opened 2021-06-23
DPC-1537: Set Transaction Time on Job Queue Batch records

jonfulk:
### Fixes [DPC-1537](https://jira.cms.gov/browse/DPC-1537)

When starting a job, we need to set the current BFD TransactionTime on the job.

### Proposed Changes

- make request to BFD to get the current transaction time
- save that transaction time on the job queue batch record

### Change Details

Now that the job service is using the bfd client, the bfd cert files are being accessed from different levels of subdirectories. This required constructing an absolute path to the filename instead of using the existing relative paths in the config. To do this, we first fetch and save the absolute path to the `decrypted` directory inside the config and save that to a variable called `DecryptedDir`. Then, wherever we access the certs, we combine this `DecryptedDir` with the file name from the config, including the tests. Some fake certs in the testdata folder also needed to be moved to the decrypted directory for ease of testing.

In this PR, we also increase the code climate cognitive complexity threshold from 9 to 11 in order to accommodate the inevitable complexity of our job Export method. 

### Security Implications

### Acceptance Validation

✅  unit tests pass
Screen Shot 2021-06-23 at 11 41 57 AM

✅  postman request confirms that the transaction time is set on a new job
Screen Shot 2021-06-23 at 11 27 49 AM
Screen Shot 2021-06-23 at 11 25 36 AM

### Feedback Requested

Any


MrMorie (approved):
🚀

PR #1352, opened 2021-07-07
DPC-1440 Feature: Proxy implementer organization POST request to attribution

charmcitygavin:
### Fixes [DPC-1440](https://jira.cms.gov/browse/DPC-1440)

Proxies a POST request to attribution service in order to establish an implementer/org relationship (and create the org if it doesn't already exist).

### Proposed Changes

* Adds `/Implementer/{implementerID}/org` POST endpoint to API
* Proxies request to attribution service and returns the new implementer/org relationship object

### Change Details

We also added a `ImplementerCtx` to `context.go` in order to extract the `ImplementerID` from the chi url param and set it into the request context.

### Security Implications

### Acceptance Validation

Tested via Postman.

Screen Shot 2021-07-07 at 3 20 29 PM
### Feedback Requested

Screen Shot 2021-07-07 at 3 20 00 PM 

Any and all feedback, please. If I'm doing anything ridiculous, or if you know of a better way, please call it out! 😄


dpc-go/dpc-api/src/apitest/helper.go
```diff
@@ -250,6 +250,11 @@ const ImplJSON = `{
     "name": "foo"
 }`
 
+// ImplOrgJSON is an Implementer/Org JSON string for testing purposes
+const ImplOrgJSON = `{
+    "npi": "0123456789"
+}`
```
dhgreene, line 256: This is ok, but better would be to follow [this pattern](https://github.com/CMSgov/dpc-app/blob/master/dpc-testing/performance/pkg/dpc/dpc.go#L116-L119) (using [this library](https://github.com/CMSgov/dpc-app/blob/5618e9dbaaab5dac8c4e2b3fcb1f6e2c5f9fb5e6/dpc-testing/performance/pkg/dpc/dpc.go#L12))

dpc-go/dpc-api/src/router/router.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 // NewDPCAPIRouter function that builds the router using chi
-func NewDPCAPIRouter(oc v2.Controller, mc v2.ReadController, gc v2.Controller, dc v2.FileController, jc v2.JobController, ic v2.Controller) http.Handler {
+func NewDPCAPIRouter(oc v2.Controller, mc v2.ReadController, gc v2.Controller, dc v2.FileController, jc v2.JobController, ic v2.Controller, implOrg v2.Controller) http.Handler {
```
dhgreene, line 17: We're probably at the point where we should be passing a struct

dpc-go/dpc-api/src/v2/implementer_org.go
```diff
@@ -0,0 +1,70 @@
...
+		log.Error("Failed to write data to response", zap.Error(err))
+		fhirror.ServerIssue(r.Context(), w, http.StatusInternalServerError, "Failed to save implementer/org relationship")
+	}
+}
+
+// Export function is not currently used for ImplementerOrgController
+//goland:noinspection GoUnusedParameter
+func (ioc *ImplementerOrgController) Export(w http.ResponseWriter, r *http.Request) {
+	w.WriteHeader(http.StatusNotImplemented)
+}
```
dhgreene, line 52: Are we putting this here to satisfy an interface? `Export()` maps to `GET /Group/[id]/$export` and nothing else--especially not non-FHIR resources.

dpc-go/dpc-api/src/apitest/helper.go
charmcitygavin: Thanks for pointing out that library! I've modified `helper.go` to use it instead of static JSON. 👍🏻

dpc-go/dpc-api/src/v2/implementer_org.go
charmcitygavin: Yes, it's to satisfy the `Client` interface in `attribution_client.go`. @MrMorie and I talked about how `attribution_client` might be expanding beyond its original scope, so perhaps it's time to create a new client. 🤔

dpc-go/dpc-api/src/router/router.go
charmcitygavin: @dhgreene Good call. Something like [this](https://github.com/CMSgov/dpc-app/pull/1352/commits/b35e72ad9392c34771fb330151eaaafa31ed2eb0)?

dhgreene (approved):
📯

PR #1357, opened 2021-07-08
DPC-1415: Return FHIR R4 for V2 Jobs

jonfulk:
### Fixes [DPC-1415](https://jira.cms.gov/browse/DPC-1415)

As part of our move to DPC V2, we will be returning R4 data from the V2 api. Since we are still using the same aggregation engine, modifications need to be made there in order to return the correct format for each type of export request.

### Proposed Changes

- add R4 FHIR context
- create a V2 BFD client that returns R4
- Replace HAPI-FHIR client's FHIR-version-dependent `ResourceType` enum with a custom one (results in cascading challenges throughout the app)
- Add `isV2` property to `JobQueueBatch`
- When processing a job, determine which BFD client to use by checking the job's `isV2` property
- Add V2 versions of: JobBatchProcessor, ResourceFetcher, and ResourceWriter
- Add V2 test data
- Update tests for V2 test data
- Add tests for V2 job processing

### Change Details

The 2 biggest challenges in this PR were 1) providing 2 different FHIRContexts at the same time and 2) determining the resource type without being constrained by a particular FHIR version. For the former, some magical R4Configuration solved the problem along with using named variables for `fhirContextR4`. For the latter, the custom enum was the solution. The custom enum is a fairly temporary solution until we fully deprecate V1 of the api. This made relaying on the enum's `getPath()` function for checking types an easier solution to implement for the sake of time when a better yet more time-consuming solution would have been to create a mapping between each FHIR-typed enum and the custom one and checking that type in each place where the enum is used.

### Security Implications

### Acceptance Validation

- all unit tests pass

Screen Shot 2021-08-04 at 8 20 29 AM

- successfully used Postman to create a group, start a job, check the status, and download the data all in V2, which resulted in the test R4 data linked here: [r4 responses.zip](https://github.com/CMSgov/dpc-app/files/6927576/r4.responses.zip)

Screen Shot 2021-08-03 at 6 28 21 PM

Screen Shot 2021-08-03 at 6 28 14 PM

### Feedback Requested

Do you see anything else that needs to be tested that isn't?
Any other feedback welcome


dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/JobBatchProcessorV2.java
```diff
@@ -0,0 +1,212 @@
...
+                .flatMap(Flowable::fromIterable);
+    }
+
+    private Map<String, String> buildHeaders(JobQueueBatch job) {
+        Map<String, String> headers = new HashMap<>();
+        headers.put(HttpHeaders.X_FORWARDED_FOR, job.getRequestingIP());
+        headers.put(Constants.BFD_ORIGINAL_QUERY_ID_HEADER, job.getJobID().toString());
+        if (job.isBulk()) {
+            headers.put(Constants.BULK_JOB_ID_HEADER, job.getJobID().toString());
+            headers.put(Constants.BULK_CLIENT_ID_HEADER, job.getProviderNPI());
```
dhgreene, line 122: What will this actually send? Is this the provider organization NPI, or if not do we need to substitute it for this?

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/ResourceFetcherV2.java
```diff
@@ -0,0 +1,255 @@
...
+    private void checkBundleTransactionTime(Bundle bundle) {
+        if (bundle.getMeta() == null || bundle.getMeta().getLastUpdated() == null) return;
+        final var bfdTransactionTime = bundle.getMeta().getLastUpdated().toInstant().atOffset(ZoneOffset.UTC);
+        if (bfdTransactionTime.isBefore(transactionTime)) {
+           // See BFD's RFC0004 for a discussion on why this type error may occur.
+           // Note: Retrying the job after a delay may fix this problem.
+            logger.error("Failing the job for a BFD transaction time regression: BFD time {}, Job time {}",
+                    bfdTransactionTime,
+                    transactionTime);
+            throw new JobQueueFailure("BFD's transaction time regression");
```
dhgreene, line 252: We determined to make this a warning in v1 (backlog ticket still outstanding). Let's start out right in v2 and just log the condition.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/JobBatchProcessorV2.java
MrBilnon: from v2, this will be a comma separated list of provider npis, not sure what we should change it to as the org npi is not unique enough to determine between different jobs among an org.

dhgreene: How about the Group UUID?

dhgreene: Comma-separated provider NPIs is fine for the moment

MrBilnon: oh yea group uuid sounds like a good one to use, only problem is I don't think we have it easily accessible here.  The only place where we would store a group id would be the request url and we would need to parse it out.  Can we just send the whole request url as the header value?

jonfulk: So do we want to stick with the list of providers or send the full request url?

jonfulk: we'll go with the full request url. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer), it looks like the url can be a header value without any escaping, if I'm reading this correctly.

MrBilnon (approved):

PR #1370, opened 2021-07-21
DPC-1438 Add Mutual TLS to GO DPC-API

MrMorie:
### Proposed Changes
DPC's v2 API service does not currently require authentication. We need to use mutual TLS to make that happen.

This PR ads Mutual TLS (MTLS) to DPC-API administrative tasks

### Change Details
- Added mutual to GO DPC API
- Server verifies client certificate
- Added flag to disable MTLS for local testing purposes. (Enabled by default)
- SAN Check: there is a new env variable in attribution called "ALLOWED_CLIENTS", this takes in a list of allowed names. At lease one of the allowed names must be included in the clients SAN extension found in certificate.starts up.

### Security Implications

No new dependencies

Yes, added MTLS

No new data stored, no interactions with new systems.

- [x] security checklist is completed for this change
SIA: https://confluence.cms.gov/pages/viewpage.action?pageId=490380420

Not at this time

### Request to API admin without a certificate
Screen Shot 2021-08-04 at 12 32 35 PM
Screen Shot 2021-08-04 at 12 32 25 PM

### Request to API admin with wrong certificate (SAN does not include an allowed DNS)
Screen Shot 2021-08-04 at 12 51 43 PM

### Request to API with valid certificate
Screen Shot 2021-08-04 at 1 39 36 PM


MrBilnon (approved):

MrBilnon (approved):

MrBilnon (dismissed):

dpc-go/dpc-api/src/service/server.go
```diff
@@ -46,7 +51,76 @@ func (s *Server) Serve(ctx context.Context) error {
...
+	}
+
+	fmt.Printf("Starting secure %s on port %d\n", s.name, s.port)
+	if err := s.server.ListenAndServeTLS("", ""); err != nil {
+		logger.WithContext(ctx).Fatal(fmt.Sprintf("Failed to start secure server: %s", s.name), zap.Error(err))
+	}
 	return nil
 }
+
+func getClientValidator(helloInfo *tls.ClientHelloInfo, cerPool *x509.CertPool) func([][]byte, [][]*x509.Certificate) error {
```
orenfromberg, line 78: This function doesn't seem necessary because the closure that it returns only checks the DNS names in the SAN field of the server certificate and not any of the client certificate information.

Even if it return a closure that did check the client certificate, it still wouldn't be necessary if the API certificate was used to sign the portals client key and cert because the handshake would only succeed in that situation.

dpc-go/shared/make-certs.sh
```diff
@@ -0,0 +1,21 @@
...
+#Generate API Cert
+openssl genrsa -out local.api.key 2048
+openssl req -new -key local.api.key -days 365 -out local.api.csr -subj "/C=US/ST=CA/L=LosAngeles/O=ACME API/CN=localhost"
+openssl x509  -req -in local.api.csr -CA local.ca.crt -CAkey local.ca.key -out local.api.crt -days 365 -sha256 -CAcreateserial  -extfile <(printf "subjectAltName=DNS:localhost,DNS:local.api.dpc.cms.gov")
+rm local.api.csr
+
+#Generate portal Cert
+openssl genrsa -out local.portal.key 2048
+openssl req -new -key local.portal.key -days 365 -out local.portal.csr -subj "/C=US/ST=CA/L=LosAngeles/O=ACME portal/CN=localhost"
+openssl x509  -req -in local.portal.csr -CA local.ca.crt -CAkey local.ca.key -out local.portal.crt -days 365 -sha256 -CAcreateserial -extfile <(printf "subjectAltName=DNS:localhost,DNS:local.portal.dpc.cms.gov")
```
orenfromberg, line 20: if we want to ensure that the only the portal can be a client to the API then should we use the API certificate and key here (`local.api.crt` and `local.api.key`) to sign the portals certificate?

dpc-go/shared/make-certs.sh
```diff
@@ -0,0 +1,21 @@
...
+#Generate Attribution Cert
+openssl genrsa -out local.attribution.key 2048
+openssl req -new -key local.attribution.key -days 365 -out local.attribution.csr -subj "/C=US/ST=CA/L=LosAngeles/O=ACME Attr/CN=localhost"
+openssl x509  -req -in local.attribution.csr -CA local.ca.crt -CAkey local.ca.key -CAcreateserial -out local.attribution.crt -days 365 -sha256 -extfile <(printf "subjectAltName=DNS:localhost,DNS:local.attribution.dpc.cms.gov")
+rm local.attribution.csr
+
+#Generate API Cert
+openssl genrsa -out local.api.key 2048
+openssl req -new -key local.api.key -days 365 -out local.api.csr -subj "/C=US/ST=CA/L=LosAngeles/O=ACME API/CN=localhost"
+openssl x509  -req -in local.api.csr -CA local.ca.crt -CAkey local.ca.key -out local.api.crt -days 365 -sha256 -CAcreateserial  -extfile <(printf "subjectAltName=DNS:localhost,DNS:local.api.dpc.cms.gov")
```
orenfromberg, line 14: if we want to ensure that the only the API can be a client to the attribution service then should we use the attribution certificate and key here (`local.attribution.crt` and `local.attribution.key`) to sign the API certificate?

dpc-go/dpc-api/src/service/server.go
```diff
@@ -46,7 +51,76 @@ func (s *Server) Serve(ctx context.Context) error {
 		fmt.Printf("Starting UNSECURE %s on port %d\n", s.name, s.port)
 		return s.server.ListenAndServe()
 	}
-	//TODO MTLS logic will go here once it is merged in.
-	logger.WithContext(ctx).Fatal("Secure server not yet implemented")
+
+	caPool, cert := getServerCertificates(ctx)
+
+	s.server.TLSConfig = &tls.Config{
```
orenfromberg, line 57: If we relied on the certificate hierarchy to validate the client to the server then would it simplify the TLS config to look like this?
```golang
s.server.TLSConfig = &tls.Config{
  Certificates: []tls.Certificate{cert},
  minVersion: tls.VersionTLS12,
  ClientAuth: tls.RequireAndVerifyClientCert,
  ClientCAs: caPool
}
```

orenfromberg (changes requested):
This branch will validate that a client certificate is signed by the root CA but I think it can be simplified a bit since the functionality that should be verifying client info does not seem to be used.

Instead of signing the portal client cert with the root CA, could we sign with the API cert and key? 
I think that will validate that the portal is the only client of the API. 

Similarly, could we sign the API client cert with the Attribution cert and key? That should also validate that the API is the only client of the Attribution service. 

Doing this may simplify the TLS config for the API server (as well as for the Attribution service), as well as validate that only the clients we expect are talking to the servers.

If we go the route of validating clients are using certs that are signed by the services they are using then would there be any benefit to checking that the client cert has a DNS name in the SAN of the server certificate?

dpc-go/dpc-api/src/service/server.go
MrMorie: The closure checks the client certificate. The server certificate does not contain any of the allowed DNS names, only the client certificates.

It was tested by making a call using the two different client certificates, where only one succeeded and the other failed due to missing DNS entries in the SAN.

But yes, if we go with the your suggested approach of signing the certs with the corresponding server certificate then this verification would not be needed.

orenfromberg: I misread that function, thanks for explaining that!

dpc-go/shared/make-certs.sh
dhgreene: How does this affect our [key rotation scenarios](https://jira.cms.gov/browse/DPC-1690)?

dpc-go/dpc-api/src/service/server.go
MrMorie: Yep 👍

dpc-go/shared/make-certs.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+#Generate CA
```
dhgreene, line 2: Can we please put a comment near the top of this file that it is only for the purpose of generating local certificates?

jonfulk:
I don't recall closing this... must've been a misclick.

dpc-go/shared/make-certs.sh
MrMorie: Resolved

dpc-go/dpc-api/src/service/server.go
```diff
@@ -42,11 +46,76 @@ func NewServer(name string, port int, securityDisabled bool, handler http.Handle
...
+		return s.startSecureServer(ctx, true)
+	}
+
+	return fmt.Errorf("invalid auth type. Supported values: NONE, TLS, MTLS")
+}
+
+func (s *Server) startSecureServer(ctx context.Context, useMTLS bool) error {
+	caPool, cert := getServerCertificates(ctx)
+
+	severTlsConf := &tls.Config{
```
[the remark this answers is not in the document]
orenfromberg: is that a spelling error?

orenfromberg (approved):

PR #1371, opened 2021-07-21
DPC-1436 Feature: Proxy requests from users for token

Sun-Mountain:
### Proposed Changes

* Proxy a request for a token to the SSAS service
* Return the token from the SSAS service to the user

### Change Details

* POST requests to `/v2/Token/auth` is proxied to SSAS, and an access token is returned
* The existing (old) `/auth/token` endpoint has been removed

### Security Implications

### Acceptance Validation

Screen Shot 2021-07-28 at 1 46 27 PM


dpc-go/dpc-api/src/v2/ssas_controller.go
```diff
@@ -166,6 +167,31 @@ func (sc *SSASController) CreateSystem(w http.ResponseWriter, r *http.Request) {
...
+	body, _ := ioutil.ReadAll(r.Body)
+	log := logger.WithContext(r.Context())
+
+	if len(body) == 0 {
+		log.Error("Body is empty")
+		fhirror.BusinessViolation(r.Context(), w, http.StatusBadRequest, "Body is required")
+		return
+	}
+
+	// TODO: May need to bring up error codes for more specific errors for troubleshooting
```
dhgreene, line 181: Do we return an OperationOutcome if SSAS is not returning a token?

charmcitygavin: Good thinking. I've added a check to return an OperationOutcome if no token is returned from SSAS. 👍🏻

dhgreene: In a more general sense, are we passing along the helpful information from SSAS regarding _why_ no token is being returned? I don't think it's a security violation to say that the `subject` and `issuer` claims didn't match, that we're dealing with an expired token or other time issue, or some other required claim is missing. Maybe I'm just not seeing how a non-happy path forwards the error to the user. 😄

charmcitygavin: The SSAS client now forwards the error to the user. Great idea! 😃

dhgreene (approved):
👍

PR #1373, opened 2021-07-22
DPC-1443 Added Mutual TLS between attribution and api (GO)

MrMorie:
### Proposed Changes

Currently the connection between DPC-API (GO) and DPC-Attribution (GO) is not secure. This PR secures that connection using MTLS (Mutual TLS)

- Connection should be secured using MTLS
- ~Server should not just validate client certificate, it should also verify the SAN~ Instead of checking for a specific SAN value to verify authorization, we will verify that the client certificate (API's) is signed with the server certificate (Attribution's)
- Certificates should be passed in as env variables

### Change Details
- Added MTL to attribution server and API's attribution client.
- Added logic for parsing certificates passed in via ENV variables.
- Added flag to disable MTLS for local testing purposes. (Enabled by default)

### Security Implications

No new software dependencies

Yes. Added MTL between API and Attribution

No new data stored or interactions with new systems.

- [x] security checklist is completed for this change
SIA: https://confluence.cms.gov/display/DAPC/DPC+v2+SIA 

Not at this moment

### Feedback Requested

Security and certificate implementation
Are we ok with the way certificate parsing is done?


dpc-go/dpc-api/src/main.go
```diff
@@ -60,7 +67,43 @@ func main() {
...
 	if err := http.ListenAndServe(fmt.Sprintf(":%s", port), apiRouter); err != nil {
 		logger.WithContext(ctx).Fatal("Failed to start server", zap.Error(err))
 	}
 }
+
+func getAttrCertificates(ctx context.Context) (*x509.CertPool, tls.Certificate) {
+	caStr := strings.ReplaceAll(conf.GetAsString("ATTRIBUTION_CLIENT_CA_CERT"), "\\n", "\n")
+	crtStr := strings.ReplaceAll(conf.GetAsString("ATTRIBUTION_CLIENT_CERT"), "\\n", "\n")
+	keyStr := strings.ReplaceAll(conf.GetAsString("ATTRIBUTION_CLIENT_CERT_KEY"), "\\n", "\n")
+
```
MrMorie, line 80: the certificate parses did not like the keys without new lines, and I was not able to pass in (non escaped) new lines as an env variable, so this is what I came up with. Any thoughts?

msnook: > the certificate parses did not like the keys without new lines, and I was not able to pass in (non escaped) new lines as an env variable, so this is what I came up with. Any thoughts?

@orenfromberg Was this an issue with the effort you did in `bcda-ssas-app`?  Does [this casting ](https://github.com/CMSgov/bcda-ssas-app/pull/84/files#diff-976bf0364f42cfc105090e325b5cea6842e6d61c8692f8106de6ec5ff371784cR54)help?

dpc-go/dpc-attribution/src/main.go
```diff
@@ -73,3 +121,51 @@ func createV1Services(attrDbV1 *sql.DB, queueDbV1 *sql.DB) (service.JobService,
...
+	}
+}
+
+func getServerCertificates(ctx context.Context) (*x509.CertPool, tls.Certificate) {
+	caStr := strings.ReplaceAll(conf.GetAsString("CA_CERT"), "\\n", "\n")
+	crtStr := strings.ReplaceAll(conf.GetAsString("SERVER_CERT"), "\\n", "\n")
+	keyStr := strings.ReplaceAll(conf.GetAsString("SERVER_CERT_KEY"), "\\n", "\n")
+
+	if caStr == "" {
+		logger.WithContext(ctx).Fatal("Missing server ca cert, env variable CA_CERT is required")
```
msnook, line 148: How is this started up/tested locally?  Should a set of test certs be persisted, that are only valid for local environment bootstrapping?

dpc-go/dpc-attribution/src/main.go
```diff
@@ -73,3 +121,51 @@ func createV1Services(attrDbV1 *sql.DB, queueDbV1 *sql.DB) (service.JobService,
...
+			KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+			DNSName:       host,
+		}
+		_, err := verifiedChains[0][0].Verify(opts)
+		return err
+	}
+}
+
+func getServerCertificates(ctx context.Context) (*x509.CertPool, tls.Certificate) {
+	caStr := strings.ReplaceAll(conf.GetAsString("CA_CERT"), "\\n", "\n")
```
msnook, line 143: For rollout to our AWS envs, will these vars need to be added to the task definition?  Will they need to be manually generated and added to SSM?
What is the plan for getting these certs added to the load balancer and referenced in TF config?
CC: @orenfromberg

dpc-go/dpc-api/src/main.go
dhgreene: @MrMorie that can work, we could base64 encode/decode, or I believe there is a way to get the unescaped newlines into an env var.

dpc-go/dpc-attribution/src/main.go
dhgreene: > For rollout to our AWS envs, will these vars need to be added to the task definition?

Yes

> Will they need to be manually generated and added to SSM?

That's acceptable. Ideally (maybe eventually?) I'd like them to be managed by the AWS cert manager.

> What is the plan for getting these certs added to the load balancer and referenced in TF config?

Yes, @orenfromberg please help. 🙏

msnook: > > For rollout to our AWS envs, will these vars need to be added to the task definition?
> 
> Yes
> 
> > Will they need to be manually generated and added to SSM?
> 
> That's acceptable. Ideally (maybe eventually?) I'd like them to be managed by the AWS cert manager.
> 
> > What is the plan for getting these certs added to the load balancer and referenced in TF config?
> 
> Yes, @orenfromberg please help. 🙏

@dhgreene Yes, the cert itself will live in AWS Cert Manager and be referenced (by ARN) in the terraform code for attachment to the internal load balancer.  However, the env vars referenced in this application code will need to pull the specific pieces (public, private, ca, etc) from SSM.

dpc-go/dpc-attribution/src/main.go
MrMorie: There is currently a flag that lets you start it up locally without MTLS, I thought about committing certificates but I wasn't sure if that is something we wanted to do.

msnook: In my opinion, the default should be MTLS and the certs should be persisted.  I can't remember if we use vault for this in the BCDA SSAS env... you might consider taking a look at that pattern for consistency.

dpc-go/dpc-api/src/main.go
dhgreene: Let's go with this; if we can get the unescaped newlines into an env var, this code shouldn't break.

dhgreene (dismissed):
Let's please talk about our deployment strategy.

MrBilnon (approved):

orenfromberg (changes requested):
I think the comments I made in #1370 also apply to this PR. I don't think the client validation is happening the way we expect. I also think the TLS config can be simplified a bit by signing the API client cert with the Attribution cert and key. Would love to know if I am missing something!

dhgreene (approved):
In conjunction with the recent ops change that configures this to be "off," this is good to go. 👍

orenfromberg (approved):

PR #1376, opened 2021-07-26
dpc - 7777 Feature:Optimize Use

sjsdfg:
### Proposed Changes

Optimize Function Use

### Change Details
`headerPrefer == "" ||` is same as `StringUtils.IsEmpty(headerPrefer)`
can see https://github.com/sjsdfg/common-lang-in-go/blob/master/StringUtils/string_utils_test.go
```
func TestIsEmpty(t *testing.T) {
	assert.Equal(t, true, IsEmpty(""))
}
```

`!StringUtils.EqualsAnyIgnoreCase` same as `StringUtils.EqualsNoneIgnoreCase`

```
func EqualsNoneIgnoreCase(str string, list ...string) bool {
	return !EqualsAnyIgnoreCase(str, list...)
}
```

### DIFF
--- a/dpc-go/dpc-api/src/go.mod
+++ b/dpc-go/dpc-api/src/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/microcosm-cc/bluemonday v1.0.4
 	github.com/pkg/errors v0.9.1
 	github.com/samply/golang-fhir-models/fhir-models v0.2.0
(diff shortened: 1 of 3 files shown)


dhgreene:
@sjsdfg Thanks! We're not accepting PRs from the general public, but we've put this in our backlog of code improvements.

PR #1378, opened 2021-07-27
DPC-1567 create and delete token

MrBilnon:
### Fixes [DPC-1567](https://jira.cms.gov/browse/DPC-1567)

Ability to create and delete tokens

### Proposed Changes

* create create token endpoint
* create delete token endpoint

### Security Implications

### Acceptance Validation

Screen Shot 2021-07-26 at 7 33 58 PM
Screen Shot 2021-07-26 at 7 34 57 PM
Screen Shot 2021-07-26 at 7 35 06 PM

### Feedback Requested

Should the create token return the id, currently in V1 it only returns the token and no other information


dpc-go/dpc-api/src/v2/ssas_controller.go
```diff
@@ -27,6 +27,94 @@ func NewSSASController(ssasClient client.SsasClient, attrClient client.Client) *
...
+		return
+	}
+
+	if mOrg.SsasSystemID == "" {
+		log.Error(fmt.Sprintf("relation with implementerID: %s and organizationID: %s is not tied to a system", implementerID, organizationID))
+		fhirror.BusinessViolation(r.Context(), w, http.StatusBadRequest, "a system was not found for this implementer/org relationship")
+		return
+	}
+
+	label := r.URL.Query().Get("label")
```
dhgreene, line 61: Can we make this part of the body?

MrBilnon: > Can we make this part of the body?

in v1 dpc, we use query are we going to change it for v2?

dhgreene: Yes. In DPC v1 we are using a framework that blurs the difference between query and form parameters. Here we aren't, and we have the freedom to be consistent. IMO if it's a required field, and we're always elsewhere asking for data in the body of the request, let's continue to do that. This is an OK breaking change.

dhgreene (approved):
👍

PR #1386, opened 2021-07-29
DPC-1627 add signature check

MrBilnon:
…key and create system

### Fixes [DPC-1627](https://jira.cms.gov/browse/DPC-1627)

Currently, create system and adding a public key does not enforce the presence of a signature for the public key. This PR enforces that presence.

### Proposed Changes

* add signature field to create system request
* check signature in both createSystem and addPublicKey and return fhir error (operation outcome) when the signature is not present.

### Security Implications


dhgreene (approved):
🐎

PR #1388, opened 2021-07-30
DPC-1302: Replicate in implementer portal - Provider Organization Management page

Sun-Mountain:
### Fixes [DPC-1302](https://jira.cms.gov/browse/DPC-1302)

The Implementer app will make calls to the Golang API for provider organization with org details and the ability to create a system with the provider org's initial token. Users will then be able to add public key/signature pairs and create new client tokens as needed and view a list of added public keys and tokens that have been added/created.

### Proposed Changes

- [x] User can view provider org page w/ provider org info
- [x] User can create system for provider org
- [x] User can create public key
- [x] User can view list of all public key ids
- [x] User can delete public key
- [x] User can create client token
- [x] User can view client token
- [x] User can view list of client tokens
- [x] User can delete client token

### Change Details

- Added `omitempty` to `ips` in `ProxyCreateSystemResponse`
- How to create public key and signature instructions
- Add `rails-controller-testing` for testing purposes
- Add how to create public key/signature pair instructions to public key form
  - Add `copy-text.js` to copy openssl commands
  - Add relevant accordion styling
  - Add `snippet.txt` with update text (`This is the snippet used to verify a key pair.`) for download to create signature.
- Styling for public keys and credentials
- Add `imp_id`, `api_service`, `missing_key`, and `render_error` methods to application controller
- `dpc-impl/app/services/api_client.rb` lines 113 - 119: Because the `create_client_token` returns the client token as a string had to add a check for the response body that if a JSON object continue with `eval()` otherwise return response body as is.
- Add Client Token Manager to help manage http calls for client token creation and deletion
- Add Public Key Manager to help manage http calls for system creations and public key creation and deletion.
  - Put in checks for valid encoding and signature presence

### Security Implications

### Acceptance Validation

Due to time constraints with the upcoming code freeze on August 5, I created [DPC-1638](https://jira.cms.gov/browse/DPC-1638) to address missing tests for the provider org show page, client token management, and public key management. This will allow the Implementer App to be able to be released on Aug 12.

I also created [DPC-1639](https://jira.cms.gov/browse/DPC-1639) to add in the 25 character limit to client token labels and the GUI that already exists in the web app.

Below are screen captures of workflows to prove their efficacy:

**View Provider Org Page**\
recording

**Create System & View Initial Token**\
recording (1)

**Adding New Public Key**\
recording

**Deleting Public Key**\
recording

**Public Key Errors**\
**Cannot submit blank form or upload Private key**\
recording

**Add New Client Token**\
recording

**Delete Client Token**\
recording

**Client Token Must Have Label**\
recording


dpc-impl/app/services/api_client.rb
```diff
@@ -65,7 +110,13 @@ def http_request(request, uri)
   def parsed_response(response)
     return self if response.body.blank?
 
-    eval(response.body)
+    begin
+      !!JSON.parse(response.body)
+    rescue
+      return response.body
+    else
+      eval(response.body)
```
dhgreene, line 118: What are we getting from `eval()` that `JSON.parse()` doesn't give us? I know our current source is trusted, but it's a dangerously wide open window, and we might reuse this somewhere less trusted in the future.

dpc-impl/app/services/public_key_manager.rb
```diff
@@ -0,0 +1,74 @@
...
+    api_client = ApiClient.new
+    api_client.create_public_key(@imp_id, @org_id,
+                                 params: { public_key: public_key,
+                                           signature: signature })
+
+    return api_response(api_client)
+  end
+
+  def create_system(org_name:, public_key:, signature:)
+    public_key = strip_carriage_returns(public_key)
```
dhgreene, line 30: Have we found general whitespace to be a problem in practice? Can we practically remove spaces too?

dpc-impl/app/services/public_key_manager.rb
```diff
@@ -0,0 +1,74 @@
...
+    return api_response(api_client)
+  end
+
+  def delete_public_key(key_id)
+    api_client = ApiClient.new
+    api_client.delete_public_key(@imp_id, @org_id, key_id)
+  end
+
+  def invalid_encoding?(key_string)
+    key = OpenSSL::PKey::RSA.new(key_string)
```
dhgreene, line 53: Noting for the future that we also intend to accept elliptical keys. This is OK for our initial sandbox release.

Can you add a TODO to that effect?

dpc-impl/app/views/provider_orgs/show.html.erb
```diff
@@ -0,0 +1,61 @@
...
+    <% end %>
+  </div>
+
+  <h1 class="ds-text-heading--4xl ds-title"><%= @org[:name] %></h1>
+  <div>
+    <div>
+      <strong>Org ID:</strong> <%= @org[:id] %>
+    </div>
+    <div>
+      <strong>Provider NPI:</strong> <%= @npi %>
```
dhgreene, line 21: If we decide to support multiple NPIs for each provider organization, how much pain will we have with the implementer portal?

dpc-impl/app/views/shared/_need-help.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="nav_box_container">
+  <div class="nav_box" id="dash_nav_help">
+    <h2>Need Help?</h2>
+    <strong>Resources:</strong>
+    <ul>
+      <li><a href="<%= ENV['STATIC_SITE_URL'] %>docsV1">DPC's User Guide</a></li>
```
dhgreene, line 6: Why v1?

dpc-go/dpc-api/docker-compose.yml
```diff
@@ -16,4 +16,4 @@ services:
       - DPC_SSAS-CLIENT_ADMIN-URL=http://ssas:3004
       - DPC_SSAS-CLIENT_PUBLIC-URL=http://ssas:3003
       - DPC_SSAS-CLIENT_CLIENT-ID=31e029ef-0e97-47f8-873c-0e8b7e7f99bf
-      - DPC_SSAS-CLIENT_CLIENT-SECRET=?
\ No newline at end of file
+      - DPC_SSAS-CLIENT_CLIENT-SECRET=
```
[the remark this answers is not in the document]
Sun-Mountain: This is only showing as something that I added because I delete the unnecessary `?` for my client secret. These variables are required to set locally to run in docker until they bootstrap it on the api side. Not sure what I can do here.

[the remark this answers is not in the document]
Sun-Mountain: it's necessary for ssas creds when running the api thru docker. I accidentally uploaded a `?` so it's marking it as something I added, but this is something that is necessary for local shtuff.

dpc-impl/app/services/api_client.rb
Sun-Mountain: eval returns a hash, where it was breaking was with the client token request because it returns just the token in a string which eval doesn't like. So I had to use `JSON.pars()` to determine if it was a JSON object and if not, return the body.

If we want to simplify, the client token would have to be delivered within a JSON object.

Sun-Mountain: I couldn't get it to rescue from `eval()`

dpc-impl/app/services/public_key_manager.rb
Sun-Mountain: Will do!

dpc-impl/app/views/provider_orgs/show.html.erb
Sun-Mountain: That shouldn't be difficult, I think the biggest lift would be to decide how to package those NPI's and deliver them in the request. (Does that make sense?)

dpc-impl/app/views/shared/_need-help.html.erb
Sun-Mountain: Because I thought we were moving to v2 documentation, but we're keeping v1 documentation for now so I needed to differentiate the docs files. If you don't like that we can keep v1 docs as `docs`, but that would require a separate ticket for the static site.

dpc-impl/app/services/api_client.rb
MrBilnon: This is my bad, the create token currently returns just a string instead of json. The mistake was looking at the v1 create token task as an example instead of the v1 resource class

MrBilnon (approved):

dpc-impl/app/services/api_client.rb
dhgreene: OK, we have a ticket for addressing this later.

dhgreene (approved):
👌

PR #1392, opened 2021-08-02
DPC-1576: Adding health checks to v2 public and admin API services

orenfromberg:
### Related to [DPC-1576](https://jira.cms.gov/browse/DPC-1576)

This branch adds health checks to the v2 public and admin API services.

### Proposed Changes

Adding health check to `/v2/_health` for both public and admin API services.

### Change Details

The checks are mostly superficial because they do not do anything more than return 200.

However, they've been stubbed to return more data once that functionality is needed.

### Security Implications

### Acceptance Validation

- [x] health checks are functional for v2 admin and public API services.

### Feedback Requested

All feedback is welcome, especially on keeping the code clean and simple.


dpc-go/dpc-api/src/service/public/router.go
```diff
@@ -20,6 +22,12 @@ func buildPublicRoutes(cont controllers) http.Handler {
 	r.With(middleware2.Sanitize).Route("/v2", func(r chi.Router) {
 		r.Use(middleware.SetHeader("Content-Type", "application/fhir+json; charset=UTF-8"))
 		r.Get("/metadata", cont.Metadata.Read)
+		r.Get("/_health", func(w http.ResponseWriter, r *http.Request) {
+			m := make(map[string]string)
+			m["database"] = "ok"
```
dhgreene, line 27: Hmm. It doesn't look like we actually contact the database here. I know historically we've done database and BFD health checks, but are we limited to those options? We're simply saying as an API, we're ping-able.

If we're going to say the database connection is ok, let's test it.

MrBilnon: Attribution is the only thing that uses the database. I think the go app doesn’t start if it can’t connect to the database. So should we just ping attribution to see if it works?

dhgreene: This is a statement that the API service _currently_ thinks it's healthy. I'd be happy enough with:
```suggestion
			m["api"] = "ok"
```

dhgreene (approved):
❤️

PR #1417, opened 2021-08-16
DPC-1639: Feature Add counter to convey label length limit to user

charmcitygavin:
### Fixes [DPC-1639](https://jira.cms.gov/browse/DPC-1639)

We enforce a limit of 25 characters for client token labels but don't let the user know. This PR adds a counter to signify that client token labels can be no longer than 25 characters.

### Proposed Changes

* Use JS function to add a character limit count, similar to how its done in `dpc-web`.

### Change Details

As a user enters the new label for a client token, the counter counts up to 25 characters, conveying to the user a character count limit of 25.

### Security Implications

This PR adds a front-end JavaScript character counter to ensure users understand that new client tokens in the implementer portal can have a maximum of 25 characters.

### Acceptance Validation

Yes, the new character counter works as expected.

Screen Shot 2021-08-16 at 12 06 05

### DIFF
--- a/dpc-impl/app/assets/javascripts/character-count.js
+++ b/dpc-impl/app/assets/javascripts/character-count.js
@@ -1,0 +1,20 @@
+function characterCount(string, countMax, counter) {
+  var input = document.getElementById(string),
+      inputValue = input.value,
+      inputLength = inputValue.length,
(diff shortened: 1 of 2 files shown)


Sun-Mountain (approved):
🚀

PR #1420, opened 2021-08-18
DPC-1437 Feature: POST /api/v2/Token/validate (proxy to SSAS)

charmcitygavin:
### Fixes [DPC-1437](https://jira.cms.gov/browse/DPC-1437)

We want `POST /api/v2/Token/validate` to say whether or not a JWT is valid and provide useful information if the JWT is missing any of the required claims. It proxies the request to a new `POST v2/introspect` SSAS endpoint.

### Proposed Changes

* Proxy DPC `POST /api/v2/Token/validate` requests to the new SSAS `POST /v2/introspect` endpoint (see [this related PR](https://github.com/CMSgov/bcda-ssas-app/pull/98))

### Change Details

* New proxy endpoint created in DPC

### Security Implications

This PR does not deal with PII/PHI at all and simply returns whether a token is valid or not.

- [x] [security checklist](https://confluence.cms.gov/display/DAPC/2021-09-02%3A+DPC+4.2+Security+Checklist) is completed for this change

### Acceptance Validation

Yes, tested in Postman and unit tests

Screen Shot 2021-09-09 at 09 38 30

Screen Shot 2021-09-09 at 09 37 43

Screen Shot 2021-09-09 at 09 36 48

### Feedback Requested

Any and all questions/suggestions welcome, as always. 😄


charmcitygavin:
@dhgreene A good question came up when reviewing [this ticket](https://jira.cms.gov/browse/DPC-1437). The [V2 tech spec](https://confluence.cms.gov/display/DAPC/DPC+Version+2+Technical+Specification) indicates that the API's `POST /api/v2/Token/validate` route should proxy to SSAS's `POST /v2/introspect` route. Since the `/v2/introspect` route didn't exist on SSAS, I added it.

However, it would also be possible to proxy requests to the existing SSAS `POST /v2/token_info` route and return whether or not the token is valid. Both approaches would work, but using the existing route would avoid adding a new endpoint that essentially makes the same check.

Do you know which is the expected approach?

dpc-go/dpc-api/src/v2/ssas_controller.go
```diff
@@ -417,6 +416,31 @@ func (sc *SSASController) GetAuthToken(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	// Validate token with SSAS Client
+	resBytes, err := sc.ssasClient.ValidateToken(r.Context(), body)
+	if err != nil {
+		log.Error("Failed to validate token", zap.Error(err))
+		fhirror.ServerIssue(r.Context(), w, http.StatusInternalServerError, fmt.Sprintf("Failed to validate token: %s", err))
+		return
+	}
```
jonfulk, line 436: SSAS returns a 400 when the token is invalid, but we're returning a 500 no matter what here. Can we catch the 400 and return that back to the caller?

jonfulk (commented):
Shouldn't we return a 400 instead of a 500 when there is a missing or invalid token in the request body? Or do we just always send a 500 with an operation outcome here?

dpc-go/dpc-api/src/v2/ssas_controller.go
charmcitygavin: In order to catch the status code it looks like we'd need to rework the doPost function on the DPC side and perhaps also the jsonError function on the SSAS side, which feels a little out of scope. However, I've updated this code above to correctly reflect the 400 error that SSAS sends when this type of error is triggered.

dhgreene (approved):
👍

PR #1449, opened 2021-08-30
DPC-1595 remove duplicate running of tests

MrBilnon:
### Fixes [DPC-1595](https://jira.cms.gov/browse/DPC-1595)

The make file has the tests running twice because both are calling up, just one builds before running and the other doesn't.

If you look at the current job logs:
https://github.com/CMSgov/dpc-app/runs/3457898433

You see that it is running the linting/security/tests twice

### Security Implications

### Acceptance Validation

Still fails when a test fails
Screen Shot 2021-08-29 at 11 23 21 PM


dhgreene (commented):
In the Acceptance Validation section, can you show an example of this streamlined version still blocking on failed tests?

dhgreene (approved):
Good catch! 💨

PR #1454, opened 2021-09-02
DPC-1677 Updated netty-codec version used by bluebutton

MrMorie:
### Proposed Changes
Update to latest netty-coded version

### Change Details
updated to 4.1.67.Final

### Security Implications
Once merged we will need to verify that all vulnerabilities related to this library have been resolved.

While not a new dependency, this PR upgrades netty-codec to a newer version to respond to a Snyk alert

No

No

- [x] [security checklist is completed](https://confluence.cms.gov/display/DAPC/2021-08-19%3A+DPC+Sprint+4.1+Security+Checklist) for this change

No, simply a version update

#### Before:
Screen Shot 2021-09-01 at 6 56 58 PM
#### After
Screen Shot 2021-09-01 at 7 01 11 PM


MrBilnon:
don't think this is really needed

MrMorie:
> don't think this is really needed because looking at the snyk findings, it is looking at a module that uses another module via a jar and can't fully comprehend the dependencies in that jar. Can be explained here: https://support.snyk.io/hc/en-us/articles/360001649458-Mismatch-of-Vulnerabilities-Why-does-the-CLI-show-a-different-number-of-vulnerabilities-than-through-the-Snyk-App-
> 
> If you look directly at the blue button module in snyk then you will see we are not using a vulnerable version

Thanks @MrBilnon , I shall exclude this dependency from our snyk scans instead 👥

dhgreene (dismissed):
👍

dhgreene (approved):
🆗

PR #1531, opened 2021-12-11
we don't use log4j, but we can update the dependencies to a non vulnerable version

MrBilnon:
### Fixes [DPC-1997](https://jira.cms.gov/browse/DPC-1997)

Log4j has an exploit, we don't use log4j for logging but that doesn't mean we don't have dependencies that bring in log4j. We can upgrade those dependencies just for the sake of updating them.

### Proposed Changes

update log4j to 2.15.0

### Security Implications


dhgreene (approved):
👍

PR #1665, opened 2022-07-01
DPC-2579: Adjust filter query for un-registered orgs

mftcodes:
This query includes un-registered orgs as well as disabled in the api orgs into the same query for the "not enabled in API" org filter. This fixes the original bug DPC-2351

## Change Details

* Changed the query using a join between organizations and registered_organizations when trying to use the filter for "not enabled in API" organizations from the Web Admin Organization view. This will not include those registered orgs that have been disabled, and orgs that have not yet been registered. Previously, unregistered orgs were not showing. 

## Security Implications

### DIFF
--- a/dpc-admin/app/models/organization.rb
+++ b/dpc-admin/app/models/organization.rb
@@ -37,7 +37,7 @@ class Organization < ApplicationRecord
   }
 
   scope :is_not_registered, -> {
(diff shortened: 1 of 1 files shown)


TiannaW24 (approved):
Superb

TiannaW24 (approved):
Even more superb now

PR #1678, opened 2022-07-22
Lg 1303 fix

lgarceau768:
## Change Details
Added the delete public key workflow into the dpc web portal, in the frontend and within the ruby on rails mvc architecture. 
Needed to expose the `:delete` method within the `routes.rb` as well as add the functionality to delete the public key within the api and that needed to be reflected within the `api_client.rb` as well as the `public_keys_controller.rb`
* 

## Security Implications
Should be no security implications no new data is being stored, the accessibility to delete the public key via the portal is the only thing added.


dpc-web/app/services/api_client.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+# rubocop:disable Metrics/ClassLength
```
mftcodes, line 3: What's this all about? We're disabling method/class lengths? Then enabling it at the end? If there is a limit set, is it there for a reason and we should stay within it, or is the limit set too low? 
This is all new to me, so take this with a grain of salt.

dpc-web/app/controllers/public_keys_controller.rb
```diff
@@ -10,6 +10,19 @@ def new
     @organization = current_user.organizations.find(params[:organization_id])
   end
 
+  def destroy
```
TiannaW24, line 13: These changes need unit tests.

dpc-web/app/services/public_key_manager.rb
```diff
@@ -40,6 +40,11 @@ def invalid_encoding?(key_string)
     true
   end
 
+  def delete_public_key(params)
```
TiannaW24, line 43: Test?

dpc-web/app/services/api_client.rb
```diff
@@ -52,6 +54,12 @@ def delete_client_token(reg_org_api_id, token_id)
     delete_request(uri_string, delegated_macaroon(reg_org_api_id))
   end
 
+  def delete_public_key(reg_org_api_id, public_key_id)
```
TiannaW24, line 57: Tests?

TiannaW24 (changes requested):
See comments regarding tests. Pretty much every touch to a .rb file should have corresponding unit tests in the _spec.rb file.

dpc-web/.rubocop.yml
```diff
@@ -20,6 +20,9 @@ Bundler/OrderedGems:
 
 # Layout/LineLength:
 #   Max: 120
+Metrics/ClassLength:
```
TiannaW24, line 23: Much preferred to disabling/re-enabling 👍🏾

dpc-web/spec/services/public_key_manager_spec.rb
```diff
@@ -15,8 +15,10 @@
 
           api_client = instance_double(APIClient)
           allow(APIClient).to receive(:new).and_return(api_client)
+          allow(api_client).to receive(:delete_public_key)
```
TiannaW24, line 18: You should have a whole new test suite here that tests the delete functionality the same way this create test suite tests the create functionality.

dpc-web/spec/controllers/public_keys_controller_spec.rb
```diff
@@ -35,6 +35,57 @@
...
+          allow(stub).to receive(:get_public_keys).and_return(stub)
+          allow(stub).to receive(:response_body).and_return(default_org_creation_response, { 'entities' => [] })
+
+          org = create(:organization, :api_enabled)
+          user.organizations << org
+
+          allow(stub).to receive(:delete_public_key).and_return(false)
+
+          get :destroy, params: { id: 1, organization_id: org.id }
+          expect(response).to render_template(:new)
```
TiannaW24, line 83: Are we expecting the correct thing here? Shouldn't a failed call render an error?

TiannaW24 (changes requested):

dpc-web/spec/controllers/public_keys_controller_spec.rb
lgarceau768: I pulled the test suite functionality from the `client_tokens_controller_spec.rb` also I am pretty sure that is correct because the only difference from the main `:new` page when there is an error its rendered as a flash message but still under the `:new` template. (See https://github.com/CMSgov/dpc-app/blob/master/dpc-web/app/controllers/public_keys_controller.rb line 49 and down)

dpc-web/app/services/api_client.rb
lgarceau768: Yes the comment is no longer in the PR

mftcodes (dismissed):
These new changes all look good to me!

dpc-web/spec/services/public_key_manager_spec.rb
TiannaW24: 👍🏾

TiannaW24 (approved):
Looks great

mftcodes (approved):
approved

PR #1679, opened 2022-08-01
Increase JobTimeout Threshold

TiannaW24:
## Change Details

*  Increase the timeout threshold for completing requests from 30 to 60
* In the case of the /Patient/$everything endpoint, this timeout is too short and results in >90% of requests to the endpoint timing out and returning 500s from the API
* Add some clarifying comments

## Security Implications

### DIFF
--- a/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
@@ -195,6 +195,7 @@ public class PatientResource extends AbstractPatientResource {
                 .execute();
 
         if (practitioner == null) {
+            // Is this the best code to be throwing here?
             throw new WebApplicationException(HttpStatus.UNAUTHORIZED_401);
         }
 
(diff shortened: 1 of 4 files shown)


dpc-api/src/main/resources/application.conf
```diff
@@ -3,7 +3,7 @@ dpc.api {
     include "queue.conf"
 
     jobTimeoutInSeconds = ${?JOB_TIMEOUT_IN_SECONDS}
-    jobTimeoutInSeconds = 30
+    jobTimeoutInSeconds = 60
```
TiannaW24, line 6: @whytheplatypus objections?

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
```diff
@@ -195,6 +195,7 @@ public Bundle everything(@ApiParam(hidden = true) @Auth OrganizationPrincipal or
                 .execute();
 
         if (practitioner == null) {
+            // Is this the best code to be throwing here?
```
whytheplatypus, line 198: Probably not, just looking at the if statement it looks more like a 403. might be out of scope for this changeset though.

dpc-api/src/main/resources/application.conf
whytheplatypus: It looks like this gets `jobTimeoutInSeconds` from some kind of var and then immediately hard codes it? Can we instead use the var and a sane default?

Otherwise no objections to the number!

whytheplatypus (approved):

PR #1681, opened 2022-08-03
Jstewart/dpc 2660 check org id before create token

mftcodes:
## Change Details

* Added the OrganizationResource into the GenerateClientTokens class in order to call `orgSearch` to ensure we are only creating tokens for current existing orgs. The system currently allows orphans tokens that can unnecessarily add data to the db. 

## Security Implications

### DIFF
--- a/dpc-api/src/main/java/gov/cms/dpc/api/tasks/tokens/GenerateClientTokens.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/tasks/tokens/GenerateClientTokens.java
@@ -59,15 +59,14 @@ public class GenerateClientTokens extends Task {
             final Organization organization = new Organization();
             organization.setId(organizationId);
 
-            // `checkOrg` parameter is meant to be optional, if not present, default `canCreateToken` to true.
-            // When `checkOrg` is provided, ensure the submitted orgId matches an org in the db before creating token.
-            Boolean canCreateToken = true;
+            /*
+             * `checkOrg` parameter is meant to be optional, if provided and "true" ensure the submitted orgId
+             * matches an org in the db before creating token. `checkOrg` Not present, just create the token.
+             */
             final Boolean needOrgValidation = !needCheckOrgCollection.isEmpty()
                     && !StringUtils.isBlank(needCheckOrgCollection.asList().get(0))
                     && needCheckOrgCollection.asList().get(0) == "true";
-            if(needOrgValidation) {
-                canCreateToken = validateOrgExists(organization);
-            }
+            Boolean canCreateToken = needOrgValidation ? validateOrgExists(organization) : true;
 
             if(canCreateToken) {
                 createTokan(output, labelCollection, expirationCollection, organization);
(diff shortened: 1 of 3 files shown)


dpc-api/src/main/java/gov/cms/dpc/api/tasks/tokens/GenerateClientTokens.java
```diff
@@ -31,40 +34,68 @@ public class GenerateClientTokens extends Task {
...
-            Optional<OffsetDateTimeParam> expiration = Optional.empty();
-            if(!expirationCollection.isEmpty() && !StringUtils.isBlank(expirationCollection.asList().get(0))){
-                expiration = Optional.of(new OffsetDateTimeParam(expirationCollection.asList().get(0)));
+            final String organizationId = organizationCollection.asList().get(0);
+            final Organization organization = new Organization();
+            organization.setId(organizationId);
+
+            Boolean orgExists = true;
+            if(!needCheckOrgCollection.isEmpty()
+                    && !StringUtils.isBlank(needCheckOrgCollection.asList().get(0))
```
lgarceau768, line 64: This might just be my code practice, but I would split up the recurring `needCheckOrgCollection.asList().get(0)` into a seperate constant

mftcodes: Yes, agree. This would usually be my code practice as well. I think I'll spend a little more time refactoring this to be a bit more clean.

mftcodes: The issue here is if nothing is passed in for this parameter, which is likely in many cases, then an assignment would fail as it would be null. That's why this check is written sequentially as it is.

mftcodes: Now using a local variable to make it more clear "why." Resolving.

dpc-api/src/main/java/gov/cms/dpc/api/tasks/tokens/GenerateClientTokens.java
```diff
@@ -31,40 +34,75 @@ public class GenerateClientTokens extends Task {
...
+             * `checkOrg` parameter is meant to be optional, if provided and "true" ensure the submitted orgId
+             * matches an org in the db before creating token. `checkOrg` Not present, just create the token.
+             */
+            final boolean needOrgValidation = !needCheckOrgCollection.isEmpty()
+                    && !StringUtils.isBlank(needCheckOrgCollection.asList().get(0))
+                    && needCheckOrgCollection.asList().get(0) == "true";
+            boolean canCreateToken = needOrgValidation ? validateOrgExists(organization) : true;
+
+            if(canCreateToken) {
+                createTokan(output, labelCollection, expirationCollection, organization);
```
krobertson3314, line 72: can you update typo / rename to 'createToken'

PR #1729, opened 2022-11-09
Update To Nokogiri

lgarceau768:
## Change Details
Minor update to `nokogiri` to get past a rubocop error in order to allow a docker build
*


lgarceau768:
Leaving this unmerged because it was thought to be an error blocking a docker build but turned out to be something different. But this could still be a solution to a Snyk issue down the road so we are leaving it labeled as`do-not-merge`

lgarceau768:
Updated to `1.3.10`

PR #1740, opened 2022-11-18
Add a make down command for start-v2-portals

TiannaW24:
## Change Details

* Add a command for bringing down the resources that are created when we use make `ci-portals-v1`

## Security Implications

### DIFF
--- a/dpcv1-portals-test.sh
+++ b/dpcv1-portals-test.sh
@@ -13,8 +13,8 @@ make website
 make admin
 
 # Prepare the environment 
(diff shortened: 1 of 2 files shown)


dpcv1-portals-test.sh
```diff
@@ -13,25 +13,25 @@ make website
 make admin
 
 # Prepare the environment 
-docker-compose -p start-v2-portals -f docker-compose.yml -f docker-compose.portals.yml up start_core_dependencies
```
TiannaW24, line 16: It was really bothering me that we called the docker project "start-v2-portals" when we're spinning up the v1 portals

mftcodes (approved):
Looks good

PR #1778, opened 2022-12-19
DPC-565 Data Headers

lgarceau768:
## Change Details
Making the `/Data` endpoint `@Produce` application/ndjson instead of producing an aplication/fhir+json

Needed to update the tests as well to contain the `Accepts application/ndjson` header
* 

## Security Implications

### DIFF
--- a/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/DataResource.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/DataResource.java
@@ -43,6 +43,7 @@ import static gov.cms.dpc.fhir.dropwizard.filters.StreamingContentSizeFilter.X_C
  */
 @Api(tags = {"Bulk Data", "Data"}, authorizations = @Authorization(value = "access_token"))
 @Path("/v1/Data")
+@Produces("application/ndjson")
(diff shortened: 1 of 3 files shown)


mftcodes:
I'm a little confused here. The ticket notes that the endpoints need to accept `application/fhir+ndjson`, which is what it was, and we reduced it down to `application/ndjson`. 
The screen shot on the ticket does show a failure with someone submitting `application/ndjson` which failed... 
The ticket also says, "... /Data endpoint to accept multiple headers ..." which makes me think it should accept both?

lgarceau768:
> I'm a little confused here. The ticket notes that the endpoints need to accept `application/fhir+ndjson`, which is what it was, and we reduced it down to `application/ndjson`. The screen shot on the ticket does show a failure with someone submitting `application/ndjson` which failed... The ticket also says, "... /Data endpoint to accept multiple headers ..." which makes me think it should accept both?

The acceptance criteria is listed as it should be equal to `application/ndjson` so that is what I went with as a point of truth!

lgarceau768:
Checked the validity here! 
Screenshot 2022-12-21 at 3 28 49 PM
It currently does not work unless the `Accept */*` header is used!

TiannaW24 (approved):

PR #1866, opened 2023-04-20
adding gitleaks precommit config

gfreeman-navapbc:
## Change Details

* adding gitleaks precommit config to all repos


gsf (approved):
Looks good

PR #1885, opened 2023-05-01
[DPC-2868] GitHub Actions Optimizations

chris-ronning-ny:
## Change Details

This is an extension of the work completed here: https://github.com/CMSgov/dpc-app/pull/1884

Now that the V2 tests have been removed, I've split the v1 tests into two separate processes that run in parallel, rather than `dpc-admin` and `dpc-web` running one after the other. This cuts down our CI runtime by roughly 10 minutes.

There is a possibility that these changes may break some step of our Jenkins pipeline. If so, I can push changes quickly, and revert this PR if necessary.


dpc-admin-portal-test.sh
```diff
@@ -12,7 +12,7 @@ make admin
 
 # Prepare the environment 
 docker-compose -p start-v1-portals -f docker-compose.yml -f docker-compose.portals.yml up start_core_dependencies
-docker-compose -p start-v1-portals -f docker-compose.yml -f docker-compose.portals.yml run --entrypoint "bundle exec rails db:create db:migrate RAILS_ENV=test" dpc_admin
```
chris-ronning-ny: Before I made this change, this script effectively wasn't running a database migration, which caused all tests in this suite to fail. My theory was that, when we were running `dpcv1-portals-test.sh`, this migration was being performed during dpc_web commands. The current line, directly lifted from `dpcv1-portals-test.sh`, was effectively doing nothing -- and because the migrations were already performed lines earlier, it didn't matter. 

This theory turned out to be correct: Changing this line resemble the equivalent in `dpc-web-portal-test.sh` causes the script to run the DB migration successfully.

michael-wojcik (approved):
LGTM. Thanks, @chris-ronning-ny. Let's definitely keep an eye on that `dpcv1-portals-test.sh` as a potential point of failure, though.

PR #1911, opened 2023-07-17
Me dpc 3494 part 2 upgrade mockito

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3494

## 🛠 Changes

Updated Mockito from 3.4.6 to 5.4.0.
Removed mocking of Method class from FHIRRequestFeatureTest.
Updated BaseResourceUnitTest to use new MockedStatic and MockedConstruction.

## ℹ️ Context for reviewers

We were running an old version of Mockito that didn't support mocking static or final methods, or constructors, and this was preventing us from pushing code coverage as far as we otherwise could.  The new version of Mockito was also causing unit tests to fail randomly and erratically.  

The new version was causing tests to fail because an old test in `FHIRRequestFeatureTest` was mocking the `Method` class from `java.lang.reflect`.  Mockito depends on this class, along with a few others from that package, and mocking them is essentially rewriting Mockito's plumbing while it's running.  This leads to any tests running afterwards potentially failing in random ways.  See issue [here](https://github.com/mockito/mockito/issues/2026#issuecomment-741663017) for a better explanation.

## ✅ Acceptance Validation

We're running the new version of Mockito, we're using some of its new features, and all unit tests are passing.

## 🔒 Security Implications

### DIFF
--- a/dpc-api/pom.xml
+++ b/dpc-api/pom.xml
@@ -171,12 +171,6 @@
             <artifactId>mockito-core</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.mockito</groupId>
-            <artifactId>mockito-inline</artifactId>
-            <version>4.2.0</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-params</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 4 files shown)


kyeah (approved):
Looks great, thanks for digging into this! ⭐

PR #1940, opened 2023-08-25
Me dpc 3576 fix logging

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3576

## 🛠 Changes

1.  Updated TokenResource so that when `/auth` is called with the same JWT in succession (a JWT replay attack), the error logged will include the correct organization id.  The org id is pulled from the macaroon sent as part of the JWT.
2. Updated default logging so that when no APPLICATION env variable is set it uses real names by default.  I.E., when running locally in Docker, the service tag in the logs will say "dpc-api" instead of "unknown_service".
3. Added make targets `start-dpc-debug` and `start-app-debug` that function just like `start-dpc` and `start-app`, but they first recompile dpc with debug extensions and start each service with the debugging agent enabled.  This lets you connect a remote debugger to them and made working on this ticket much, much easier.

## ℹ️ Context for reviewers

In the case of a JWT being replayed, the error logged included a random org id instead of the real one contained in the JWT.

## ✅ Acceptance Validation

Ran locally and observed the new logs.

## 🔒 Security Implications


dpc-aggregation/src/main/resources/application.conf
```diff
@@ -55,7 +55,7 @@ dpc.aggregation {
         environment = ${?ENV}
         version = "unknown_version"
         version = ${?APP_VERSION}
-        application = "unknown_application"
+        application = "dpc-aggregation"
```
chris-ronning-ny, line 58: I'm confused as to why this was like this in the first place 😅

chris-ronning-ny (approved):
Looks good to me, thanks for fixing this.

dpc-aggregation/src/main/resources/application.conf
MEspositoE14s: Yeah, that was entirely my fault.  In a previous PR I set the default, not thinking it would ever get used because the env variable would always be set.  I obviously didn't think it all the way through, though 🤷‍♂️

PR #1942, opened 2023-09-07
dpc-3600: Add new bfd certs to keystore file

Miki543:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3600

## 🛠 Changes

- bb.keystore

## ℹ️ Context for reviewers

Adds new bfd certs to dpc keystore file

## ✅ Acceptance Validation
image

## 🔒 Security Implications


kyeah (commented):
Thanks @Miki543! For local development we only need the `prod-sbx` certificate, could we remove the production cert?

(It's not a huge problem to have it there since BFD prod should reject DPC's client-side certificate anyways, but better to keep the keystore clean.)

kyeah (approved):
Thanks, looks good to me!

I noticed we have an `ops/config/encrypted/bb.keystore` file but I don't see the decrypted filepath referenced anywhere, so I think you've updated the ones that are in use. 

At some point we should probably confirm the encrypted keystore isn't needed and then remove it, but this PR looks good as is 👍

PR #1943, opened 2023-09-11
[DPC-3560] Update ConsentResource#search to take multiple IDs

ashley-weaver:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3560

## 🛠 Changes

(What was added, updated, or removed in this PR.)

## ℹ️ Context for reviewers

(Background context, more in-depth details of the implementation, and anything else you'd like to call out or ask reviewers.)

## ✅ Acceptance Validation

## 🔒 Security Implications


dpc-consent/src/main/java/gov/cms/dpc/consent/resources/ConsentResource.java
```diff
@@ -64,9 +64,10 @@ public Response create(@ApiParam(value = "Consent resource") Consent consent) {
     public List<Consent> search(
             @ApiParam(value = "Consent resource _id") @QueryParam(Consent.SP_RES_ID) Optional<UUID> id,
             @ApiParam(value = "Consent resource identifier") @QueryParam(Consent.SP_IDENTIFIER) Optional<UUID> identifier,
-            @ApiParam(value = "Patient Identifier") @QueryParam(Consent.SP_PATIENT) Optional<String> patientId) {
+            @ApiParam(value = "Patient Identifier") @QueryParam(Consent.SP_PATIENT) Optional<String> patientId,
+            @ApiParam(value = "List of patient identifiers") @QueryParam("patients") Optional<String> patientIds) {
```
kyeah, line 68: Hmm, seeing this actually makes me realize this may not be FHIR-compliant.

@MEspositoE14s — Is there a FHIR-compliant way to allow an endpoint to search by multiple IDs? A quick skim on the FHIR website says we should actually allow this using the existing query parameter `patientId`?

http://hl7.org/fhir/search.html#combining

```
In order to search for unions of results (values joined by "OR"), values can be separated by a comma (",") character. 

 GET [base]/Patient?given=valueGivenA,valueGivenB
```

kyeah: Note that the consent service is internal so it doesn't _really_ have to be FHIR-compliant from a user-facing perspective, but since it's already adhering to the FHIR spec we should try to keep inline with it as much as we can.

kyeah (commented):
Thanks! Flagged one question and included Mike in there since it's a FHIR thing I hadn't considered when writing the ticket requirements. Other comments are just some test suggestions 👋

dpc-consent/src/main/java/gov/cms/dpc/consent/resources/ConsentResource.java
MEspositoE14s: Yeah, accepting a comma separated list is usually the way it's done if it's implemented at all, so it would look something like this...
`https://sandbox.dpc.com/Consent?identifier=mbi1,mbi2,mbi3` or `https://sandbox.dpc.com/Consent?identifier=mbisystem|mbi1,mbisystem|mbi2,mbisystem|mbi3`
There's no rule that you have to support every possible case, though, and most places don't if they support batching parameters at all.  There's a lot of wiggle room in FHIR.

On that note, if we ever get around to reimplementing dpc v2, it might be worth doing the server in [HAPI FHIR](https://hapifhir.io/hapi-fhir/docs/server_plain/rest_operations_search.html#or-relationship-query-parameters) where some of this is supported out of the box.  We already use it client side so it won't be a super big stretch.

dpc-consent/src/test/java/gov/cms/dpc/consent/resources/ConsentResourceTest.java
```diff
@@ -202,6 +201,48 @@ final void searchConsentResource_finds_validPatientParam(String system, String p
...
+
+        final Bundle sut = client
+                .search()
+                .forResource(Consent.class)
+                .where(new StringClientParam("patients").matches().value(patientIds))
+                .encodedJson()
+                .returnBundle(Bundle.class)
+                .execute();
+
+        final Consent found = (Consent) sut.getEntryFirstRep().getResource();
```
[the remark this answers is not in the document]
ashley-weaver: this test is only expecting one result, but I can add another test that will return multiple patients

dpc-consent/src/main/java/gov/cms/dpc/consent/resources/ConsentResource.java
ashley-weaver: so `patientIds` isn't needed at all, and instead I should edit how `patientId` works?

MEspositoE14s: If we want to be strictly FHIR compliant then probably.  The other upside is that the caller in dpc-aggregation is using a FHIR client to make the call, so it's probably easier for whoever gets that ticket.  It adds a little extra work to this ticket, though.

kyeah: yeah it's a little extra work but I think the change should be _fairly_ minimal if we can do that for FHIR compliance. Thank you!

kyeah (approved):
Thank you!

ashley-weaver:
@kyeah I addressed your comments if you want to rereview

kyeah (approved):
thank you!

PR #1971, opened 2023-11-03
[DPC-3660]: Increase code coverage for dpc-common Common Entities 

alex-dzeda:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3660

## 🛠 Changes

Add unit tests to the following files to ensure coverage of methods:
AddressEntity.java
AttributionRelationship.java
ContactEntity.java
ContactPointEntity.java
EndpointEntity.java
NameEntity.java
OrganizationEntity.java
PatientEntity.java
PersonEntity.java
ProviderEntity.java
RosterEntity.java

## ℹ️ Context for reviewers

Entity classes represent a good opportunity to increase code coverage.

## ✅ Acceptance Validation

New unit tests pass, as well as the existing unit-tests continue to pass. 

## 🔒 Security Implications

- [ ] This PR adds a new software dependency or dependencies.
- [ ] This PR modifies or invalidates one or more of our security controls.
- [ ] This PR stores or transmits data that was not stored or transmitted before.
- [ ] This PR requires additional review of its security implications for other reasons.

If any security implications apply, add Jason Ashbaugh (GitHub username: StewGoin) as a reviewer and do not merge this PR without his approval.

PR #1974, opened 2023-11-14
Ensure dpc-common entity tests are run as part of test suite

kyeah:
## 🎫 Ticket

N/A

## 🛠 Changes

- Update usage of org.junit to org.junit.jupiter.api
- Update `PersonEntityTest` to create a new entity for each test
- Auto-formatting updates from the IDE (sorry 😅)

## ℹ️ Context for reviewers

Followup to #1971. @alex-dzeda noticed that the coverage didn't increase in Sonarqube after merging the PR; I tested locally with `make unit-tests` and observed that the new tests were not run as part of the test suite locally or included in local code coverage metrics.

After comparison with other, existing tests, I determined that the issue was using `org.junit.Test` and `org.junit.Asserts` instead of the jupiter equivalents.

I also needed to update `PersonEntityTest` to create a new entity for each test, since it was running into NullPointerException issues with the existing approach.

## ✅ Acceptance Validation

Verified the local code coverage reflected the new tests after running `make unit-tests`. This report is generated under:

```
dpc-common/target/site/jacoco/gov.cms.dpc.common.entities/index.html
```

Screen Shot 2023-11-14 at 12 29 41 AM

### DIFF
--- a/dpc-common/src/test/java/gov/cms/dpc/common/entities/AddressEntityTest.java
+++ b/dpc-common/src/test/java/gov/cms/dpc/common/entities/AddressEntityTest.java
@@ -1,30 +1,30 @@
 package gov.cms.dpc.common.entities;
 
-import static org.junit.Assert.assertEquals;
-import org.junit.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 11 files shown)


alex-dzeda (approved):
Looks good, great catch! Thank you.

PR #1979, opened 2023-11-27
DPC-3740: Set up frontend tooling and CSS pipeline

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3740

### Tooling Changes

- Remove webpacker, which was deprecated and replaced by other tooling in Rails 7
- Replace webpacker SASS --> CSS pipeline with basic sass compilation
- Add ViewComponent, USWDS and Lookbook integration/setup
- Update docker setup to enable live-reload (updating application without rebuilding docker container)
- Remove unused NPM dependencies

### Minor adjustments

- Update application layout to include grid container setup from USWDS (i.e. add global margins for pages)
- Move devise users endpoints to live within the portal scope, to ensure they are routed properly in deployed environments

## ℹ️ Context for reviewers

We talked through some frontend work in refinement, and I realized that we still had more cleanup and groundwork to do than I expected. In particular:

- Webpacker was deprecated in Rails 7 and replaced with a set of different options (import maps, jsbundling-rails + cssbundling-rails, shakapacker.) 
  - For now I've only migrated the SASS --> CSS pipeline and not the JS pipeline (since we aren't leveraging any JS libraries at the moment that we need to compile.)
- We didn't have a pattern set up for Viewcomponent + USWDS. I set up an example Alert component to demonstrate the folder structure with tests and a Lookbook preview.

### Asides

- I had attempted to import the uswds_components library but it was not available on rubygems — opened an issue with them to see if they can fix that.
- I also noted that `minitest` is the test library of choice by Rails now, but I have stuck with an RSpec test since it is already set up and configured for code coverage. We may wish to move to minitest in the near future.

## ✅ Acceptance Validation

Tested development locally. I have a slightly out-of-date gif below of the live reload + Lookbook development process. To replicate...

1. Check out this branch
2. Run `make portal` and `make start-portals`
3. Visit localhost:3001/portal/lookbook

live-reload

## 🔒 Security Implications

- [x] This PR adds a new software dependency or dependencies.

**NOTE** 

@StewGoin -- This PR adds mainly development-only dependencies, _except_ for ViewComponent which was previously discussed and approved in https://jira.cms.gov/browse/DPC-3631.


dpc-portal/docker/entrypoint.sh
meliGuzman, line 1: What's the benefit/reasoning behind this switch? (I'm new to ruby) I read online that ./bin/dev boots up all of your processes from a single command. Is that what bundle exec was doing earlier but now it's more organized/handles more?

dpc-portal/config/routes.rb
meliGuzman, line 1: What does this file do? is that if you search for an endpoint while the app is running and if it identifies the environment as local development then it pops up the lookbook page for you to test UI changes?

docker-compose.portals.yml
```diff
@@ -96,6 +96,7 @@ services:
       dockerfile: dpc-portal/Dockerfile
     image: dpc-web-portal:latest
     volumes:
+      - "./dpc-portal/:/dpc-portal/"
```
meliGuzman, line 99: Could you explain what this volume for the dpc portal is going to be for?

kyeah: This mounts the entire portal directory into the Docker container at runtime, so file changes in your code editor will be immediately noticed and updated in your application. That means you can write your code and get realtime feedback of the changes without having to rebuild the docker image 😄

meliGuzman: That is a life saver! Thank you for explaining. I remember the front end dev in my previous team always had to rebuilt the docker image which could take some generous time

jdettmannnava (approved):

PR #1980, opened 2023-11-29
DPC-3735: Enqueue admin and portal jobs in unique sidekiq queues

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3735

## 🛠 Changes

- Update actionmailer configuration to add jobs in specific queues (this setting is used by Devise)
- Update base ApplicationJob to add jobs in specific queues
- Update sidekiq commands in entrypoints to process specific queues

## ℹ️ Context for reviewers

We are currently experiencing inconsistent issues with emails being sent out from the sandbox portal.

Emails are queued as sidekiq jobs within Redis and then processed by sidekiq workers. However, the dpc-web, dpc-admin, and dpc-portal sidekiq applications are all configured to enqueue and process jobs from the same Redis instance and the same default queue, which means that a job meant for the dpc-web application may be processed by the -admin or -portal worker. This ends in failure.

This PR configures sidekiq and actionmailer to use `admin` and `portal` queues for managing jobs from those applications. 

You can read more about using different queues here: https://github.com/sidekiq/sidekiq/wiki/Advanced-Options and also here specifically for Rails / ActiveJob: https://github.com/sidekiq/sidekiq/wiki/Active-Job#queues

This will allow the following:

- Sign up confirmation and password reset emails should be sent out successfully by dpc-web
- Organization assignment emails should be sent out successfully by dpc-admin
- No jobs should be processed by the dpc-portal application (yet)

## ✅ Acceptance Validation

Tested with current master branch in dev and did not receive password or confirmation codes most of the time.

After deploying, was able to consistently receive password and confirmation codes:

CleanShot 2023-11-29 at 13 25 36@2x

## 🔒 Security Implications

### DIFF
--- a/dpc-admin/app/jobs/application_job.rb
+++ b/dpc-admin/app/jobs/application_job.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class ApplicationJob < ActiveJob::Base
+  # Ensure jobs get sent to a specialized admin queue. Our web applications share
+  # a single Redis instance and process jobs based on their queue name.
+  queue_as :admin
+
   # Automatically retry jobs that encountered a deadlock
   # retry_on ActiveRecord::Deadlocked
 
--- a/dpc-web/config/application.rb
+++ b/dpc-web/config/application.rb
@@ -59,7 +59,11 @@ module DpcWebsite
 
     # Sending mail with`DeliveryJob` has been deprecated. Work has been moved to `MailDeliveryJob`
     config.action_mailer.delivery_job = "ActionMailer::MailDeliveryJob"
-
+    
+    # Ensure mailer jobs get sent to a specialized web queue. Our web applications share
+    # a single Redis instance and process jobs based on their queue name.
+    config.action_mailer.deliver_later_queue_name = "admin"
+    
     config.to_prepare { Devise::Mailer.layout "mailer" }
 
     # Mail throttling
--- a/dpc-web/config/application.rb
+++ b/dpc-web/config/application.rb
@@ -62,7 +62,7 @@ module DpcWebsite
     
     # Ensure mailer jobs get sent to a specialized web queue. Our web applications share
     # a single Redis instance and process jobs based on their queue name.
-    config.action_mailer.deliver_later_queue_name = "admin"
+    config.action_mailer.deliver_later_queue_name = "web"
     
     config.to_prepare { Devise::Mailer.layout "mailer" }
 
--- a/docker-compose.portals.yml
+++ b/docker-compose.portals.yml
@@ -80,7 +80,7 @@ services:
     build:
       context: .
       dockerfile: dpc-admin/Dockerfile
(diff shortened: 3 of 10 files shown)


jdettmannnava (commented):
Should this be set up for dpc-web, just for consistency?

dpc-admin/app/jobs/application_job.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class ApplicationJob < ActiveJob::Base
+  # Ensure jobs get sent to a specialized admin queue. Our web applications share
+  # a single Redis instance and process jobs based on their queue name.
+  queue_as :admin
```
meliGuzman, line 6: @kyeah Just to check my understanding: so redis stores your jobs and was processing the jobs using the default queue before but because of that, the job could be picked up by any worker which is problematic. Why does it end in failure? Is it because each job has a different structure and/or is processed differently based off the specific worker? Could you point to the file(s) that has the code for the workers where it would fail? (I'm a curious :nerd_face:)

kyeah: Hi! Yes, so the main Rails applications will store jobs in the shared Redis instance under the `default` and `mailer` queues. The sidekiq workers for each application were also configured to process jobs from those queues.

However, the sidekiq workers rely on job definitions and database models that are specified within their respective folders, so the `dpc-admin` worker only knows about models and jobs in the `dpc-admin` folder.

So for example, sometimes the `dpc-portal` worker would pick up a job, and since it's a very barebones codebase it would usually fail. This email job from the Admin Portal relies on a user record being passed to it, but the dpc-portal doesn't have the same User model so it'll error out while trying to retrieve the user from the database.

https://github.com/CMSgov/dpc-app/blob/master/dpc-admin/app/mailers/user_mailer.rb

dpc-web/config/application.rb
```diff
@@ -59,7 +59,11 @@ class Application < Rails::Application
 
     # Sending mail with`DeliveryJob` has been deprecated. Work has been moved to `MailDeliveryJob`
     config.action_mailer.delivery_job = "ActionMailer::MailDeliveryJob"
-
+    
+    # Ensure mailer jobs get sent to a specialized web queue. Our web applications share
+    # a single Redis instance and process jobs based on their queue name.
+    config.action_mailer.deliver_later_queue_name = "admin"
```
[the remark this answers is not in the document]
kyeah: oops, thanks!

ashley-weaver (approved):
LGTM!

dpc-admin/app/jobs/application_job.rb
meliGuzman: Ah! Okay, that makes more sense to me now. Thank you, Kevin!

MEspositoE14s (approved):

jdettmannnava (approved):
LGTM

PR #1983, opened 2023-11-29
[DPC-3683] Initial Client Code for CPI API Gateway

chris-ronning-ny:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3683

## 🛠 Changes

This PR establishes a client class that, upon instantiation, authenticates with CMS IDM in the Val environment. As API calls are implemented, this class will continue to manage the access token.

## ℹ️ Context for reviewers

In order to implement an RSpec test for this using VCR, as discussed, the Docker container I'm using needs to be able to connect to the VPN via Zscaler, so that it can make the appropriate API calls. Given that I haven't been able to make this happen, I've opted to exclude unit tests from this PR. Tests will be added in DPC-3684.

## ✅ Acceptance Validation

## 🔒 Security Implications


chris-ronning-ny:
@jdettmannnava I think we should create a separate ticket for Ansible. It's not really a blocker at this point, so no reason to expand the scope.

dpc-portal/docker/entrypoint.sh
```diff
@@ -10,6 +10,9 @@ fi
 if [ "$1" == "portal" ]; then
   # Start the database service (and make accessible outside the Docker container)
   echo "Starting Rails server..."
+  
+  echo "Migrating the database..."
+  bundle exec rails db:migrate
```
meliGuzman, line 15: Ruby newbie Q: Why is this needed? Doesn't look like you're touching the db in your code here?

dpc-portal/spec/lib/cpi_api_gw_client_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+      client_credentials_strategy_instance = instance_double(OAuth2::Strategy::ClientCredentials)
+      allow(OAuth2::Client).to receive(:new).and_return(oauth_client)
+      allow(oauth_client).to receive(:client_credentials).and_return(client_credentials_strategy_instance)
+      expect(client_credentials_strategy_instance).to receive(:get_token).and_return('test_token')
+
+      client = CPIAPIGatewayClient.new
+      expect(client.token).to eq('test_token')
+    end
+  end
+end
```
meliGuzman, line 19: I'm new to ruby so this file throws me off a bit. Is this a way to set methods/functions for your class?  I saw that  "it" is like a method inside a class like how we have them in java. Is this also sort of testing since you have that expect assertion?

dpc-portal/docker/entrypoint.sh
kyeah: This is not really related to the PR/ticket but this line was missing from our initial application setup. We need to run database migrations before starting the Rails server, otherwise Rails will complain 😄

dpc-portal/spec/lib/cpi_api_gw_client_spec.rb
meliGuzman: Thank you, Chris!!!

dpc-portal/lib/cpi_api_gw_client/cpi_api_gw_client.rb
```diff
@@ -0,0 +1,23 @@
...
+# A client for requests to the CPI API Gateway
+class CPIAPIGatewayClient
+  attr_accessor :token
+
+  def initialize
+    client_id = ENV.fetch('CPI_API_GW_CLIENT_ID', nil)
+    client_secret = ENV.fetch('CPI_API_GW_CLIENT_SECRET', nil)
+    @client = OAuth2::Client.new(client_id, client_secret,
+                                 site: 'https://impl.idp.idm.cms.gov/',
+                                 token_url: '/oauth2/aus2151jb0hszrbLU297/v1/token')
```
MEspositoE14s, line 14: My only question is if it will be a problem if `site` ends with a "/" and `token_url` begins with one?  Will that make the final token url `https://impl.idp.idm.cms.gov//oauth2/aus2151jb0hszrbLU297/v1/token`?

MEspositoE14s (approved):
I don't know how qualified I am to review Rails code, but it seems pretty straightforward and all the tests work, so looks good to me!

dpc-portal/lib/cpi_api_gw_client/cpi_api_gw_client.rb
chris-ronning-ny: Nope, this is per spec. See https://gitlab.com/oauth-xx/oauth2#customize-authorize_url-and-token_url

PR #1986, opened 2023-12-01
DPC-3740: Install lookbook in deployed environments

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3740

## 🛠 Changes

- Move lookbook dependency from :development group to :production group in Gemfile

## ℹ️ Context for reviewers

In #1979 we installed lookbook and configured it to deploy in the dev environment. However, the lookbook dependency is only installed in the :development RAILS_ENV group. It needs to be installed in the :production group, which is the RAILS_ENV used in all deployed environments.

In the future we should make sure this dependency is not installed in sandbox/prod, but until we get to production launch readiness we can keep this in the production group.

## 🔒 Security Implications

- [x] This PR adds a new software dependency or dependencies.

cc. @StewGoin

### DIFF
--- a/dpc-portal/Gemfile
+++ b/dpc-portal/Gemfile
@@ -29,9 +29,11 @@ gem 'health_check'
 gem 'jbuilder', '~> 2.7'
 gem 'kaminari'
 gem 'lograge'
+gem 'lookbook', '>= 2.1.1' # install lookbook so it can be deployed in lower environments.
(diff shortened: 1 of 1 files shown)


chris-ronning-ny (approved):

kyeah:
CleanShot 2023-12-01 at 13 58 31@2x

Healthy deployment!

PR #1996, opened 2023-12-14
Fix Portal <> API local connection and organize Makefile

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-xxx

### Organizational

- Organize Makefile into sections
- Replace lots of `.PHONY` lines with descriptive comments
- Remove `start-local` and `start-local-api` commands (this is unused/the stuff in the docker-compose-local.yml file was merged into the default docker-compose file)
- Remove references to "dpc v1"
- Rename `make start-core-dependencies` to `make start-db` because that seems more straightforward

### Functional Changes

- Add `make api` target for building the Java services/modules without running tests
- Add utility targets for `start-web`, `start-admin`, `start-portal` because I don't always want to run all of them together 🙃 
- Add utility targets for `make psql` (to get into the database) and `make portal-console` (to open a Rails console)
- Remove all custom namespacing when running docker-compose (e.g. `docker-compose -p dpc-v2`). Different namespaces created different docker networks, so the Rails apps were not able to talk to the API services.

## ℹ️ Context for reviewers

Some of this is organizational, but this also introduces a few utilities to streamline development.

This also fixes the broken connectivity between the Rails apps and the API services locally when using `make start-portals`, since it was creating the Rails apps in a separate docker network.

## ✅ Acceptance Validation

Ran a bunch of make commands for local development. Note that if you already had Rails applications started using `make start-portals`, you'll want to bring them down manually before running the new commands — otherwise Docker will try to spin up two parallel Rails apps that will probably conflict on the exposed port.

```
$ docker-compose -p dpc-v2 -f docker-compose.yml -f docker-compose.portals.yml down
```

Full deployment also confirmed successful.

## 🔒 Security Implications


Makefile
```diff
@@ -9,33 +12,107 @@ venv/bin/activate: requirements.txt
...
 website:
 	@docker build -f dpc-web/Dockerfile . -t dpc-web
 
-.PHONY: admin
+admin: ## Builds the sandbox admin website
 admin:
 	@docker build -f dpc-admin/Dockerfile . -t dpc-web-admin
 
-.PHONY: portal
+portal: ## Builds the DPC portal
```
chris-ronning-ny, line 52: Just some feedback, I was a bit confused when I ran `make portal` in my local and it didn't result in a container being built. I did figure out that `make start-portal` is the command I was looking for. Maybe some commentary to clarify the practical purpose of these commands would be helpful in the future.

chris-ronning-ny (commented):
Tested some of these locally, and generally think this will make the dev cycle for the portal a lot smoother. Thanks for doing this! See one bit of feedback

Also to sanity check -- it might be a good idea to deploy this branch to dev or test, to make sure that this interface change doesn't break CI.

chris-ronning-ny (approved):

PR #2014, opened 2024-01-10
Automatically generate golden macaroon on local dev startup

kyeah:
## 🎫 Ticket

N/A

## 🛠 Changes

- Update docker entrypoint scripts to generate golden macaroon on startup

## ℹ️ Context for reviewers

This removes a manual step required to call the API and generate golden macaroons. If the API is running when a frontend application starts, we will automatically generate a new macaroon If the API is not running, we will continue as normal and log a message noting that there is no macaroon.

## ✅ Acceptance Validation

Ran `make portal && make start-portal` with and without a golden macaroon set in the docker-compose file and observed both behaviors.

## 🔒 Security Implications

### DIFF
--- a/docker-compose.portals.yml
+++ b/docker-compose.portals.yml
@@ -22,6 +22,9 @@ services:
       - REDIS_URL=redis://redis
       - DATABASE_URL=postgresql://db/dpc-website_development
       - TEST_DATABASE_URL=postgresql://db/dpc-website_test
+      - API_METADATA_URL=http://api:3002/v1
+      - API_ADMIN_URL=http://api:9900
+      - GOLDEN_MACAROON=${GOLDEN_MACAROON}
       - DB_USER=postgres
       - DB_PASS=dpc-safe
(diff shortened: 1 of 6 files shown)


docker-compose.portals.yml
```diff
@@ -22,6 +22,9 @@ services:
       - REDIS_URL=redis://redis
       - DATABASE_URL=postgresql://db/dpc-website_development
       - TEST_DATABASE_URL=postgresql://db/dpc-website_test
+      - API_METADATA_URL=http://api:3002/v1
+      - API_ADMIN_URL=http://api:9900
```
chris-ronning-ny, line 26: How does this work? Would a dev need to update their hosts file?

chris-ronning-ny (approved):
nice!

docker-compose.portals.yml
kyeah: Services within the docker-compose files can find each other through their service names — since `api` is a service within the `docker-compose.yml` file, The portal services can find it using the `api` hostname 🌟

PR #2021, opened 2024-01-16
[DPC-3783] Update JobBatchProcessor to only check most recent consent

chris-ronning-ny:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3783

## 🛠 Changes

The JobBatchProcessor has been updated so that, when checking if a patient is currently opted out, it only checks the most recent consent record. This way, if a patient decides to opt back in, the user will not be incorrectly flagged as opted out.

## ℹ️ Context for reviewers

Previous code checked the entire set of consent records for an opt out, effectively ignoring any opt-ins.

## ✅ Acceptance Validation

## 🔒 Security Implications

### DIFF
--- a/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/JobBatchProcessor.java
+++ b/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/JobBatchProcessor.java
@@ -224,14 +224,14 @@ public class JobBatchProcessor {
     private boolean isOptedOut(Optional<List<ConsentResult>> consentResultsOptional) {
         if (consentResultsOptional.isPresent()) {
             final List<ConsentResult> consentResults = consentResultsOptional.get();
-            long optOutCount = consentResults.stream().filter(consentResult -> {
-                final boolean isActive = consentResult.isActive();
-                final boolean isOptOut = ConsentResult.PolicyType.OPT_OUT.equals(consentResult.getPolicyType());
-                final boolean isFutureConsent = consentResult.getConsentDate().after(new Date());
-                return isActive && isOptOut && !isFutureConsent;
-            }).count();
-            return optOutCount > 0;
+            final ConsentResult latestConsent = Collections.max(consentResults, Comparator.comparing(consent -> consent.getConsentDate()));
+            final boolean isActive = latestConsent.isActive();
+            final boolean isOptOut = ConsentResult.PolicyType.OPT_OUT.equals(latestConsent.getPolicyType());
+            final boolean isFutureConsent = latestConsent.getConsentDate().after(new Date());
+            return isActive && isOptOut && !isFutureConsent;
         }
+        // This should never execute. Log an error.
+        logger.error("Consent result is unexpectedly empty.");
         return true;
     }
 


dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/JobBatchProcessor.java
```diff
@@ -224,14 +224,14 @@ private Pair<Optional<List<ConsentResult>>, Optional<OperationOutcome>> getConse
...
         if (consentResultsOptional.isPresent()) {
             final List<ConsentResult> consentResults = consentResultsOptional.get();
-            long optOutCount = consentResults.stream().filter(consentResult -> {
-                final boolean isActive = consentResult.isActive();
-                final boolean isOptOut = ConsentResult.PolicyType.OPT_OUT.equals(consentResult.getPolicyType());
-                final boolean isFutureConsent = consentResult.getConsentDate().after(new Date());
-                return isActive && isOptOut && !isFutureConsent;
-            }).count();
-            return optOutCount > 0;
+            final ConsentResult latestConsent = Collections.max(consentResults, Comparator.comparing(consent -> consent.getConsentDate()));
```
jdettmannnava, line 227: What happens if consentResults is empty, which should be the common case? Should we check for empty first and return false?

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/JobBatchProcessor.java
```diff
@@ -224,14 +224,14 @@ private Pair<Optional<List<ConsentResult>>, Optional<OperationOutcome>> getConse
...
+            final boolean isActive = latestConsent.isActive();
+            final boolean isOptOut = ConsentResult.PolicyType.OPT_OUT.equals(latestConsent.getPolicyType());
+            final boolean isFutureConsent = latestConsent.getConsentDate().after(new Date());
+            return isActive && isOptOut && !isFutureConsent;
         }
+        // This should never execute. Log an error.
+        logger.error("Consent result is unexpectedly empty.");
```
jdettmannnava, line 234: Not really empty; more like null

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/JobBatchProcessor.java
chris-ronning-ny: I would expect that we would never actually execute [here](https://github.com/CMSgov/dpc-app/blob/cr/dpc-3783/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/JobBatchProcessor.java#L75), so this function would never run. Since I can't really confirm that though, I think I'll add a check anyways, just to be safe.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/JobBatchProcessor.java
```diff
@@ -224,14 +224,17 @@ private Pair<Optional<List<ConsentResult>>, Optional<OperationOutcome>> getConse
...
-            }).count();
-            return optOutCount > 0;
+            if (consentResults.isEmpty()) {
+                return false;
+            }
+            final ConsentResult latestConsent = Collections.max(consentResults, Comparator.comparing(consent -> consent.getConsentDate()));
+            final boolean isActive = latestConsent.isActive();
+            final boolean isOptOut = ConsentResult.PolicyType.OPT_OUT.equals(latestConsent.getPolicyType());
+            final boolean isFutureConsent = latestConsent.getConsentDate().after(new Date());
+            return isActive && isOptOut && !isFutureConsent;
```
MEspositoE14s, line 234: I know this goes against what's specified in the ticket, but if the most recent ConsentResult is post dated for the future, should we ignore it and look for one that's currently active instead.  

I.E. If the most recent consent result has a start date of next week and is active, the beneficiary technically isn't opted out until next week.  Should we allow their data to be downloaded until then?

jdettmannnava: Isn't that what isFutureConsent means?

MEspositoE14s: I think so, although I have no idea if they'll actually ever send us a future opt out.  It might be better to merge this as is, and then update dpc-consent or the consent service to filter out future opt outs instead.

chris-ronning-ny: It's also not super tricky to just filter out future opt outs within the body of this function. I can foresee a situation in which we want the consent service to include future opt-outs; possibly better to make filtering the responsibility of the caller.

jdettmannnava: Yes, this feels a bit over-engineered already, as I think we hard-code the opt-out-preference as Active and as of now. Adding future-preference-change should be a complete feature added "as needed". I think we can leave the checks as is.

chris-ronning-ny: Sounds good, I'll leave this as is for now. @kyeah flagging you for visibility.

jdettmannnava (approved):
LGTM!

PR #2032, opened 2024-02-07
PLT-295 Add opt-out-import lambda code from dpc-ops

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-295

## 🛠 Changes

Added code for the opt-out-import lambda.

## ℹ️ Context for reviewers

Code for the opt-out-import lambda is being [removed from dpc-ops](https://github.com/CMSgov/dpc-ops/pull/660) and added to dpc-app to allow separation of concerns. The DPC team is responsible for the code in the lambda, and will test and deploy that code from this dpc-app repo.

## ✅ Acceptance Validation

The next PR will put workflows into place for tests and deploys.

## 🔒 Security Implications

- [x] This PR requires additional review of its security implications for other reasons.

Given that we are moving software from a private repo to a public one, I have reviewed for any sensitive information, but I've added @StewGoin for awareness.

PR #2034, opened 2024-02-12
PLT-295 Add workflows for opt-out-import

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-295

## 🛠 Changes

Added test and deploy workflows for the opt-out-import lambda.

## ℹ️ Context for reviewers

These are initial versions of the workflows for the opt-out-import lambda, lightly modified from similar code in https://github.com/CMSgov/ab2d-lambdas/tree/main/.github/workflows. They can serve as examples and should work for the most part, but will be owned by the DPC team going forward. For example, the team may wish to move more of the logic into the Makefile and reference `make` commands in the workflows.

I've also added preliminary "ignores" to ci-workflow.yml to avoid api, web, admin, and portal tests on changes that only affect opt-out-import. I would recommend ticketing up work to split that workflow up so that api tests only run when api paths are modified, etc.

## ✅ Acceptance Validation

These will fail on merge without the code in main from #2032, but I would like to merge them in first anyway. That will allow me to test and tweak in #2032 before merging.

## 🔒 Security Implications

- [x] This PR stores or transmits data that was not stored or transmitted before.

This PR introduces workflows that interact directly with our AWS accounts from the dpc-app repo. To do so, the existing OIDC provider is used for role assumption, and account role ARNs are saved as variables in this repo.


StewGoin (approved):
👍

.github/workflows/ci-workflow.yml
```diff
@@ -1,6 +1,11 @@
 name: "DPC CI Workflow"
 
-on: [push]
```
kyeah, line 3: Admittedly I did sometimes find it helpful to run these on every push, but I recognize that is extremely resource-wasteful and I can just open a draft PR 😄

.github/workflows/ci-workflow.yml
```diff
@@ -1,6 +1,11 @@
...
+on:
+  pull_request:
+    paths-ignore:
+      - .github/workflows/opt-out-import-*
+      - lambda/**
+  workflow_dispatch: # Allow manual trigger
```
kyeah, line 8: Will this workflow have the `permission` block added at a later date?

kyeah (approved):
Looks fine to me. Is it because of the infrastructure we have set up that we aren't able to use a single workflow with an environment variable for running? Or is that an intentional choice to avoid mixing runs from different environments into the same Actions tab view?

.github/workflows/ci-workflow.yml
gsf: Draft PR will work if you have a branch going that you isn't ready for review. The workflow_dispatch trigger below will also allow you to run it on any branch at any time.

.github/workflows/ci-workflow.yml
gsf: The `permissions` block included in the other workflows is only needed if the workflow uses the repo token to assume a role in one of our AWS accounts.

gsf:
@kyeah The workflow to deploy to test is separate so it can trigger on merge to main. The deploy to dev differs from prod and sbx since (like test) it builds and uploads rather than copying from the test environment. Prod and sbx could be one workflow with an env input but keeping them separate makes it quicker to look at actions and see just the history of deploys to either prod or sbx.

PR #2037, opened 2024-02-13
Me/dpc 3874 fix ci app

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3874

## 🛠 Changes

Drop `dpc-app_pgdata` both before and after running integration tests.

## ℹ️ Context for reviewers

Normally we would drop this volume after the integration tests run in `dpc-test.sh`, but when a test fails it would get skipped.  This would leave test data from the current failed run in the DB for the next run, and it was causing future integration tests and the `DPC - Docker Build` pipeline to occasionally fail.  See [here](https://management.dpc.cms.gov/job/DPC%20-%20Docker%20Build/3528/console) for example.  Now we drop the DB before the tests run as well, to make sure the DB is clean.

Note: The specific tests that were failing had to do with opt outs.  We make sure we can load a particular patient resource, then opt them out, then make sure we can't load them.  If the DB isn't wiped, the patient starts already opted out and we're not able to load them at all.

## ✅ Acceptance Validation

Pipeline is running successfully now.

## 🔒 Security Implications

### DIFF
--- a/dpc-test.sh
+++ b/dpc-test.sh
@@ -64,6 +64,7 @@ docker-compose up start_api_dependencies
 docker-compose up --exit-code-from tests tests
 
 docker-compose down
+docker volume rm dpc-app_pgdata
 docker-compose up start_core_dependencies
 docker-compose up start_api_dependencies
 
--- a/dpc-test.sh
+++ b/dpc-test.sh
@@ -56,6 +56,7 @@ if [ -n "$REPORT_COVERAGE" ]; then
 fi
 
 docker-compose down
+docker volume rm dpc-app_pgdata
 docker-compose up start_core_dependencies
 docker-compose up start_api_dependencies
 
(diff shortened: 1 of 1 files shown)


dpc-test.sh
```diff
@@ -56,6 +56,7 @@ if [ -n "$REPORT_COVERAGE" ]; then
 fi
 
 docker-compose down
+docker volume rm dpc-app_pgdata
```
jdettmannnava, line 59: Won't this make the volume completely useless?

MEspositoE14s: In terms of the these particular tests, yeah.  I like having it in the docker-compose for running integration tests locally, though.  It makes them easier to debug when I can look in the DB and see what happened.

jdettmannnava (approved):
LGTM

PR #2040, opened 2024-02-20
DPC-3858: Persist user on login

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3858
## 🛠 Changes

Actually saving a user after logging in via login-dot-gov.

## ℹ️ Context for reviewers

Also saves the hash of the ssn, though not technically part of the ticket

## ✅ Acceptance Validation

Manual testing shows user created in db

### DIFF
--- a/dpc-portal/app/controllers/login_dot_gov_controller.rb
+++ b/dpc-portal/app/controllers/login_dot_gov_controller.rb
@@ -6,9 +6,10 @@ class LoginDotGovController < Devise::OmniauthCallbacksController
 
   def openid_connect
     auth = request.env['omniauth.auth']
-    user = User.new(email: auth.info.email,
-                    given_name: auth.extra.raw_info.given_name,
-                    family_name: auth.extra.raw_info.family_name)
+    user = User.find_or_create_by(provider: auth.provider, uid: auth.uid) do |user_to_create|
+      assign_user_properties(user_to_create, auth)
+    end
+    ssn_cookie(auth.extra.raw_info.social_security_number)
     sign_in(:user, user)
   end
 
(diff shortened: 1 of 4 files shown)


dpc-portal/app/controllers/login_dot_gov_controller.rb
```diff
@@ -6,9 +6,10 @@ class LoginDotGovController < Devise::OmniauthCallbacksController
 
   def openid_connect
     auth = request.env['omniauth.auth']
-    user = User.new(email: auth.info.email,
-                    given_name: auth.extra.raw_info.given_name,
-                    family_name: auth.extra.raw_info.family_name)
+    user = User.find_or_create_by(provider: auth.provider, uid: auth.uid) do |user_to_create|
+      assign_user_properties(user_to_create, auth)
+    end
+    ssn_cookie(auth.extra.raw_info.social_security_number)
```
kyeah, line 12: Should this be removed for this ticket? It's not being used yet and I think we have some open questions still about whether we need to do this or if we can call the OIDC/Login.gov user info endpoint as-needed w/ the OIDC session credentials as part of the "Add new org" flow.

(Alberto mentioned the [user-info endpoint](https://developers.login.gov/oidc/user-info/) and I agree that we should investigate that further before we move to store the SSN in cookies 🍪)

dpc-portal/app/controllers/login_dot_gov_controller.rb
```diff
@@ -21,4 +22,23 @@ def failure
...
     end
   end
+
+  private
+
+  def assign_user_properties(user, auth)
+    user.email = auth.info.email
+    user.given_name = auth.extra.raw_info.given_name
+    user.family_name = auth.extra.raw_info.family_name
+    user.password = user.password_confirmation = Devise.friendly_token[0, 20]
```
kyeah, line 32: Should we add a comment here that this won't be used/usable in production and is only necessary to make Devise happy?

dpc-portal/app/controllers/login_dot_gov_controller.rb
jdettmannnava: So, I think the user-info endpoint is called by the openid connect gem as part of the magic flow by which we get information after the user is redirected to our site. I think we would need to cache the authorization token we get as part of this process, which I think has a [lifespan of 6 minutes](https://developers.login.gov/oidc/token/), if we wanted to call it again. This strikes me as somehow worse than saving the hashed ssn in a cookie, but that is just one dev's opinion.

I don't mind taking this out and deferring the decision, though.

kyeah (approved):
thank you!

dpc-portal/app/controllers/login_dot_gov_controller.rb
kyeah: Ah, thank you! Good point; I think we'll want to look in further detail/run a spike on this. I believe there may be a refresh_token in the gem that can be reused for those short-lived access tokens (on a previous project, we did something similar although that was for oauth2.) But yes, something to defer to another ticket.

PR #2047, opened 2024-02-22
DPC-3864 AO organization detail view

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3864

## 🛠 Changes

- Component for showing AO organization view page (an organization's credential delegates)
- CdOrgLink class to mock until it has an [ActiveRecord](https://confluence.cms.gov/x/GBGkOg)

## ℹ️ Context for reviewers

Uses same controller as CD sees credentials, so I toggle based on an "ao" parameter. Will be replaced when a user needs to log in and have permissions to do things.

## ✅ Acceptance Validation

Compare Screenshots with those in [ticket](https://jira.cms.gov/browse/DPC-3864):
empty
active
pending


dpc-portal/app/components/page/credential_delegate/list_component.html.erb
```diff
@@ -0,0 +1,34 @@
+<div>
+  <h1><%= @organization.name %></h1>
+  <div class="margin-bottom-5">NPI: <%= @organization.npi %></div>
```
kyeah, line 3: Hmm these don't seem to show up correctly in the preview (although I assume the actual component code here is correct.)

CleanShot 2024-02-27 at 14 01 46@2x

kyeah: Although I suspect the card/table part of this page will end up being a sub-component used as part of the larger Organization Detail view in the future, so I don't know if it'll be necessary to include the Organization name/npi here at all. May be simpler to remove it with that assumption/future state in mind (e.g. I know that this the list of CDs will eventually move into being a tab in the Org Detail view.)

kyeah (commented):
Looks pretty good to me, made a note about the organization info at the top. thanks!

dpc-portal/app/components/page/credential_delegate/list_component.html.erb
jdettmannnava: It was an odd class initialization issue I hadn't seen before.

kyeah (approved):
The Portal CI passed so I'm approving 😄 Reran your DPC API tests, looked like a temporary failure with BB2.0.

PR #2051, opened 2024-02-29
Remove codeclimate reporting

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-xxx

## 🛠 Changes

- Remove CodeClimate
- Remove tee'ing to local log files since we're on Fargate and streaming directly to Cloudwatch --> Splunk

## ℹ️ Context for reviewers

Remove CodeClimate as discussed internally, since...

1. We already rely on Sonarqube
2. Other teams don't use CodeClimate
3. Dependabot PRs don't have CC configured (so they can't be merged) and we aren't sure who still has admin access to CC

## ✅ Acceptance Validation

Unit tests

## 🔒 Security Implications

Already OK'd by @StewGoin separately, but tagging for awareness.


dpc-test.sh
```diff
@@ -18,41 +18,28 @@ function _finally {
 trap _finally EXIT
 
 if [ -n "$REPORT_COVERAGE" ]; then
-   echo "┌──────────────────────────────────────────┐"
```
ashley-weaver, line 21: should the line on the top and bottom of the box be shortened too?

kyeah: done, ty!

gsf:
Commenting after the fact here to say I'm highly in favor of dropping CodeClimate. Thanks for dropping it and for the other cleanup here.

PR #2061, opened 2024-03-06
[DPC-3905] Add admin endpoint for fetching an organization by NPI

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3905

## 🛠 Changes

Renamed the current `/Admin/Organization` end point to indicate that it's searching by NPI and not id.

## ℹ️ Context for reviewers

Despite the fact that it said it was searching by id, the admin end point we already have is actually searching by NPI.  The initial tests we had never caught it because they were only checking for a 200, not verifying the results.  I added a new integration test to AdminResourceTest that performs the search by NPI and ensures that a valid org is returned.

## ✅ Acceptance Validation

Passes new integration tests.

## 🔒 Security Implications

### DIFF
--- a/dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/AdminResourceTest.java
+++ b/dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/AdminResourceTest.java
@@ -1,22 +1,25 @@
 package gov.cms.dpc.api.resources.v1;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
+import ca.uhn.fhir.rest.client.api.IGenericClient;
+import gov.cms.dpc.api.APITestHelpers;
+import gov.cms.dpc.api.AbstractSecureApplicationTest;
+import gov.cms.dpc.fhir.DPCIdentifierSystem;
+import gov.cms.dpc.testing.APIAuthHelpers;
+import org.apache.http.HttpHeaders;
+import org.eclipse.jetty.http.HttpStatus;
+import org.hl7.fhir.dstu3.model.Bundle;
+import org.hl7.fhir.dstu3.model.Organization;
+import org.junit.jupiter.api.Test;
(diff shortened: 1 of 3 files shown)


dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/AdminResourceTest.java
```diff
@@ -53,4 +38,26 @@ void testNoGoldenMacaroon() throws IOException, URISyntaxException {
         assertEquals(HttpStatus.UNAUTHORIZED_401, conn.getResponseCode());
         conn.disconnect();
     }
+
+    @Test
+    void testSearchByNpi() {
+        IGenericClient client = APIAuthHelpers.buildAdminClient(ctx, getAdminResourceURL(), GOLDEN_MACAROON, false);
+
+        Map<String, List<String>> searchParams = new HashMap<>();
+        searchParams.put("npis", Collections.singletonList(DPCIdentifierSystem.MBI.getSystem() + "|" + APITestHelpers.ORGANIZATION_NPI));
```
kyeah, line 47: This should be NPPES as the system, right?

```suggestion
        searchParams.put("npis", Collections.singletonList(DPCIdentifierSystem.NPPES.getSystem() + "|" + APITestHelpers.ORGANIZATION_NPI));
```

kyeah (approved):
ah...good catch. since we're storing the NPI on the portal side, we don't actually need to search by organization ID. thx!

dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/AdminResourceTest.java
MEspositoE14s: Yep, you're right.  

Technically it doesn't matter much, though, because when the request eventually gets to dpc-attribution we strip the system off, anyway.  The request will fail as invalid if we don't have a system at all, but we can just put some random text there and it'll work, regardless.

PR #2055, opened 2024-03-06
PLT-340 Add opt out export

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-340

## 🛠 Changes

Added opt-out-export code from dpc-ops.

## ℹ️ Context for reviewers

Moving opt-out-export code over from dpc-ops as we did with opt-out-import. Along with deploy workflows, I've added initial attempts at integration tests.

## ✅ Acceptance Validation

See checks, though additional PRs may be required to get deploys and tests working as expected.

## 🔒 Security Implications

- [x] This PR requires additional review of its security implications for other reasons.

As in #2032, I've added @StewGoin since we are moving software from a private repo to a public one. I have reviewed for any sensitive information.

PR #2064, opened 2024-03-11
DPC-3888: Updates for successful export lambda execution

kyeah:
## 🎫 Ticket

[DPC-3888](https://jira.cms.gov/browse/DPC-3888)

## 🛠 Changes

- Pull BFD assume role ARN from parameter store instead of environment variable
- Only use BFD assume role for S3 upload — use regular AWS session for DPC parameter store retrieval
- Avoid writing local file on read-only lambda. Instead, write directly to in-memory bytes buffer and upload to S3

## ℹ️ Context for reviewers

A few changes to make the export lambda pass when running it on real infrastructure. Recommend reviewing with "Hide whitespace changes" enabled since my editor did some auto-formatting 🙈 

This is a followup to #2055. Split out to make it easier to review actual code changes instead of just moving code between the repos.

## ✅ Acceptance Validation

Ran the lambda manually with an empty event in AWS Console.

## 🔒 Security Implications


lambda/opt-out-export/main.go
```diff
@@ -168,6 +169,17 @@ func generateAlignmentFileName(now time.Time) (string) {
...
 
+func getAssumeRoleSession(session *session.Session) (*session.Session, error) {
+	parameterName := fmt.Sprintf("/opt-out-import/dpc/%s/bfd-bucket-role-arn", os.Getenv("ENV"))
+	assumeRoleArn, err := getSecret(session, parameterName)
+
+	if err != nil {
+		return nil, fmt.Errorf("Failed to retrieve assume role arn: %w", err)
+	}
+
+	return newSession(assumeRoleArn)
```
MEspositoE14s, line 180: To get everything running locally, I had to switch to this:

```suggestion
	if isTesting {
		return getAwsSession()
	} else {
		return newSession(assumeRoleArn)
	}
```

and then add: 
`aws --endpoint-url=http://localhost:4566 ssm put-parameter \
    --name "/opt-out-import/dpc/local/bfd-bucket-role-arn" \
    --value "assume_role_arn_123456789" \
    --type "String"`
and `aws --endpoint-url=http://localhost:4566 ssm get-parameter --name "/opt-out-import/dpc/local/bfd-bucket-role-arn"`
to populate_aws.sh

lambda/opt-out-export/dpcaws/s3.go
```diff
@@ -13,26 +15,20 @@ var osOpen = os.Open
...
-
+func UploadFileToS3(s *session.Session, fileName string, buff bytes.Buffer, s3Bucket string, s3Path string) error {
 	// Upload file to bucket
 	uploader := newUploader(s)
 
 	_, s3Err := upload(*uploader, &s3manager.UploadInput{
 		Bucket: aws.String(s3Bucket),
 		Key:    aws.String(s3Path + "/" + fileName),
-		Body:   file,
+		Body:   bytes.NewReader(buff.Bytes()),
```
MEspositoE14s, line 27: Part of me was worried about trying to do this with a buffer because I thought it might get too big once we start adding lots of patients.  Then I just looked up how much memory is available to a lambda function and it can go pretty high, so I don't think we'll have an issue!

MEspositoE14s (approved):
Looks pretty good to me outside of the one failing test.

PR #2065, opened 2024-03-11
DPC-3888: Handle SQS/SNS-wrapped event notifications for Opt Out Import

kyeah:
## 🎫 Ticket

[DPC-3888](https://jira.cms.gov/browse/DPC-3888)

## 🛠 Changes

- Unwrap S3 events from SNS events from SQS events

## ℹ️ Context for reviewers

The pipeline from S3 --> Lambda includes hops from S3 --> SNS --> SQS --> Lambda, and each of those steps end up wrapping the previous message in another event wrapper.

This updates our parsing code to unwrap the S3 event from those other wrappers.

## ✅ Acceptance Validation

I have manually verified in AWS that the uploaded integration test file is parsed correctly.

CleanShot 2024-03-11 at 16 17 46@2x

## 🔒 Security Implications

### DIFF
--- a/lambda/opt-out-import/parsers.go
+++ b/lambda/opt-out-import/parsers.go
@@ -102,7 +102,7 @@ func ParseSQSEvent(event events.SQSEvent) (*events.S3Event, error) {
 	unmarshalTypeErr := new(json.UnmarshalTypeError)
 	if errors.As(err, &unmarshalTypeErr) {
 		log.Warn("Skipping event due to unrecognized format for SNS")
-		return nil, err
+		return nil, nil
 	} else if err != nil {
 		return nil, err
 	}
--- a/lambda/opt-out-import/parsers.go
+++ b/lambda/opt-out-import/parsers.go
@@ -3,14 +3,17 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"strings"
 	"time"
 
+	"github.com/aws/aws-lambda-go/events"
 	"github.com/google/uuid"
 	"github.com/ianlopshire/go-fixedwidth"
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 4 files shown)


lambda/opt-out-import/parsers_test.go
```diff
@@ -128,3 +134,39 @@ func TestParseRecord_InvalidData(t *testing.T) {
...
+		fmt.Println(err)
+	}
+
+	val, err := json.Marshal(s3event)
+
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	body := fmt.Sprintf("{\"Type\" : \"Notification\",\n  \"MessageId\" : \"123456-1234-1234-1234-6e06896db643\",\n  \"TopicArn\" : \"my-topic\",\n  \"Subject\" : \"Amazon S3 Notification\",\n  \"Message\" : %s}", strconv.Quote(string(val[:])))
```
kyeah, line 162: I can probably do this more cleanly, but using this for now since it works.

lambda/opt-out-import/parsers.go
```diff
@@ -90,3 +93,30 @@ func ConvertSharingPreference(pref string) (string, error) {
 		return "", errors.New(fmt.Sprintf("Unexpected value %s for sharing preference", pref))
 	}
 }
+
+// TODO: Iterate over records
```
chris-ronning-ny, line 97: Just out of curiosity, is this particularly likely? Seems like we'll be handling one record at a time as they come in.

chris-ronning-ny (approved):
I'm surprised you had to implement `ParseSQSEvent` manually! Looks good to me, I would love it if we had a built-in way to do it, but this looks like it'll do the job.

lambda/opt-out-import/parsers.go
kyeah: yeah, I'm not worried about it right now. More of a "just in case!" scenario to keep track of for the future so we aren't surprised in 16 months if it ever happens 😅

PR #2066, opened 2024-03-13
Fix USWDS asset paths

kyeah:
## 🎫 Ticket

N/A

## 🛠 Changes

- Fix USWDS asset paths to match compiled paths
- Update comments to be more accurate

## ℹ️ Context for reviewers

This fixes our USWDS initializer to add the correct path to the Rails asset paths. This will allow us to use tags like `image_tag` and `asset_path` without having to specify the `@uswds/uswds/dist` prefix for all of the assets.

Previously, this initializer didn't actually do anything because it had the wrong prefix. We didn't notice because we aren't referencing USWDS images or fonts in any of our custom HTML. 

## ✅ Acceptance Validation

Verified locally in Lookbook by adding the location_city.svg icon from USWDS:

CleanShot 2024-03-13 at 09 35 04@2x

```
<%= image_tag 'img/usa-icons/location_city.svg' %>
```

## 🔒 Security Implications


kyeah:
Going to close this for now since it's not a priority.

PR #2072, opened 2024-03-14
DPC-3822 add authorization

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3822
## 🛠 Changes

- Controllers load organization based on ProviderOrganization
- List of organizations only show organizations with either AO or CD role
- Controllers verify user has authorization (as AO or CD) on actions related to the organization
- Organization class removed in favor of ProviderOrganization
- MockOrg removed from lookbook previews
- Routes use ProviderOrganization id rather than dpc-api org id

## ℹ️ Context for reviewers

- AOs can do anything a CD can do
- Because of changes to organizations in specs, I **strongly** advise turning off whitespace in diff.

## ✅ Acceptance Validation

- Many rspec tests
- Manual testing  of functionality locally
- Manual review of lookbook previews locally

## 🔒 Security Implications


dpc-portal/app/models/cd_org_link.rb
```diff
@@ -9,7 +9,7 @@ class CdOrgLink < ApplicationRecord
   def show_attributes
     { full_name: "#{user.given_name} #{user.family_name}",
       email: user.email,
-      verification_code: invitation.verification_code,
+      verification_code: invitation&.verification_code,
```
ashley-weaver, line 12: for my own education, what does the & do here?

jdettmannnava: It only calls if invitation is not nil

dpc-portal/app/controllers/application_controller.rb
```diff
@@ -1,4 +1,26 @@
 # frozen_string_literal: true
 
+# Parent class of all controllers
 class ApplicationController < ActionController::Base
+  before_action :block_prod_sbx
```
kyeah, line 5: 👍

dpc-portal/app/controllers/application_controller.rb
```diff
@@ -1,4 +1,26 @@
...
+  private
+
+  def block_prod_sbx
+    redirect_to root_url if ENV.fetch('ENV', nil) == 'prod-sbx'
+  end
+
+  def load_organization
+    @organization = ProviderOrganization.find(params[:organization_id])
+  rescue ActiveRecord::RecordNotFound
+    render file: "#{Rails.root}/public/404.html", layout: false, status: :not_found
```
kyeah, line 16: Should this be a catch-all as part of the ApplicationController? I noticed it in other controller(s?) as well.

```ruby
rescue_from ActiveRecord::RecordNotFound, with: :error_not_found

def error_not_found
    render file: "#{Rails.root}/public/404.html", layout: false, status: :not_found
end
```

dpc-portal/app/controllers/credential_delegate_invitations_controller.rb
```diff
@@ -4,6 +4,7 @@
 class CredentialDelegateInvitationsController < ApplicationController
   before_action :authenticate_user!
   before_action :load_organization
+  before_action :ao?
```
kyeah, line 7: I think the upcoming "accept invitation" pages are recommended to be part of this controller under new methods, so might want to pre-emptively set this only for `new` and `create`.

dpc-portal/app/controllers/application_controller.rb
```diff
@@ -1,4 +1,26 @@
...
+  end
+
+  def can_access?
+    redirect_to organizations_path unless current_user.can_access?(@organization)
+  end
+
+  def ao?
+    redirect_to organizations_path unless current_user.ao?(@organization)
```
kyeah, line 24: We haven't discussed this (and not a blocker for this PR), but we should sync with design on whether this should redirect or be a 404 (and if any unknown portal routes will be a redirect or 404.)

kyeah: Also wonder from a method naming perspective if something like `require_ao` or `redirect_if_not_ao` is more descriptive here 👋

dpc-portal/app/controllers/organizations_controller.rb
```diff
@@ -3,21 +3,23 @@
...
-    @organizations = [@organization]
+    @organizations = current_user.provider_organizations
     render(Page::Organization::OrganizationListComponent.new(organizations: @organizations))
   end
 
   def show
-    if params[:ao]
-      provider_organization = ProviderOrganization.find_by(dpc_api_organization_id: @organization.api_id)
-      @invitations = Invitation.where(provider_organization:,
+    if params[:ao] && current_user.ao?(@organization)
```
kyeah, line 17: Do we still need the :ao param here?

dpc-portal/spec/components/page/organization/show_component_spec.rb
```diff
@@ -207,13 +207,13 @@
...
         form1 = <<~HTML
-          <form class="button_to" method="post" action="/portal/organizations/99790463-de1f-4f7f-a529-3e4f59dc7131/ip_addresses/addr-id-1">
+          <form class="button_to" method="post" action="/portal/organizations/2/ip_addresses/addr-id-1">
             <input type="hidden" name="_method" value="delete" autocomplete="off" />
             <button class="usa-button" type="submit">Yes, revoke address</button>
           </form>
         HTML
         form2 = <<~HTML
-          <form class="button_to" method="post" action="/portal/organizations/99790463-de1f-4f7f-a529-3e4f59dc7131/ip_addresses/addr-id-2">
+          <form class="button_to" method="post" action="/portal/organizations/2/ip_addresses/addr-id-2">
```
kyeah, line 216: Small security note thinking ahead to an ATO — there may be some desire here to convert these into non-sequential IDs to do some obfuscation to the end-user.

(Sorry, just creating more decisions for us to document 😄)

kyeah (commented):
LGTM, no major blocking inline comments.

I did note that the TOS Form is still using `placeholder` as the organization ID when linking to `sign_tos`, so it hits an error when going through the Add an Org flow. I assume you wanted to replace that with the organization path ID.

dpc-portal/app/controllers/organizations_controller.rb
jdettmannnava: I think so. Until we get the design for the AO show page, we need a way to toggle what the user sees. We could do it another way, but I thought this would be the easiest in the current implementation.

dpc-portal/spec/components/page/organization/show_component_spec.rb
jdettmannnava: I know this is a security risk, but I think it is mitigated by our strong auth/auth.

dpc-portal/app/controllers/application_controller.rb
jdettmannnava: I think (for comms) we (broadly defined) are supposed to be making a spreadsheet of all failure conditions. Hopefully, this will be addressed there!

dpc-portal/app/controllers/application_controller.rb
jdettmannnava: I just allowed organization_id to be overridden to get this all in one place.

I am hoping 404's will eventually be handled more on a case-by-case basis and this is just a stopgap.

dpc-portal/app/controllers/organizations_controller.rb
kyeah: Ah okay, I figure the additional condition (current_user.ao?(@organization)) already handled this in a way where we would either show the AO page (if they are an AO) or the CD page otherwise.

I _thought_ also that we already got the AO show page designs where they wanted the Tabbed layout but were fine with showing all the views together for now (CD management and token/ip/public key management) but I may be misremembering.

jdettmannnava: We do have designs. I just don't think this ticket is the best place to fix it, because the changes are already pretty extensive.
It would have been better if the original implementer (me) would have just pulled them together, but I didn't.

PR #2077, opened 2024-03-19
[DPC-3961] Split downloadS3File for local development

chris-ronning-ny:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3961

## 🛠 Changes

This resolves an issue introduced [in a previous PR](https://github.com/CMSgov/dpc-app/pull/2055/files?diff=split&w=0#diff-c0543357d02d58ebd5ea8e708f60c113e1172cc9b3aa7b64947a3afb4af7db48L175) that broke `make run-local` at `downloadS3File`. The code here simply bypasses the need to assume a role to access the S3 bucket during local development. 

## ℹ️ Context for reviewers

This unblocks local development for the time being. Integration tests still run from GitHub Actions, but they don't appear to work in a local environment (they didn't before either, so not a regression).

The issue may have been exacerbated by the use of the `IS_TESTING` environment variable, which is used for multiple purposes. At the moment, `make run-local` relies on that env var to be `true`, and while some of our code treats it as only necessary for unit and integration tests, the code in the aforementioned PR treats it as necessary for local development. This should be re-considered moving forward.

## ✅ Acceptance Validation
- `make run-local` executes successfully in a local dev environment.
- GH Actions run successfully.

## 🔒 Security Implications

### DIFF
--- a/lambda/opt-out-import/main.go
+++ b/lambda/opt-out-import/main.go
@@ -210,23 +210,40 @@ func createSession() (*session.Session, error) {
 }
 
 func downloadS3File(bucket string, file string) ([]byte, error) {
-	cfg, err := createV2Cfg()
-	if err != nil {
-		return []byte{}, err
-	}
-
-	downloader := manager.NewDownloader(s3v2.NewFromConfig(*cfg))
-	buff := &aws.WriteAtBuffer{}
-	numBytes, err := downloader.Download(context.TODO(), buff, &s3v2.GetObjectInput{
-		Bucket: aws.String(bucket),
-		Key:    aws.String(file),
-	})
-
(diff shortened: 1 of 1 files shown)


kyeah (approved):
Thanks, LGTM! 

There was an intention for local development to work as long as AWS_ASSUME_ROLE_ARN is set, but it sounds like that is also a little wonky depending on what that value is set to. This should bypass the issue for now until we can refactor things more cleanly 🌟

chris-ronning-ny:
@kyeah for what it's worth, I couldn't figure out a value for `AWS_ASSUME_ROLE_ARN` that works in local dev.

PR #2078, opened 2024-03-19
Me/dpc 3860 hapi exception fix

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3860

## 🛠 Changes

- Updated `HAPIExceptionHandler` so that all exceptions it handles return the correct HTTP status.
- Bumps Jib so that we can use the newest version of Docker without getting the LayerSources error when attempting to build dpc.
- Fixed broken CLI tests that won't run locally, but were running in GHA.

## ℹ️ Context for reviewers

The previous version of this code used an injected `ResourceInfo` object to determine if the end point from which the exception was thrown was a FHIR end point or not.  If it was a FHIR end point the correct http status was returned, and if it wasn't another exception was thrown and a 500 was returned.  The problem is that at least some of the time, and only in our persisted environments, Jersey failed to fill out the `ResourceInfo` object.  That lead to this handler returning 500's when it shouldn't have.

In the new version, we don't care if the end point the exception was thrown from was a FHIR end point or not.  In either case, we return the correct http status to the caller and fill out the response body with an `OperationOutcome`. 

Also, this handler only works on `BaseServerResponseExceptions`, which are defined in HAPI FHIR.  The only way to get a `BaseServerResponseException` outside of a FHIR end point is to have a non-FHIR end point make a FHIR call with the HAPI client and for it to throw an exception.  I'm not sure that's possible in our code, so worrying about whether an exception comes from a FHIR end point or not is probably moot.

## ✅ Acceptance Validation

Code updated and automated and manual tests are passing.

## 🔒 Security Implications

### DIFF
--- a/dpc-api/src/test/java/gov/cms/dpc/api/cli/OrganizationTests.java
+++ b/dpc-api/src/test/java/gov/cms/dpc/api/cli/OrganizationTests.java
@@ -6,6 +6,7 @@ import gov.cms.dpc.api.DPCAPIService;
 import gov.cms.dpc.api.cli.organizations.OrganizationDelete;
 import gov.cms.dpc.api.cli.organizations.OrganizationList;
 import gov.cms.dpc.api.cli.organizations.OrganizationRegistration;
+import gov.cms.dpc.testing.DummyJarLocation;
 import io.dropwizard.core.cli.Cli;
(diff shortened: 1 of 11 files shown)


ashley-weaver (approved):
LGTM

PR #2084, opened 2024-03-21
[DPC-3968] Use correct filename prefix for prod and lower environments

chris-ronning-ny:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3968

## 🛠 Changes

This updates the functions that handle filename generation and parsing for both lambdas. Specifically, in production, we expect to receive and generate files with a `P` prefix; for all lower environments, the prefix should be `T`.

## ℹ️ Context for reviewers

(Background context, more in-depth details of the implementation, and anything else you'd like to call out or ask reviewers.)

## ✅ Acceptance Validation

## 🔒 Security Implications


lambda/opt-out-export/main_test.go
```diff
@@ -201,9 +201,19 @@ func TestGetAwsSession(t *testing.T) {
...
+
+func TestGenerateRequestFileName_Prod(t *testing.T) {
+	testEnv := os.Getenv("ENV")
+	os.Setenv("ENV", "prod")
 
+	now, _ := time.Parse("2006-01-02 15:04:05", "2010-01-01 12:00:00")
 	fileName := generateRequestFileName(now)
 	assert.Equal(t, "P#EFT.ON.DPC.NGD.REQ.D100101.T1200000", fileName)
+
+	os.Setenv("ENV", testEnv)
```
jdettmannnava, line 218: Does this need to be above the assertion in case of fail?

lambda/opt-out-import/parsers_test.go
```diff
@@ -20,15 +20,25 @@ func TestParseMetadata(t *testing.T) {
...
+
+	// prod environment with correct prefix
+	testEnv := os.Getenv("ENV")
+	os.Setenv("ENV", "prod")
+	expTime, _ = time.Parse(time.RFC3339, "2019-01-23T11:22:00Z")
 	metadata, _ = ParseMetadata("blah", "P.NGD.DPC.RSP.D190123.T1122001.IN")
 	assert.Equal(t, "P.NGD.DPC.RSP.D190123.T1122001.IN", metadata.Name)
 	assert.Equal(t, expTime.Format("D060102.T150405"), metadata.Timestamp.Format("D060102.T150405"))
+
+	os.Setenv("ENV", testEnv)
```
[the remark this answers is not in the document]
jdettmannnava: unless there is some sort of after hook or finally available...

lambda/opt-out-import/utils_test.go
```diff
@@ -0,0 +1,29 @@
...
+func TestGenerateConfirmationFileName_Prod(t *testing.T) {
+	testEnv := os.Getenv("ENV")
+	os.Setenv("ENV", "prod")
+	responseFileName := "bfdeft01/dpc/in/P.NGD.DPC.RSP.D240123.T1122001.IN"
+	now, _ := time.Parse("2006-01-02 15:04:05", "2010-01-01 12:00:00")
+	confirmationFileName := GenerateConfirmationFileName(responseFileName, now)
+
+	assert.Equal(t, "bfdeft01/dpc/out/P#EFT.ON.DPC.NGD.CONF.D100101.T1200000", confirmationFileName)
+
+	os.Setenv("ENV", testEnv)
```
jdettmannnava, line 28: Above assert?

lambda/opt-out-export/main_test.go
chris-ronning-ny: Can you explain your rationale? From what I can tell, a failing assertion won't actually stop the execution of the function. So, `os.Setenv` will still be called even if the assertion fails. 

To me, setting this environment variable afterwards makes the intent a lot clearer than setting it before the assertion. Given that it wouldn't change the behavior anyways, I'd like to keep it the same. 

CC @kyeah since Jeff is out today

chris-ronning-ny: I tested this to prove my statement above.

Code:
```
func TestExecution(t *testing.T) {
	println("Before assertion")
	assert.Equal(t, true, false)
	println("After assertion")
}
```

Output:
```
=== RUN   TestExecution
Before assertion
    parsers_test.go:21:
        	Error Trace:	parsers_test.go:21
        	Error:      	Not equal:
        	            	expected: true
        	            	actual  : false
        	Test:       	TestExecution
After assertion
```

kyeah: From a consistency perspective, `defer` calls _are_ the best practice within Go for stating that something needs to happen at the end of a function (regardless of what happens inside of the function.) This ensures that the "thing that changes" and the "thing that reverts the change" are packaged together clearly back-to-back:

```go
testEnv := os.Getenv("ENV")

os.Setenv("ENV", "prod")
defer os.Setenv("ENV", testEnv)
```

Otherwise you may have a lot of code in between the "thing that changes" and the "thing that reverts the change".

In this case it's a pretty short test function and you explained the assertion behavior so I wouldn't care _that much_ if we kept it as is 😄 But using `defer` states the intention clearly, whereas if I had seen this code outside of the PR context I would have assumed that the cleanup could be skipped based on a failed assertion.

kyeah (approved):
Looks good, thank you! 🌟

PR #2103, opened 2024-03-28
DPC-3965 Allow local login to DPC-Web

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3965

## 🛠 Changes

Conditional logic added so only store session in redis if not local

## ℹ️ Context for reviewers

Session was being stored in redis, but development cache is by default in memory. It seemed better to keep the default cache functionality for development than use redis.

## ✅ Acceptance Validation

Logged in locally

## 🔒 Security Implications

### DIFF
--- a/dpc-web/config/environments/development.rb
+++ b/dpc-web/config/environments/development.rb
@@ -12,20 +12,11 @@ Rails.application.configure do
   # Show full error reports.
   config.consider_all_requests_local = true
 
-  # Enable/disable caching. By default caching is disabled.
-  # Run rails dev:cache to toggle caching.
-  if Rails.root.join('tmp', 'caching-dev.txt').exist?
-    config.action_controller.perform_caching = true
+  # don't cache controller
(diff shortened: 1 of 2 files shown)


kyeah (approved):
Would be nice to have the local architecture mirror production (using Redis), but working is better than broken 😄 Strange that it doesn't work with the dockerized Redis instance.

Would be good to test other Redis-dependent functionality like emails and such (not related to this PR — just thinking about verifying the Redis behavior locally.)

jdettmannnava:
> Would be nice to have the local architecture mirror production (using Redis), but working is better than broken 😄 Strange that it doesn't work with the dockerized Redis instance.
> 
> Would be good to test other Redis-dependent functionality like emails and such (not related to this PR — just thinking about verifying the Redis behavior locally.)

Oh, it does work with local redis if we change the caching mechanism in config/environments/development.rb

PR #2109, opened 2024-04-01
DPC-3974 Share AO/CD organization view component

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3974

## 🛠 Changes

- Tabbed navigation component added
- Compound Show component added
- Show Component renamed

## ℹ️ Context for reviewers

Both AOs and CDs can manage credentials, but only AOs can manage credential delegates.
@kyeah suggested that tabbing between views by AOs would be better in JavaScript than making page calls, so that is what we went with.

## ✅ Acceptance Validation

Tests. Screenshots
### As AO
useraccess
Credentials

### As CD
cd-creds

## 🔒 Security Implications


ashley-weaver (approved):
LGTM

dpc-portal/app/components/core/navigation/tabbed_component.html.erb
```diff
@@ -0,0 +1,42 @@
...
+      if (document.querySelector(selectors[current_index])) {
+        document.querySelector(selectors[current_index]).style.display = 'none';
+      }
+    }
+    current_index++;
+  }
+}
+document.addEventListener("DOMContentLoaded", function() {
+    if (document.querySelector("#<%= @header_id %>")) {
+      document.querySelector("#<%= @header_id %>").style.display = 'block';
```
kyeah, line 22: Hmm wondering if we could implement this a bit differently. I noticed that the header is never revealed in Lookbook (probably because the DOMContentLoaded event happens when lookbook is opened, not when the component is rendered.)

CleanShot 2024-04-03 at 17 41 41@2x

Is it possible to always show the header, and render the non-start-index list items with display: none without the startup javascript?

dpc-portal/app/components/core/navigation/tabbed_component.rb
```diff
@@ -0,0 +1,26 @@
...
+    class TabbedComponent < ViewComponent::Base
+      # links should be a list of tuples:
+      # [text, href, selector_for_link, current?]
+      # href in case javascript disabled
+      def initialize(header_id, links)
+        super
+        @header_id = header_id
+        @start_index = -1
+        links.each_with_index do |link, index|
+          @start_index = index if link[3]
```
kyeah, line 19: what happens if there's no start index? or multiple start indexes? 🤔

dpc-portal/app/components/page/organization/compound_show_component.html.erb
```diff
@@ -0,0 +1,7 @@
+<div>
+  <h1><%= @organization.name %></h1>
+  <div class="margin-bottom-5">NPI: <%= @organization.npi %></div>
+  <%= render(Core::Navigation::TabbedComponent.new('test_header', @links)) if @show_cds %>
```
kyeah, line 4: Should we use a realistic header ID here like `organization_nav`?

kyeah (commented):
LGTM overall, thank you! functionality wise it looks good, made a note regarding something I saw in Lookbook.

dpc-portal/app/components/core/navigation/tabbed_component.rb
jdettmannnava: No start index: tabs appear but no content
Multiple "show me"s: last tab is selected
I can added comments to the component!

dpc-portal/app/components/core/navigation/tabbed_component.html.erb
jdettmannnava: Have pushed this back to Clara, as even the raw example from uswds disappears when viewport < 64rem.

jdettmannnava:
Closing because extended header disappears

kyeah (approved):
Works in the UI and in Lookbook now. Nice work, thank you!

PR #2138, opened 2024-04-17
[DPC-3981] Me/dpc 3981 fix smoke tests

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3981

## 🛠 Changes

Updates `make smoke/local` so that when run on Jenkins it purges JMeter's utility library and rebuilds it based on dpc-smoketest's class path.

## ℹ️ Context for reviewers

The original ask for this ticket was to fix failing DPC smoke tests in Jenkins.  They were failing because we had updated DropWizard, and in that process bumped the jackson-annotations library from 1.13.3 to 2.16.1.  This shouldn't have been a problem, but since dpc-smoketest is a JMeter component, when run through Jenkins its dependencies are pulled from JMeter's static utility library where they have to be manually updated (See [here](https://jmeter.apache.org/usermanual/get-started.html#classpath) for more info).  Despite updating our pom file with the new dependency, the JMeter library still contained the old version.  The fix was to manually copy `jackson-annotations.2.16.1.jar` to the JMeter library and delete `jackson-annotations.2.13.3.jar`.

This changes in this PR are to rebuild JMeter's static library on the Jenkins server each time the smoke test is run so we won't have to worry about this in the future.  The process goes like this:

1. Check to see if the JMeter utility library directory exists.
2. Delete all of the *.jar files from the library.
3. Use Maven to copy all of dpc-smoketest's dependency jars into the JMeter library.
4. Run as usual.

## ✅ Acceptance Validation

- Smoke tests are currently passing (See [here](https://management.dpc.cms.gov/job/DPC%20-%20Smoke%20Test/)).
- A version of dpc-smoketest with logging turned up was run on Jenkins to ensure the correct version of the jackson-annotations jar was being loaded.
- When run on a system without a JMeter library, i.e. on your local machine, everything still runs the same.

## 🔒 Security Implications

### DIFF
--- a/dpc-common/src/main/java/gov/cms/dpc/fhir/FHIRExtractors.java
+++ b/dpc-common/src/main/java/gov/cms/dpc/fhir/FHIRExtractors.java
@@ -210,7 +210,7 @@ public class FHIRExtractors {
      * @param identifiers List of {@link Identifier}s to search
      * @param system {@link DPCIdentifierSystem} to search identifiers for
      * @return {@link Identifier}
-     * @throws IllegalArgumentException
+     * @throws IllegalArgumentException thrown if no identifier is found
      */
     public static Identifier findMatchingIdentifier(List<Identifier> identifiers, DPCIdentifierSystem system) throws IllegalArgumentException {
         return identifiers
--- a/dpc-common/src/test/java/gov/cms/dpc/common/models/JobCompletionModelUnitTest.java
+++ b/dpc-common/src/test/java/gov/cms/dpc/common/models/JobCompletionModelUnitTest.java
@@ -1,11 +1,9 @@
 package gov.cms.dpc.common.models;
 
-import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import org.junit.jupiter.api.Test;
 
-import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
@@ -15,16 +13,6 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 class JobCompletionModelUnitTest {
     @Test
     public void testDeserialize() throws IOException {
-        Package pack = JsonFormat.class.getPackage();
-        System.out.println(pack);
-        System.out.println(JsonFormat.Feature.READ_DATE_TIMESTAMPS_AS_NANOSECONDS.toString());
-
-        String classpath = System.getProperty("java.class.path");
-        String[] classPathValues = classpath.split(File.pathSeparator);
-        for (String classPath: classPathValues) {
-            System.out.println(classPath);
-        }
-
         ObjectMapper mapper = new ObjectMapper().registerModule(new JavaTimeModule());
 
         // Load serialized JobCompletionModel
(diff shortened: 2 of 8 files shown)


dpc-common/src/main/java/gov/cms/dpc/fhir/FHIRExtractors.java
```diff
@@ -210,7 +210,7 @@ public static Pair<String, String> parseTag(String tag) {
      * @param identifiers List of {@link Identifier}s to search
      * @param system {@link DPCIdentifierSystem} to search identifiers for
      * @return {@link Identifier}
-     * @throws IllegalArgumentException
+     * @throws IllegalArgumentException thrown if no identifier is found
```
MEspositoE14s, line 213: This wasn't necessary for this PR, but not having a description was logging a warning when we build and I wanted to get rid of it.

dpc-common/src/test/java/gov/cms/dpc/common/models/JobCompletionModelUnitTest.java
```diff
@@ -0,0 +1,26 @@
...
+        ObjectMapper mapper = new ObjectMapper().registerModule(new JavaTimeModule());
+
+        // Load serialized JobCompletionModel
+        InputStream inputStream = getClass().getClassLoader().getResourceAsStream("JobCompletionModel.json");
+        String json = new String(inputStream.readAllBytes(), StandardCharsets.UTF_8);
+
+        // Try to deserialize JobCompletionModel
+        JobCompletionModel jobCompletionModel = mapper.readValue(json, JobCompletionModel.class);
+        assertEquals("http://dev.dpc.cms.gov/api/v1/Group/82771a76-f010-44bd-9f7a-7fb02b130f11/$export_format=json", jobCompletionModel.getRequest());
+    }
```
MEspositoE14s, line 25: This test was added when I thought this was going to be a parsing error.  It verifies that our JobCompletionModel class can be parsed correctly by Jackson.  It turned out the error was something completely different, but I figured I'd leave it in place since it's already written.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 bzt
 wheel>=0.38.0 # not directly required, pinned by Snyk to avoid a vulnerability
-numpy>=1.22.2 # not directly required, pinned by Snyk to avoid a vulnerability
\ No newline at end of file
+numpy>=1.22.2 # not directly required, pinned by Snyk to avoid a vulnerability
+setuptools
```
MEspositoE14s, line 4: This was necessary to get the `pip install` part of `make smoke/local` to work on my machine.

dpc-smoketest/pom.xml
```diff
@@ -132,6 +132,12 @@
             <artifactId>gson</artifactId>
             <version>2.9.0</version>
         </dependency>
+        <!-- Not necessary for dpc-smoketest, but it needs to be copied to JMeter's static lib on Jenkins when we copy the classpath -->
+        <dependency>
+            <groupId>kg.apc</groupId>
+            <artifactId>cmdrunner</artifactId>
+            <version>2.2</version>
```
jdettmannnava, line 139: Should this have `<scope>provided</scope>` or `<scope>test</scope>` to keep it out of the actual running of the thing? It seems to be part of `mvn dependency:copy-dependencies` regardless.

MEspositoE14s: Good idea!  Just added test scope.

jdettmannnava (approved):
LGTM

PR #2150, opened 2024-04-22
[DPC-4038] Me/dpc 4038 fix smoke tests

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4038

## 🛠 Changes

Forces build and installation of dpc-common before attempting to copy dpc-smoketest's dependencies to JMeter.

## ℹ️ Context for reviewers

Forcing the install of dpc-common makes sure that the dpc-common.jar and dpc-testing.jar are available when Maven attempts to build and run the smoke tests on Jenkins.  

This isn't necessary every time smoke tests need to run, only when we change dpc-common or dpc-testing or purge the Maven cache, but this was the easiest way to make sure that they're always available.

## ✅ Acceptance Validation

Ran against dev and made sure smoke tests passed.

## 🔒 Security Implications

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,12 +19,10 @@ smoke:
 	if [ -d $$JENKINS_DIR ]; then \
       	echo "Rebuilding JMeter lib"; \
       	rm $$JENKINS_DIR/*.jar; \
-      	mvn dependency:copy-dependencies -pl dpc-smoketest -DoutputDirectory=$$JENKINS_DIR; \
+      	mvn dependency:copy-dependencies -pl dpc-smoketest -DoutputDirectory=$$JENKINS_DIR -am; \
     else \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -19,6 +19,7 @@ smoke:
 	if [ -d $$JENKINS_DIR ]; then \
       	echo "Rebuilding JMeter lib"; \
```
jdettmannnava, line 20: These are a mix of spaces and tabs, which is kind of making me crazy. Everything else seems to have a single tab

MEspositoE14s: I thought I had IntelliJ configured to handle these automatically, but I guess not 😩.

Anyway, it should be all tabs like the rest of the file now.  We should think about setting up a linter in our GHA to catch these.

PR #2158, opened 2024-04-30
[DPC-4039] set logger to info in relevant namespaces

chris-ronning-ny:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4039

## 🛠 Changes

Explicitly sets the logging level of certain namespaces to INFO, so that broken dashboards work.

## ℹ️ Context for reviewers

A blanket INFO level for all unspecified namespaces potentially could have resulted in a flood of logs from all over our codebase, and a higher AWS bill. This approach ensures we're only turning on logs that we want to pick up on the dashboards.

## ✅ Acceptance Validation

Deployed this branch to `dev` and ran a number of API calls, ensuring that logs were captured as expected, and that they render successfully from the dashboards in question (pointed to dev).

## 🔒 Security Implications

### DIFF
--- a/dpc-aggregation/src/main/resources/application.yml
+++ b/dpc-aggregation/src/main/resources/application.yml
@@ -44,6 +44,7 @@ logging:
   level: ERROR
   loggers:
     "liquibase": INFO
+    "gov.cms.dpc.aggregation.engine": INFO
     "gov.cms.dpc.queue.DistributedBatchQueue": DEBUG
     "org.hibernate": ERROR
     "org.hibernate.SQL": ERROR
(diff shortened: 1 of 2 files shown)


ashley-weaver (approved):
👍

dpc-aggregation/src/main/resources/application.yml
```diff
@@ -44,6 +44,7 @@ logging:
   level: ERROR
```
MEspositoE14s, line 44: It might be worth thinking about setting our default to WARN at some point, but this should get the job done 👍

MEspositoE14s (approved):

PR #2164, opened 2024-05-06
Me/dpc 4005 missing logs

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4005

## 🛠 Changes

- Set loggers for `DPCAuthFilter`, `GeneratedRequestIdFilter` and `LogResponseFilter` to `INFO` so that contextual information is written to our Splunk logs for easier debugging.
- Set default log levels to `WARN` instead of `ERROR`.

## ℹ️ Context for reviewers

We noticed that certain logs from the above classes were no longer getting written out, and it was because they are logged at `INFO` and our log levels were set to `ERROR`.  

We talked about potentially setting the default log level to `INFO` in a previous ticket (#2158), but the thought was that it might lead to flooding our logs.  As a bit of a compromise, I set them to WARN in this ticket, but I'm open to changing them if anyone feels strongly about it.  If we find more missing logs we can always turn logging up to `INFO`, and if the logs get a little too crowded we can turn them back to `ERROR`.

## ✅ Acceptance Validation

Ran locally and ensured the logs are getting produced.

## 🔒 Security Implications

### DIFF
--- a/dpc-aggregation/src/test/resources/test.application.yml
+++ b/dpc-aggregation/src/test/resources/test.application.yml
@@ -42,12 +42,14 @@ server:
             application: ${APPLICATION:-"dpc-aggregation"}
 
 logging:
-  level: ERROR
+  level: WARN
   loggers:
     "liquibase": ERROR
     "gov.cms.dpc.queue.DistributedBatchQueue": DEBUG
     "org.hibernate": ERROR
     "org.hibernate.SQL": ERROR
+    "gov.cms.dpc.api.auth": INFO
+    "gov.cms.dpc.common.logging.filters": INFO
   appenders:
     - type: console
       timeZone: UTC
--- a/dpc-aggregation/src/test/resources/test.application.yml
+++ b/dpc-aggregation/src/test/resources/test.application.yml
@@ -42,7 +42,7 @@ server:
             application: ${APPLICATION:-"dpc-aggregation"}
 
 logging:
-  level: WARN
+  level: DEBUG
   loggers:
     "liquibase": ERROR
     "gov.cms.dpc.queue.DistributedBatchQueue": DEBUG
--- a/dpc-aggregation/src/test/resources/test.application.yml
+++ b/dpc-aggregation/src/test/resources/test.application.yml
@@ -42,7 +42,7 @@ server:
             application: ${APPLICATION:-"dpc-aggregation"}
 
 logging:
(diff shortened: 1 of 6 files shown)


dpc-aggregation/src/test/resources/test.application.yml
```diff
@@ -42,12 +42,14 @@ server:
             application: ${APPLICATION:-"dpc-aggregation"}
 
 logging:
-  level: ERROR
+  level: WARN
```
jdettmannnava, line 45: Do we want more logs on test?

MEspositoE14s: I was back and forth on that, but yeah, it's probably better most of the time.  Realistically, you can always set it however you want it when you're testing locally, anyway.

jdettmannnava: I wouldn't even load in CI....

MEspositoE14s: One benefit of having the extra logs in CI is that when a check fails it's nice to be able just click on the link and see why.  

Regardless, CI runs the test DropWizard config, so AFAIK the only way to exclude the logs from CI is to completely turn them off in the test config file.  I'm not sure if it would be better to have the default be more logs for both or no logs for either.  I don't feel super strongly about it either way, though, because you can always set it to whatever you want as needed.

chris-ronning-ny (approved):
Looks good to me, maybe give a quick think about storage constraints wrt above comment? I don't really foresee it being a problem.

dpc-attribution/src/main/resources/application.yml
```diff
@@ -32,7 +32,7 @@ server:
             application: ${APPLICATION:-"dpc-attribution"}
 
 logging:
-  level: ERROR
+  level: WARN
```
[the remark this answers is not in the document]
MEspositoE14s: As a general rule, I think WARN should only be used for things that we should be alerted to, so I like having it turned on.  This comes with the big caveat, though, that we're diligent in our code to only use WARN and ERROR for things that truly need our attention.  If it turns out that's not the case, or if our logs start getting too big, I'm totally open to going back to ERROR.  In my mind, this is less of a black and white issue and more of a shades of gray, how does everyone feel about it sort of thing.

PR #2170, opened 2024-05-13
docs: Update portal README with manual workflow instructions

kyeah:
## 🎫 Ticket

N/A

## 🛠 Changes

- Add scripts to generate GOLDEN_MACAROON as part of `make portal-console` and `make portal-sh`
- Update README with instructions on how to go through Org setup and AO invitation flow
- Reorganize README a bit to consolidate test instructions and such

## ℹ️ Context for reviewers

Wanted to go through the current workflow and had to trawl through the code a bit to figure out AO invitations as it is currently built. Added documentation while getting it to work.

## ✅ Acceptance Validation

Walked through the steps as written in order to get AO invitation working for myself.

## 🔒 Security Implications

### DIFF
--- a/dpc-portal/README.md
+++ b/dpc-portal/README.md
@@ -8,8 +8,8 @@ This is the Data Point of Care (DPC) Portal website. It allows users to manage a
 
 The DPC website can be run locally via docker. Follow the below steps to build and run the website into a Docker container.
 
--   Run `make portal` to build the docker image, and `make start-portal` to start dpc-portal in a docker container.
 -   The DPC API is required for much of the Portal's functionality. To run the DPC API, run `make start-api`.
+-   Run `make portal` to build the docker image, and `make start-portal` to start dpc-portal in a docker container.
 -   Run `make down-portals` to stop dpc-portal.
 
 The following commands can be useful for manual interaction:
@@ -52,6 +52,8 @@ One of the following fake SSNs are required on your Login.gov sandbox account in
 -   900888888
 -   666222222
 
+If you need to use another SSN, you can add it to the ao_ssns variable in the [fake enrollment roles endpoint](/dpc-portal/spec/support/fake_cpi_gateway.rb).
+
 Most generated NPIs will return a successful AO/organization verification when run against the [fake CPI API Gateway](/dpc-portal/spec/support/fake_cpi_gateway.rb). More specific scenarios are supported below:
 
 | Fake Organization NPI | Scenario                                       |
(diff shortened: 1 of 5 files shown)


dpc-portal/bin/console
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env sh
```
jdettmannnava, line 1: I don't think this should be necessary, as this happens when the docker container is initialized, so these variables should already be set?

dpc-portal/README.md
```diff
@@ -1,54 +1,97 @@
...
+    ```
 
-## View Components
+## Manual Workflows
 
-We utilize the [ViewComponent](https://viewcomponent.org/) library to create custom components. These components are documented and viewable within [Lookbook](http://localhost:3100/portal/lookbook).
+To support manual workflows, the system offers rake commands that perform administrative commands for new accounts. These rake commands can be run within a portal shell (`make portal-sh`).
 
-Emails are not viewable in lookbook, but rather [here](http://localhost:3100/portal/rails/mailers/).
+-   `rails dpc:make_org` - Creates a new provider organization.
```
jdettmannnava, line 36: I actually don't know if this is useful any more.

dpc-portal/bin/console
kyeah: It's only run as part of the entrypoint script / as part of running the rails or sidekiq apps. When you run a separate interactive shell or rails console, it doesn't have the env var set there.

dpc-portal/README.md
kyeah: oops! correct — didn't look closely. the provider_organization record is created as part of invite_ao, maybe we can just remove make_org then.

dpc-portal/bin/console
jdettmannnava: Ah, since I always run against a web container, I was confused

jdettmannnava: `docker exec -it dpc-app-dpc_portal-1 rails c`
which is actually in my .zshrc

jdettmannnava: Same with shell

jdettmannnava: So in order to adjust three files to any changes in initialization, maybe recommend running against a web container?

kyeah: Hmm our commands are functionally the same (the docker-compose command in the Makefile just uses the container name defined in the docker-compose.portals.yml file.)

```
$ docker exec -it dpc-app-dpc_portal-1 rails c
irb(main):002> ENV['GOLDEN_MACAROON']
=> ""
```

I also confirmed by clicking through the UI that GOLDEN_MACAROON is set within the rails app and can interact with the DPC API.

Do you have GOLDEN_MACAROON set in your shell / zshrc? 🤔

dpc-portal/README.md
```diff
@@ -1,54 +1,88 @@
...
 
-# Running via Docker
+# Local Development
+
+## Running via Docker
 
 The DPC website can be run locally via docker. Follow the below steps to build and run the website into a Docker container.
 
 -   Run `make portal` to build the docker image, and `make start-portal` to start dpc-portal in a docker container.
+-   The DPC API is required for much of the Portal's functionality. To run the DPC API, run `make start-api`.
```
jdettmannnava, line 12: Doesn't this need to go first to make the golden macaroon?

dpc-portal/README.md
```diff
@@ -1,54 +1,88 @@
...
+    $ rails dpc:invite_ao INVITE=Bob,Hoskins,bob@example.com,7838426501
 
-Read more about the assets pipeline [here](/docs/portal/assets-pipeline.md).
+    Invitation created for Bob Hoskins for Organization 7838426501
+    http://localhost:3100/portal/organizations/2/invitations/4/accept
+    ```
+
+## Local AO Verification Scenarios
+
+One of the following fake SSNs are required on your Login.gov sandbox account in order to successfully pass AO-Organization verification:
```
jdettmannnava, line 48: Maybe add:
'If you need to use another ssn, you can add it to the [ao_ssns variable ](https://github.com/CMSgov/dpc-app/blob/65c5e7b4adab82ae75aa4005dbb6a4ebb199a86d/dpc-portal/spec/support/fake_cpi_gateway.rb#L76) in the enrollment roles endpoint'

dpc-portal/README.md
```diff
@@ -1,54 +1,88 @@
...
+You can check the status of any jobs by going to the [Sidekiq dashboard](http://localhost:3100/portal/sidekiq).
 
-Then run `make start-portal`. You can confirm that this worked by opening the Rails console and checking `ENV.fetch('GOLDEN_MACAROON')`.
+## Accessing the CPI API Gateway Locally
 
-You can check the status of any jobs by going to the Sidekiq dashboard at `http://localhost:3100/portal/sidekiq`
+By default, the local portal is connected to a mock CPI API Gateway server with preset responses. To connect with the VAL CPI API Gateway environment, you'll need to be running Zscaler. To connect the local portal container over Zscaler, you'll want to add the following directly before line 5 of the Dockerfile and rerun the build steps:
+
+`COPY Zscaler-Root-CA.pem /usr/local/share/ca-certificates/ZScaler-Root-CA.pem`
+
```
jdettmannnava, line 85: Maybe also add:
You will also need to set the following environment variables in docker-compose.portals.yml
CPI_API_GW_BASE_URL=https://val.cpiapi.cms.gov/
CMS_IDM_OAUTH_URL=https://impl.idp.idm.cms.gov/

jdettmannnava (approved):
LGTM!

PR #2173, opened 2024-05-16
DPC-3848 Weekly Verification: Chained

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3848

## 🛠 Changes

- Verify AO Job queues another version of itself if it processed any links, queues Verify ProviderOrganization Job otherwise
- Verify ProviderOrganization Job queues another version of itself if it processed any orgs

## ℹ️ Context for reviewers

Completes the run instead of just a few.
Note: tailing the sidekiq log, locally it takes four seconds to start another job, so the wait fussiness is probably unnecessary.
See [mural](https://app.mural.co/t/nava4113/m/nava4113/1715885224810/4f338609a953b834ac00886af970b22a17765cf8?sender=ud90669672db0f13ad83c0120) for flowcharts highlighting differences between the algorithms.

## ✅ Acceptance Validation

Unit tests and [performance test](https://github.com/CMSgov/dpc-app/tree/jd/dpc-3848-performance-chain)

## 🔒 Security Implications


dpc-portal/app/jobs/verify_provider_organization_job.rb
```diff
@@ -0,0 +1,35 @@
...
+  def perform
+    service = AoVerificationService.new
+    @start = Time.now
+    unless orgs_to_check.each do |org|
+             service.check_org_med_sanctions(org.npi)
+             service.get_approved_enrollments(org.npi)
+             org.update!(last_checked_at: Time.now)
+    rescue AoException => e
+      handle_error(org, e.message)
+           end.empty?
```
kyeah, line 20: This chained statement is confusing to me 🤔 could we break the logic down in a more human-readable way? I read this several times and i am still not 100% sure what `unless orgs_to_check.each do ... end.empty?` means exactly (I didn't think `[list].each` returned anything?)

dpc-portal/app/concerns/verification.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def update_org_sanctions(org, message)
+      org.update!(entity_error_attributes(message))
+      org.ao_org_links.where(verification_status: true).each do |link|
+        link.update!(link_error_attributes(message))
+      end
+    end
+
+    def enqueue_job(klass)
+      wait = (Time.now - @start).seconds
```
kyeah, line 33: Is there a specific intention behind how this wait is calculated?

dpc-portal/app/jobs/verify_ao_job.rb
```diff
@@ -0,0 +1,72 @@
...
+class VerifyAoJob < ApplicationJob
+  queue_as :portal
+  include Verification
+
+  def perform
+    @start = Time.now
+    service = AoVerificationService.new
+    if links_to_check.each do |link|
+         check_link(service, link)
+         update_success(link)
```
kyeah, line 16: small nit but good reminder for myself about automated formatting checks 😄 Looks like this tabbing/spacing is bigger than the rest of the file

dpc-portal/app/jobs/verify_ao_job.rb
```diff
@@ -0,0 +1,72 @@
...
+    rescue AoException => e
+      handle_error(link, e.message)
+       end.empty?
+      enqueue_job(VerifyProviderOrganizationJob)
```
kyeah, line 20: I have some small concerns about relying on chained jobs — wonder about unexpected errors/failures that would prevent enqueuing and delay future verification checks. But as you mentioned when we chatted — easier to remove this portion than to add it in retroactively.

dpc-portal/app/jobs/verify_ao_job.rb
```diff
@@ -0,0 +1,72 @@
...
+    AoOrgLink.transaction do
+      link.update!(last_checked_at: Time.now)
+      link.provider_organization.update!(last_checked_at: Time.now)
+      link.user.update!(last_checked_at: Time.now)
+    end
+  end
+
+  def handle_error(link, message)
+    AoOrgLink.transaction do
+      link.update!(link_error_attributes(message))
```
kyeah, line 41: It would be good for us to eventually document the business logic here, e.g. why are we updating certain tables based on certain scenarios.

At this point, the one question I have is whether we need/want to set the ao_org_link verification status to `false` in every scenario (e.g. if the org has sanctions.) Since it's just a boolean, I can see it being a little confusing (why is the link status false? Is it because the AO is no longer an AO, or because the org has sanctions? etc.) 

From a UX perspective it probably doesn't matter though — once we see an org has a verification issue, we know what the screens should look like.

docker-compose.portals.yml
```diff
@@ -169,6 +170,9 @@ services:
       - GOLDEN_MACAROON=${GOLDEN_MACAROON}
       - API_ADMIN_URL=http://api:9900
       - API_METADATA_URL=http://api:3002/v1
+      - CPI_API_GW_BASE_URL=http://localhost:4567/
+      - CMS_IDM_OAUTH_URL=http://localhost:4567/
+      - RUBY_YJIT_ENABLE=1
```
kyeah, line 175: do we need to add env vars to the sidekiq terraform as well on the dpc-ops side?

kyeah (commented):
no major comments, thank you!

dpc-portal/app/concerns/verification.rb
jdettmannnava: Assuming configured to do around 25 calls, so would not want to start until a second passed. When I was testing locally, it took about 4 seconds between job queued and job actually started, so it is probably not a useful thing to do.

dpc-portal/app/jobs/verify_provider_organization_job.rb
jdettmannnava: `[].each` returns the list. Python has a nice for... else... flow, and this was the ruby equivalent I found on stack overflow. Definitely should be refactored if it is not clear.

dpc-portal/app/jobs/verify_ao_job.rb
jdettmannnava: Nope! This is just weirdness from if/each/else format. The functions are two spaces in from links_to_check.each. Another good reason to refactor.

dpc-portal/app/jobs/verify_ao_job.rb
jdettmannnava: These results are still in flux; there is an open question around AO under med sanctions means API off, for example.

dpc-portal/app/jobs/verify_ao_job.rb
jdettmannnava: I think we would still run the jobs fairly often, because why not?

docker-compose.portals.yml
jdettmannnava: I will add in the ops repo

kyeah (approved):
LGTM 👍 As discussed offline, can take or leave the `wait` functionality for now.

PR #2198, opened 2024-06-19
DPC-4126: Fix JSON logging and add standardized fields

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4126

## 🛠 Changes

- Hide Foreman service log prefixes in nonprod (deployed environments)
- Add request-global log attributes for current_user info, organization info, and any raised exceptions
- Fix request_id attribute

## ℹ️ Context for reviewers

Noticed a few issues while auditing the logs produced by the portal in splunk:

- Since we are using foreman to run the fake CPI API GW in nonprod, it is prefixing the service name. This breaks the JSON formatting of each log line, which breaks splunk parsing. Added sed commands to fix this.
- The request_id is not logged because it uses the Rails default `log_tags` functionality. Since we use lograge for request logs, this is not picked up in the logs.
- The normal logs using `Rails.logger` are not printed via JSON.

I also added some helpful information to each request related to the current logged in user, the relevant organization for the API/page request, and any exception details if there was an uncaught exception raised during the request.

## ✅ Acceptance Validation

Verified locally.

Lograge custom log with current_user and organization:

```json
{
  "method": "GET",
  "path": "/portal/organizations/1/invitations/1/accept",
  "format": "html",
  "controller": "InvitationsController",
  "action": "accept",
  "status": 200,
  "allocations": 11634,
  "duration": 72.23,
  "view": 25.18,
  "db": 1.14,
  "ddsource": "ruby",
  "environment": "local",
  "level": "info",
  "request_id": "b225bb9b-72bf-400e-9b28-733fbc9e16e1",
  "request_user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:125.0) Gecko/20100101 Firefox/125.0",
  "request_ip": "REDACTED",
  "current_user": {
    "id": 1,
    "external_id": "60119c5b-606f-4526-b33c-4e691e2d4d3b",
    "pac_id": null
  },
  "organization": {
    "id": 1,
    "dpc_api_organization_id": null,
    "is_authorized_official": false,
    "is_credential_delegate": false
  }
}
```

Custom Rails.logger log:

```json
{
  "level": "INFO",
  "time": "2024-06-19T18:56:02.963+00:00",
  "environment": "local",
  "message": "This is a test",
  "request_id": "16b9c4c6-e84a-4bd0-805f-9a34d966f475",
[...]

PR #2200, opened 2024-06-21
DPC-3823 Credential Audit Log

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-3823

## 🛠 Changes

- Added CredentialAuditLog model with table
- Added method to application_controller for logging changes to credentials
- Call above method in each credential controller
- Log when client tokens removed from provider organization

## ℹ️ Context

We want to have a log of all credentials managed via the portal for auditing, debugging, and supporting future enhancements.

## 🧪 Validation

Automated and manual testing.

### DIFF
--- a/dpc-portal/app/controllers/application_controller.rb
+++ b/dpc-portal/app/controllers/application_controller.rb
@@ -81,4 +81,15 @@ class ApplicationController < ActionController::Base
     has_ao_link = current_user.ao_org_links.where(provider_organization: @organization).exists?
     has_ao_link ? 'verification' : 'cd_access'
   end
+
+  def log_credential_action(credential_type, dpc_api_credential_id, action)
+    log = CredentialAuditLog.create(user: current_user,
+                                    credential_type:,
+                                    dpc_api_credential_id:,
+                                    action:)
+    return if log.valid?
+
+    logger.error("Unable to create client token CredentialAuditLog #{action} on #{token_id}")
+  end
+
 end
--- a/dpc-portal/app/controllers/application_controller.rb
+++ b/dpc-portal/app/controllers/application_controller.rb
@@ -82,13 +82,13 @@ class ApplicationController < ActionController::Base
     has_ao_link ? 'verification' : 'cd_access'
   end
 
-  def log_credential_action(credential_type, dpc_api_credential_id, action)
-    log = CredentialAuditLog.create(user: current_user,
-                                    credential_type:,
-                                    dpc_api_credential_id:,
-                                    action:)
-    return if log.valid?
+  def log_credential_action(credential_type, action)
+    log = CredentialAuditLog.new(user: current_user,
+                                 credential_type:,
+                                 provider_organization: @organization,
+                                 action:)
+    return if log.save
 
-    logger.error("CredentialAuditLog failure: unable to #{action} #{credential_type} #{dpc_api_credential_id}")
+    logger.error("CredentialAuditLog failure: unable to #{action} #{credential_type} for #{@organization.id}")
   end
 end
--- a/dpc-portal/app/controllers/application_controller.rb
+++ b/dpc-portal/app/controllers/application_controller.rb
@@ -89,13 +89,13 @@ class ApplicationController < ActionController::Base
     CurrentAttributes.save_user_attributes(current_user)
   end
 
-  def log_credential_action(credential_type, action)
+  def log_credential_action(credential_type, dpc_api_credential_id, action)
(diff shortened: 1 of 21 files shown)


dpc-portal/db/migrate/20240620201824_create_credential_audit_logs.rb
```diff
@@ -0,0 +1,11 @@
+class CreateCredentialAuditLogs < ActiveRecord::Migration[7.1]
+  def change
+    create_table :credential_audit_logs do |t|
+      t.bigint :user_id, null: false
+      t.integer :credential_type, null: false
+      t.string :dpc_api_credential_id, null: false
+      t.integer :action, null: false
+      t.timestamps
```
kyeah, line 8: Seems like it would be beneficial to include indexes for searching by:

- user_id
- dpc_api_credential_id

dpc-portal/app/controllers/application_controller.rb
```diff
@@ -81,4 +81,14 @@ def code_prefix
...
   end
+
+  def log_credential_action(credential_type, dpc_api_credential_id, action)
+    log = CredentialAuditLog.create(user: current_user,
+                                    credential_type:,
+                                    dpc_api_credential_id:,
+                                    action:)
+    return if log.valid?
+
+    logger.error("CredentialAuditLog failure: unable to #{action} #{credential_type} #{dpc_api_credential_id}")
```
kyeah, line 92: not relevant, but this is one area where https://github.com/CMSgov/dpc-app/pull/2198 would come in handy (to record these attributes as JSON key/values instead of just part of the stringy message)

kyeah (approved):
Looks good, thank you! The fault-tolerant part of my brain is worried about the action  and the audit log not being atomic (so it's possible a credential is created/deleted without an associated audit log), but the audit log is more of a "best-effort" anyways since people can create/delete these resources directly through the API.

dpc-portal/app/controllers/application_controller.rb
jdettmannnava: Sounds like a good idea. I will wait to merge until 2198.

dpc-portal/app/controllers/application_controller.rb
```diff
@@ -88,4 +88,14 @@ def set_current_request_attributes
     CurrentAttributes.save_request_attributes(request)
     CurrentAttributes.save_user_attributes(current_user)
   end
+
+  def log_credential_action(credential_type, action)
+    log = CredentialAuditLog.new(user: current_user,
+                                 credential_type:,
+                                 provider_organization: @organization,
```
kyeah, line 95: I may have missed (or don't remember 🙈) a discussion — was there a specific reason for moving to tracking based on the provider_organization instead of on the specific credential? In particular, doing it like this prevents us from potentially updating the UI in the future to:

- list the author of a specific credential
- knowing which credentials were created directly in the API vs. in the portal
- which credential this action was for specifically

dpc-portal/app/models/credential_audit_log.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# Record of an action on a credential type
+class CredentialAuditLog < ApplicationRecord
+  validates_presence_of %i[credential_type action]
+  enum credential_type: %i[client_token ip_address public_key]
+  enum action: %i[add remove]
+
+  belongs_to :user, optional: true
```
kyeah, line 9: Let's include a comment here about why it's optional here (had a question about this until I read the PR further 😄)

dpc-portal/app/models/provider_organization.rb
```diff
@@ -66,6 +67,11 @@ def disable_rejected
     ctm = ClientTokenManager.new(dpc_api_organization_id)
     ctm.client_tokens.each do |token|
       ctm.delete_client_token(token.with_indifferent_access)
+      next if CredentialAuditLog.create(credential_type: :client_token,
+                                        provider_organization: self,
+                                        action: :remove)
+
+      logger.error("CredentialAuditLog failure: unable to remove client token for #{id}")
```
kyeah, line 74: Not related to this PR, but we need to ensure this action happens in the future right? Seems like we need a separate ticket to move the disable_rejected logic into a job that can be retried in the future if an error occurs.

jdettmannnava: I think a ticket to monitor failures first, then make more robust?

kyeah (approved):
looks good, ty!

PR #2204, opened 2024-06-27
Me/dpc 4117 emit queue metric registry

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4117

## 🛠 Changes

Created a new `AwsDistributedBatchQueue` that extends the existing `DistributedBatchQueue` to emit a CloudWatch metric that reports the number of job batches in the queue.  

It is configured in the `application.yml` file of `dpc-aggregation`.  If emitting metrics is turned off in the config they'll be written to the logs instead.

Requires https://github.com/CMSgov/dpc-ops/pull/714 to be merged first.

## ℹ️ Context

We are in the process of updating `dpc-aggregation` to scale based on the number of job batches in its queue, and this is the first step.

## 🧪 Validation

- Made sure all tests are passing locally.
- Pushed to dev and made sure metrics were posted.
image


dpc-api/src/main/java/gov/cms/dpc/api/DPCAPIConfiguration.java
```diff
@@ -189,6 +191,18 @@ public void setLookBackExemptOrgs(List<String> lookBackExemptOrgs) {
         this.lookBackExemptOrgs = lookBackExemptOrgs;
     }
 
+    // Never used in dpc-api, but required for the queue service
+    @Override
+    public int getPollingFrequency() {
+        return 500;
```
jdettmannnava, line 197: Not raise a "why you calling me?" error?

dpc-queue/src/main/java/gov/cms/dpc/queue/JobQueueModule.java
```diff
@@ -34,10 +41,17 @@ public void configure() {
             binder.bind(IJobQueue.class)
                     .to(MemoryBatchQueue.class)
                     .in(Scopes.SINGLETON);
-        } else {
+
+        } else if (configuration().getDpcAwsQueueConfiguration() == null) {
```
jdettmannnava, line 45: I prefer the 'if it isn't null use it' structure, but that is just a personal preference

dpc-queue/src/main/java/gov/cms/dpc/queue/JobQueueModule.java
```diff
@@ -56,4 +70,48 @@ int provideBatchSize() {
...
+
+    @Provides
+    CloudWatchAsyncClient provideCloudWatchAsyncClient() {
+        return CloudWatchAsyncClient
+            .builder()
+            .region(Region.of(configuration().getDpcAwsQueueConfiguration().getAwsRegion()))
+            .build();
+    }
+
+    @Provides
```
jdettmannnava, line 85: Provides and inject?

dpc-queue/src/main/java/gov/cms/dpc/queue/JobQueueModule.java
```diff
@@ -1,14 +1,21 @@
 package gov.cms.dpc.queue;
 
+import com.codahale.metrics.*;
```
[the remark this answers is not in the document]
MEspositoE14s: I let IntelliJ add that for me, but I probably should've been more specific and added each class individually just as a matter of good practice.  Will change!

dpc-queue/src/main/java/gov/cms/dpc/queue/JobQueueModule.java
MEspositoE14s: This one's tough because we're not really using it either way, but yeah, that probably makes for an easier read.

dpc-queue/src/main/java/gov/cms/dpc/queue/JobQueueModule.java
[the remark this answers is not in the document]
jdettmannnava: Oh, I just meant I think the jar comes from the pom in dpc-macaroon, and I don't know if that is appropriate

dpc-queue/src/main/java/gov/cms/dpc/queue/JobQueueModule.java
MEspositoE14s: These are one of the ways you can tell Guice how to inject dependencies.  Using this as an example:
```java
@Provides
@Inject
ScheduledReporter provideScheduledReporter(MetricRegistry metricRegistry, CloudWatchAsyncClient cloudWatchAsyncClient)
``` 

`@Inject` tells Guice that when I call this method I'm not going to provide you with the parameters.  It's up to you to go find a `MetricRegistry` and a `CloudWatchAsyncClient ` on your own and inject them into the method call.

`@Provides` tells Guice that when something out there is asking for a `ScheduledReporter `, it can call this method to create one and inject it.

As an example, the way the whole process works goes like this:

1. When it initializes, `dpc-aggregation` will add the `JobQueueModule` to its Guice bundle and call the module's configure() method to bind `IJobQueue` to the `AwsDistributedBatchQueue` class.
2. At some point, `dpc-aggregation` will ask for an `IJobQueue` and Guice will try to build an `AwsDistributedBatchQueue` and pass it along.
3. For each parameter in the constructor for `AwsDistributedBatchQueue`, Guice will look for "providers" that can create them.
4. One of the parameters is a `ScheduledProvider`, and Guice will use the method marked with the `@Provides` annotation above to create it.  

The same thing happens with `dpc-api`, except when `JobQueueModule` is bound there it will use the `DistributedBatchQueue` class.

dpc-api/src/main/java/gov/cms/dpc/api/DPCAPIConfiguration.java
MEspositoE14s: Yeah, that's probably a good idea.

jdettmannnava (approved):
LGTM

PR #2205, opened 2024-06-28
DPC-4132 Only create user on connecting to org

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4132

## 🛠 Changes

- login controller no longer creates user
- invitations controller requires login dot gov tokens instead of sign in
- invitations controller upserts user on confirm registration

## ℹ️ Context

We don't want the users table full of people unrelated to DPC, so we are only creating users after a successful invitation acceptance.

## 🧪 Validation

Automated and manual testing.


MEspositoE14s (approved):
Seems good to me!

PR #2209, opened 2024-07-10
DPC-4078 automate sandbox approval

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4078

## 🛠 Changes

- GrantAccessJob added
- User performs GrantAccessJob when confirmed_at set to non-nil
- Other specs changed to use ActiveJob::TestHelper because queue polluted by GrantAccessJob

## ℹ️ Context

We are not really vetting access to the sandbox, so this job will grant access automatically. We will need to set a Splunk alert on log failure so we can manually add in case of failure.

## 🧪 Validation

Manual and automated testing

### DIFF
--- a/dpc-web/app/jobs/grant_access_job.rb
+++ b/dpc-web/app/jobs/grant_access_job.rb
@@ -2,14 +2,29 @@
 
 class GrantAccessJob < ApplicationJob
   def perform(user_id)
-    # Get user
-    # Check if registered
-    # Check if organization exists (by name)
-    # Make org
-    # Make registered org
     user = User.find user_id
 
-    organization = Organization.find_or_create_by(name: user.requested_organization) do |org|
+    organization = find_or_create_org(user)
+
+    unless organization.registered_organization.present?
+      registered_organization = organization.build_registered_organization
+      registered_organization.build_default_fhir_endpoint
+      registered_organization.save!
+    end
+
+    # Add the user after creating the RegisteredOrganization to make sure email is sent
+    organization.users << user
+    logger.info 'GrantAccessJob success'
+    organization
+  rescue ActiveRecord::RecordNotFound => e
+    logger.error "GrantAccessJob failure: #{e.message}"
+  rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotSaved => e
+    record = e.record
+    logger.error "GrantAccessJob failure creating #{record.class.name}: #{record.errors.full_messages.join(' | ')}"
+  end
+
+  def find_or_create_org(user)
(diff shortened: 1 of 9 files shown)


dpc-web/app/jobs/grant_access_job.rb
```diff
@@ -0,0 +1,38 @@
...
+    organization
+  rescue ActiveRecord::RecordNotFound => e
+    logger.error "GrantAccessJob failure: #{e.message}"
+  rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotSaved => e
+    record = e.record
+    logger.error "GrantAccessJob failure creating #{record.class.name}: #{record.errors.full_messages.join(' | ')}"
+  end
+
+  def find_or_create_org(user)
+    Organization.find_or_create_by!(name: user.requested_organization) do |org|
```
MEspositoE14s, line 27: I hate to go down this rabbit hole because there's no way we can ever be 100% perfect, but can we make this case insensitive, or fuzzy in any other way?  If someone blatantly fat fingers the org name we'll never catch it, but we might fix the occasional typo here or there.

MEspositoE14s (approved):

dpc-web/app/jobs/grant_access_job.rb
jdettmannnava: @kyeah - thoughts? I don't think it is necessary, but I am open to discussion!

kyeah: I think we have several organizations currently with the same name but different casing. Since we're working with existing data, I would recommend we either:

1. keep it case-sensitive here instead of essentially picking one at random 😄 
2. Add slightly more involved logic which makes it case-insensitive, but _only_ if there's exactly one organization that's found. 

In the current manual process, I do add people in a case-insensitive manner so I think it would be good to mirror that to a reasonable extent via Option 2 if possible.

jdettmannnava: FYI Rails is super efficient. It does a count(*) for the case insensitive match (with a limit of 2), and only actually fetches the record itself if there is exactly one. I tried a more complicated solution in which I looked for an exact match within the case-insensitive matches, but that was actually less efficient.

jdettmannnava: @kyeah and @MEspositoE14s : I am using a postgres-specific operator here...

MEspositoE14s: Not ashamed to admit I had to google what that was 😆

MEspositoE14s (approved):

PR #2212, opened 2024-07-16
Me/dpc 4143 filter secrets from logs

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4143

## 🛠 Changes

Added a secrets filter to DropWizard logs in dpc-api, dpc-attribution, dpc-aggregation and dpc-consent.
Secrets are configured in each module's application.yml file.

## ℹ️ Context

We had an issue where a secret was accidentally leaked, this should prevent that going forward.

## 🧪 Validation

- Ran locally and made sure secrets were filtered from the logs.  
- Added unit tests to verify the filter works as expected.


dpc-common/src/main/java/gov/cms/dpc/common/logging/filters/SecretLoggingFilter.java
```diff
@@ -0,0 +1,56 @@
...
+				i.remove();
+			}
+		}
+
+		return new Filter<>() {
+			@Override
+			public FilterReply decide(ILoggingEvent event) {
+				for (String secret : secrets) {
+					if (event.getFormattedMessage().contains(envVars.get(secret))) {
+						logger.warn("Suppressing log, attempted to write " + secret + " in " + event.getLoggerName());
```
jdettmannnava, line 48: This took me a second -- it made me think you were about to right QQT^IQ1qjkaasf; in the log!

MEspositoE14s: I double checked this three different times to make sure I didn't accidentally do that!  

For the sake of slightly better clarity and not confusing the next person, renamed `secret` -> `secretName`.

dpc-common/pom.xml
```diff
@@ -147,6 +147,10 @@
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-params</artifactId>
         </dependency>
+        <dependency>
+            <groupId>uk.org.webcompere</groupId>
```
jdettmannnava, line 151: How did you choose this?

MEspositoE14s: I was looking for a library that would make it easier to temporarily set env variables for unit tests and it did the job.  It's only brought in for tests, from the main .pom...
```
<dependency>
  <groupId>uk.org.webcompere</groupId>
  <artifactId>system-stubs-jupiter</artifactId>
  <version>2.1.6</version>
  <scope>test</scope>
</dependency>
```

MEspositoE14s: It seemed cleaner than mocking the system call to `System.getenv()` inside filter class.

jdettmannnava (approved):
LGTM

PR #2214, opened 2024-07-18
[DPC-4171] Upgrade HAPI FHIR

ashley-weaver:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4171

## 🛠 Changes

The upgrade to 6.x only required an update to one test, so this PR covers two major version upgrades.

### DIFF
--- a/dpc-common/pom.xml
+++ b/dpc-common/pom.xml
@@ -48,10 +48,6 @@
             <artifactId>hapi-fhir-structures-dstu3</artifactId>
         </dependency>
         <dependency>
-            <groupId>${hapi.fhir.groupID}</groupId>
-            <artifactId>hapi-fhir-validation</artifactId>
-        </dependency>
-        <dependency>
             <groupId>javax.ws.rs</groupId>
             <artifactId>javax.ws.rs-api</artifactId>
         </dependency>
--- a/dpc-common/src/test/java/gov/cms/dpc/fhir/validations/AttestationValidationTest.java
+++ b/dpc-common/src/test/java/gov/cms/dpc/fhir/validations/AttestationValidationTest.java
@@ -3,6 +3,7 @@ package gov.cms.dpc.fhir.validations;
 import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.context.support.DefaultProfileValidationSupport;
 import ca.uhn.fhir.validation.FhirValidator;
+import ca.uhn.fhir.validation.SingleValidationMessage;
 import ca.uhn.fhir.validation.ValidationResult;
 import gov.cms.dpc.fhir.validations.profiles.AttestationProfile;
 import gov.cms.dpc.testing.BufferedLoggerHandler;
@@ -52,8 +53,21 @@ class AttestationValidationTest {
     void definitionIsValid() {
         final StructureDefinition provenanceDefinition = dpcModule.fetchStructureDefinition(AttestationProfile.PROFILE_URI);
         final ValidationResult result = fhirValidator.validateWithResult(provenanceDefinition);
-        assertAll(() -> assertEquals(1, result.getMessages().size(), "Should have a single message"),
-                () -> assertTrue(result.getMessages().get(0).getMessage().contains("Provenance.target: derived min (0) cannot be less than base min (1)")));
+
+        StringBuilder resultMessages = new StringBuilder();
+        for (SingleValidationMessage msg: result.getMessages()) {
+            resultMessages.append(msg.getMessage()).append(" ");
+        }
+
+        assertAll(
+                () -> assertEquals(3, result.getMessages().size(), "Should have exactly 3 messages"),
+                () -> assertTrue(resultMessages.toString().contains(
+                        "Element Provenance.target: derived min (0) cannot be less than the base min (1)")),
+                () -> assertTrue(resultMessages.toString().contains(
+                        "The slice definition for Provenance.reason has a minimum of 0 but the slices add up to a minimum of 1")),
+                () -> assertTrue(resultMessages.toString().contains(
+                        "The repeating element has a pattern. The pattern will apply to all the repeats (this has not been clear to all users)"))
+        );
     }
 
     @Test
--- a/dpc-common/src/test/java/gov/cms/dpc/fhir/validations/AttestationValidationTest.java
+++ b/dpc-common/src/test/java/gov/cms/dpc/fhir/validations/AttestationValidationTest.java
@@ -55,14 +55,23 @@ class AttestationValidationTest {
     void definitionIsValid() {
         final StructureDefinition provenanceDefinition = dpcModule.fetchStructureDefinition(AttestationProfile.PROFILE_URI);
         final ValidationResult result = fhirValidator.validateWithResult(provenanceDefinition);
+
         ArrayList<String> errorMessages = new ArrayList<>();
         for (SingleValidationMessage msg: result.getMessages()) {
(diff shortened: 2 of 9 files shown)


dpc-common/pom.xml
```diff
@@ -47,10 +47,6 @@
             <groupId>${hapi.fhir.groupID}</groupId>
             <artifactId>hapi-fhir-structures-dstu3</artifactId>
         </dependency>
-        <dependency>
-            <groupId>${hapi.fhir.groupID}</groupId>
-            <artifactId>hapi-fhir-validation</artifactId>
-        </dependency>
```
ashley-weaver, line 53: this dependency was duplicated in this file.

dpc-common/src/test/java/gov/cms/dpc/fhir/validations/AttestationValidationTest.java
```diff
@@ -52,8 +54,24 @@ static void setup() {
     void definitionIsValid() {
         final StructureDefinition provenanceDefinition = dpcModule.fetchStructureDefinition(AttestationProfile.PROFILE_URI);
         final ValidationResult result = fhirValidator.validateWithResult(provenanceDefinition);
-        assertAll(() -> assertEquals(1, result.getMessages().size(), "Should have a single message"),
-                () -> assertTrue(result.getMessages().get(0).getMessage().contains("Provenance.target: derived min (0) cannot be less than base min (1)")));
+
```
jdettmannnava, line 57: I don't understand why increasing failure from 1 to 4 is not a problem?

dpc-common/src/test/java/gov/cms/dpc/fhir/validations/PatientValidationTest.java
```diff
@@ -50,7 +53,22 @@ static void setup() {
     void definitionIsValid() {
         final StructureDefinition patientDefinition = dpcModule.fetchStructureDefinition(PatientProfile.PROFILE_URI);
         final ValidationResult result = fhirValidator.validateWithResult(patientDefinition);
-        assertTrue(result.isSuccessful(), "Should have passed");
+
```
jdettmannnava, line 56: I don't understand why going from success to 2 errors is not a problem?

dpc-common/src/test/java/gov/cms/dpc/fhir/validations/AttestationValidationTest.java
ashley-weaver: One of the messages (about the slice definition) is for the same reason as the already-expected message, the repeating element message I don't think there's a "fix" for, and I wasn't entirely sure about the last one.

dpc-common/src/test/java/gov/cms/dpc/fhir/validations/PatientValidationTest.java
```diff
@@ -50,7 +53,22 @@ static void setup() {
...
+        ArrayList<String> resultMessages = new ArrayList<>();
+        for (SingleValidationMessage msg: result.getMessages()) {
+            resultMessages.add(msg.getMessage());
+        }
+        String resultMessagesString = String.join(", ", resultMessages);
+
+        List<String> expectedMessages = List.of(
+                "Found # expecting a token name",
+                " The slice definition for Patient.identifier has a minimum of 0 but the slices add up to a minimum of 1"
```
MEspositoE14s, line 65: For this particular error, we can fix it by adding `"min": 1` to DPCPatient.json in the Patient.identifier block.  

I'm still trying to figure out the one about "Found # expecting a token name".  It seems like the HAPI FHIR validator is having a problem with the StructureDefinition provided by the HAPI FHIR library, but I'm not 100% sure yet.

Either way, we should probably fix the validations that we can fix.

dpc-common/src/test/java/gov/cms/dpc/fhir/validations/AttestationValidationTest.java
```diff
@@ -69,7 +86,7 @@ void testHasReason() {
         final ValidationResult result = fhirValidator.validateWithResult(provenance);
 
         assertAll(() -> assertFalse(result.isSuccessful(), "Should not have passed"),
-                () -> assertEquals(1, result.getMessages().size(), "Should have a single message"));
+                () -> assertEquals(1, result.getMessages().size(), "Should have two messages for reason slice"));
```
ashley-weaver, line 89: This went from one to two errors because the validation now accounts for missing nested values, so the two messages are for Provenance.reason and Provenance.reason.provenance-reason.

dpc-common/src/test/java/gov/cms/dpc/fhir/validations/PatientValidationTest.java
```diff
@@ -104,7 +105,7 @@ void testIdentifier() {
 
         final ValidationResult result = fhirValidator.validateWithResult(patient);
         assertAll(() -> assertFalse(result.isSuccessful(), "Should have failed validation"),
-                () -> assertEquals(1, result.getMessages().size(), "Should have a single failure"));
+                () -> assertEquals(2, result.getMessages().size(), "Should have two failures for ID slice"));
```
ashley-weaver, line 108: This went from one to two errors because the validation now accounts for missing nested values, so the two messages are for Patient.identifier and Patient.identifier.patient-mbi.

dpc-common/src/test/java/gov/cms/dpc/fhir/validations/AttestationValidationTest.java
```diff
@@ -63,12 +63,11 @@ void definitionIsValid() {
 
         List<String> expectedMessages = List.of(
                 "Element Provenance.target: derived min (0) cannot be less than the base min (1)",
```
MEspositoE14s, line 65: Similar to the other error, you can fix this one by setting min to 1 on line 23 of DPCAttestation.json.  This is basically saying that our Provenance.target can't have a min of 0, because the base Provenance.target it's derived from in the Attestation resource specifies that it has to have at least 1.

dpc-common/src/test/java/gov/cms/dpc/fhir/validations/AttestationValidationTest.java
```diff
@@ -63,12 +63,11 @@ void definitionIsValid() {
...
-                "The slice definition for Provenance.reason has a minimum of 0 but the slices add up to a minimum of 1",
                 "The repeating element has a pattern. The pattern will apply to all the repeats (this has not been clear to all users)",
```
MEspositoE14s, line 66: This one I don't think is a big deal.  Maybe we should exclude errors where the severity is INFORMATION and only consider the ones marked ERROR?

dpc-common/src/test/java/gov/cms/dpc/fhir/validations/AttestationValidationTest.java
ashley-weaver: if I change min to 1, all the other Attestation tests fail, so I think I'll keep it as-is.

MEspositoE14s: That's because once you set the value to 1, then all of the other Provenance resources in the tests will need a target  element added to be considered valid.  

Ideally, that's probably what should've been done 5 years ago when this was written, but you're right that changing it now can break things.  I just checked our sample Postman collection, and even it doesn't have a target, so making this change would break that, too.  Guess we'll just have to live with it 🤷

MEspositoE14s (approved):

jdettmannnava (approved):
LGTM

PR #2225, opened 2024-07-30
DPC-4180 remove prefix from logs

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/.DPC-4180

## 🛠 Changes

dpc-portal exec foreman sed filter updated

## ℹ️ Context

DPC Portal uses foreman to run both the rails app and the fake cpi api gateway. Foreman prefixes log messages with the proc emiting the log. The logs are output in json format, and these prefixes interfere with json processing in splunk. We tried to remove them with sed (the **best** editor), but its regular expression did not work in the container.

## 🧪 Validation

Pushed to dev and looked at the logs in splunk, which no longer had the prefix and could be parsed in the search bar.
### Screenshots
Before
Before

After
After


lukey-luke (approved):
Nice catch Jeff! LGTM

I think assigning the regex to a variable would make it easier to read and eliminate the need to understand what the regex means e.g.

```
export LOG_STR="06:21:25 web.1   | {'level': 'debug', 'time': 'blahblahblah'}"
export EVERYTHING_AFTER_PIPE_REGEX="s/[^\|]*\|//"                             
echo $LOG_STR | sed $EVERYTHING_AFTER_PIPE_REGEX                              
 {'level': 'debug', 'time': 'blahblahblah'}

```

PR #2230, opened 2024-08-06
Fix GitHub Actions

chris-ronning-ny:
## 🎫 Ticket
No Ticket

## 🛠 Changes

This updates the docker-compose commands that run in our GitHub Actions, so that our builds will succeed. See https://github.com/orgs/community/discussions/116610 for relevant discussion.

PR #2231, opened 2024-08-06
Attempt to fix Jenkins build

chris-ronning-ny:
## 🎫 Ticket

No ticket

## 🛠 Changes

The [PR to fix our GH Actions builds](https://github.com/CMSgov/dpc-app/pull/2230) appears to have broken Jenkins. So, this represents an attempt to fix the Jenkins build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,30 +77,30 @@ start-dpc: start-app start-portals
 
 start-db: ## Start the database
 start-db:
-	@docker-compose up start_core_dependencies
+	@docker compose up start_core_dependencies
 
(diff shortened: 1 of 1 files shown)


kyeah (approved):
Thanks! I see a bunch of `docker-compose` references in documentation and comments still — we will probably want to update those so we aren't mixing and matching (seems like that can cause some confusion.)

chris-ronning-ny:
@kyeah good call. I updated `docker-compose` references in docs + comments to `docker compose`.

PR #2236, opened 2024-08-09
DPC-4204: use https for link to invitation

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4204

## 🛠 Changes

- Templates for cd and ao invites explicitly use https if Rails.env.production? == true

## ℹ️ Context

- Link to invitation in production should be https protocol
- Rails.env just means not running in development or test mode. It should be 'production' for dev and test (and eventually prod) servers

## 🧪 Validation

Automated testing


ashley-weaver (approved):
LGTM

PR #2237, opened 2024-08-12
DPC-4221: Remove unnecessary hashing

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4221

## 🛠 Changes

Hashing of ssn for comparison with ssn in information from CPI API Gateway removed. Now doing straight string matching.

## ℹ️ Context

Original implementation assumed we would be storing the ssn in the session, so we set it up to compare hashes. We are not storing it in the session (we either retrieve it from our IdP for the check, or we use a stored foreign key), so the hashing no longer serves any useful purpose.

## 🧪 Validation

Manual and Automated testing


chris-ronning-ny (approved):
LGTM

PR #2240, opened 2024-08-13
Me/dpc 4219 update org and ao status

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4219

## 🛠 Changes

When an AO successfully registers their and the organizations `verification_status` is set to `approved`.

## ℹ️ Context

This is part of the AO registration workflow.

## 🧪 Validation

Added a couple tests to make sure the values are set when an AO registers and not set when a CD registers.

### DIFF
--- a/dpc-portal/app/controllers/invitations_controller.rb
+++ b/dpc-portal/app/controllers/invitations_controller.rb
@@ -36,8 +36,6 @@ class InvitationsController < ApplicationController
     Rails.logger.info(['User logged in',
                        { actionContext: LoggingConstants::ActionContext::Registration,
                          actionType: LoggingConstants::ActionType::UserLoggedIn }])
-
-    verify_ao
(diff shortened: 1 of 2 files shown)


dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -36,6 +36,8 @@ def register
     Rails.logger.info(['User logged in',
                        { actionContext: LoggingConstants::ActionContext::Registration,
                          actionType: LoggingConstants::ActionType::UserLoggedIn }])
+
+    verify_ao
```
jdettmannnava, line 40: I would probably put the logic in create_ao_org_link

MEspositoE14s: That definitely flows a little nicer.

chris-ronning-ny (approved):

PR #2241, opened 2024-08-13
refresh rosterEntity object w/ latest data before returning api response

lukey-luke:
## 🎫 Ticket

This is a work-in-progress draft for [DPC-4196](https://jira.cms.gov/browse/DPC-4196). This PR makes an additional call to RosterDAO before returning API response to client

## 🛠 Changes
Call `session.refresh()` in `{{baseUrl}}/v1/Group/{{group_id}}/$add`

## ℹ️ Context
I believe `this.rosterDAO.getEntity(rosterID)` is being cached.

> _Whenever you pass an object to save(), update() or saveOrUpdate(), and whenever you retrieve an object using load(), get(), list(), iterate() or scroll(), that object is added to the internal cache of the Session._

from [19.3. Managing the caches](https://docs.jboss.org/hibernate/core/3.3/reference/en/html/performance.html#performance-sessioncache)

> _It is possible to re-load an object and all its collections at any time, using the refresh() method. This is useful when database triggers are used to initialize some of the properties of the object._
> 
> ```
> sess.save(cat);
> sess.flush(); //force the SQL INSERT
> sess.refresh(cat); //re-read the state (after the trigger executes)
> ```
> 
from [11.3 Loading an object](https://docs.jboss.org/hibernate/core/3.6/reference/en-US/html/objectstate.html#objectstate-loading)

## 🧪 Validation

manually tested in Postman

Screenshot 2024-08-13 at 12 54 12 PM

### DIFF
--- a/dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/RosterDAO.java
+++ b/dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/RosterDAO.java
@@ -28,6 +28,11 @@ public class RosterDAO extends AbstractDAO<RosterEntity> {
         return Optional.ofNullable(this.get(rosterID));
     }
 
+    public void refresh(RosterEntity roster) {
+        currentSession().flush();
+        currentSession().refresh(roster);
+    }
+
     public List<RosterEntity> findEntities(UUID resourceID, UUID organizationID, String providerNPI, String patientReference) {
 
         // Build a selection query to get records from the database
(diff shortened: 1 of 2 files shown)


dpc-common/src/main/java/gov/cms/dpc/common/hibernate/attribution/DPCAbstractDAO.java
```diff
@@ -0,0 +1,16 @@
...
+
+import io.dropwizard.hibernate.AbstractDAO;
+import org.hibernate.SessionFactory;
+
+public class DPCAbstractDAO<E> extends AbstractDAO<E> {
+    public DPCAbstractDAO(SessionFactory sessionFactory) {
+        super(sessionFactory);
+    }
+
+    public void refresh(E databaseObject) {
```
MEspositoE14s, line 11: Can you add a quick [Javadoc](https://www.baeldung.com/javadoc) header on this method?  IntelliJ will fill a lot of it out when you type the `/**` and then it'll start populating all over the place in the IDE.

MEspositoE14s (commented):
Looks good to me.  You mentioned this being part 1, are you planning on adding onto this PR for the other operations, creating a new one, or just spinning off a new ticket?

MEspositoE14s (approved):
LGTM!  There have been a few changes to main, but I don't think any of them will conflict when you try to merge.

PR #2244, opened 2024-08-16
DPC-4187: Log user id to new relic if available

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4187

## 🛠 Changes

- New Relic license key added to secure envs
- Explicit New Relic config added to portal
- ApplicationController addes user id to New Relic transactions as a custom attribute

## ℹ️ Context

We want to be able to base on-call pages on currently active users.
- Addition of NEW_RELIC_MONITOR_MODE in docker compose disables monitoring locally (will log if set to true)
- Addition of newrelic.yml allows us to make sure it is registered (log will exist if running) and also debug. Based on dpc-web configuration

## 🧪 Validation

Showing up in New Relic verified:
Screenshot 2024-08-16 at 4 20 30 PM


dpc-portal/spec/requests/organizations_spec.rb
```diff
@@ -34,6 +34,11 @@
         get '/organizations'
         expect(assigns(:links)).to eq [link]
       end
+
+      it 'logs user_id to new relic' do
+        expect(NewRelic::Agent).to receive(:add_custom_attributes).with({ user_id: user.id })
```
jdettmannnava, line 39: New Relic uses camelCase for the variables, so it could be userId, but user_id is more Railsy and also distinguishes it as a custom attribute.

lukey-luke (approved):
Looks good to me. I'll bother you next week on steps to test in deployed environment.

PR #2250, opened 2024-08-22
enable IP address resource for non-prod

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4237

## 🛠 Changes

- Forbidden response to access IP Address endpoints only in prod
- Tests for IP Address endpoints re-enabled

## ℹ️ Context

Revert of [disabling commit](https://github.com/CMSgov/dpc-app/commit/0b07319e3d33007c48ae6d7512d60acbbc692cd6)
We had disabled this endpoint pending [DPC-3768](https://jira.cms.gov/browse/DPC-3768), but we need it for pentesting, so it needs to be enabled for lower environments.
I tried to mock environment parser to make sure it fails on prod, but it simply refused to be mocked.

## 🧪 Validation

Manual and automated tests.
Works locally:
Screenshot 2024-08-21 at 3 56 30 PM


MEspositoE14s (approved):
For an easy way to mock the environment you can use `SystemStubs`.  We have a previous example [here](https://github.com/CMSgov/dpc-app/blob/8b507776ca345381232b47e9ed55c7cbc1f7c7cb/dpc-common/src/test/java/gov/cms/dpc/common/logging/filters/SecretLoggingFilterUnitTest.java#L21).

Otherwise, 👍

jdettmannnava:
> For an easy way to mock the environment you can use `SystemStubs`. We have a previous example [here](https://github.com/CMSgov/dpc-app/blob/8b507776ca345381232b47e9ed55c7cbc1f7c7cb/dpc-common/src/test/java/gov/cms/dpc/common/logging/filters/SecretLoggingFilterUnitTest.java#L21).
> 
> Otherwise, 👍

I don't think this works in integration tests, at least I couldn't make it do so.

PR #2252, opened 2024-08-26
DPC-4079 Opt Out integration testing

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4079

## 🛠 Changes

- Export Makefile updated to allow all tests, unit tests, and integration tests
- Export S3 utils added DownloadFileFromS3 to support test
- Export TestIntegrationGenerateRequestFile verifies file with appropriate number of lines uploaded to S3
- Import TestIntegrationImportResponseFile verifies that the confirmation file is correct
- Import TestIntegrationImportResponseFile verifies that the database is updated correctly
- Add verify-bucket stanzas to github workflows for integration opt out lambdas

## ℹ️ Context

Integration tests were not validating the files created.
Removed profile from session.go because it wouldn't run locally until I removed it :shrug:
I tried to check the database but couldn't connect

## 🧪 Validation

Tests pass locally
Github actions pass


kyeah (commented):
Oh interesting! I just realized we have...two different layers of "integration testing" going on here.

1. Integration testing with local (fake) BFD S3, which is what you've updated
2. Integration testing with real infrastructure in the Test environment, which happens entirely in Github Actions (and is more akin to our smoke tests):

https://github.com/CMSgov/dpc-app/blob/main/.github/workflows/opt-out-import-test-integration.yml
https://github.com/CMSgov/dpc-app/blob/main/.github/workflows/opt-out-export-test-integration.yml

These changes are helpful in and of themselves, but it sounds like maybe we'll want a separate ticket to upgrade those Github Actions tests too 🤔

chris-ronning-ny (approved):
Looks reasonable to me!

kyeah (approved):
Missed the updates to this one 😄 Thank you!

PR #2259, opened 2024-09-03
DPC-4252 Credential Delegate name missing

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4252

## 🛠 Changes

- User names added/updated when registering
- Column removed from "Active Credential Delegates" table

## ℹ️ Context

Testing revealed we were not showing the Credential Delegate's name in the "Active Credential Delegates" table. It turns out that we had stopped persisting the name on create.
As long as we were in there, we removed a column for deleting a Credential Delegate's access to an organization, as we don't have the functionality and will not be implementing it for MVP.

## 🧪 Validation

Manual and Automated testing.
Screenshot 2024-09-03 at 1 29 45 PM


MEspositoE14s (approved):
Seems pretty straight forward 👍

PR #2263, opened 2024-09-06
DPC-4244 logout of login

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4244

## 🛠 Changes

- Changed default logout behavior to manually log out then redirect to idp for them to log out
- Added logged_out response to logindotgov controller to handle return from above
- Consolidated idp environment variables in application controller

## ℹ️ Context

We want users to be able to log out of the idp after they log out of the portal.
Required update to idp to allow redirect to logged_out

## 🧪 Validation

Manual and automated testing


chris-ronning-ny (approved):
lgtm

PR #2266, opened 2024-09-10
Me/dpc 4231 smoketest failing

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4231

## 🛠 Changes

Fixed the following issues:

### FHIR client timeouts on internal calls
When dpc-api would call dpc-attribution, or when dpc-aggregation would call dpc-consent, the FHIR client would occasionally time out.  We were setting a default timeout of 5 seconds when calling BFD and that setting was being carried over.  Now, dpc-api has a separate "attribution client" and dpc-aggregation has a separate "consent client" that each have their own timeout settings and are configured from their service's `application.yml` file.  This is abstracted into a `FHIRClientConfiguration` class.

### Our WAF is rate limiting us
Our WAF is currently rate limiting us to 300 requests every 5 minutes.  In the old version of the smoke tests, we were submitting 100 patients, 4 providers, 4 rosters and a bunch of GET requests one at a time, 3 times each, once for each smoke test thread.  In the new version, within each thread we batch all of the patient and provider submissions using their `$submit` operations, eliminating a few hundred requests and getting us under the new rate limit.  We also slow down the polling to monitor our export requests to once every 5 seconds, instead of every second.

### DPC-API and DPC-Attribution don't scale
When running the three smoke test threads in parallel, DPC becomes unresponsive and the load balancer begins returning HTTP 504 (gateway timeout) responses.  The one worker we have running in dev, test and prod-sbx just can't keep up with 3 smoke tests running at the same time.  To get the tests working, I updated the ramp up time so that the 3 threads are started over a period of 90 seconds instead of 30, so the work is spread out a bit.  This solves today's problem, but we'll need a real scaling solution before we eventually go live.

### Id bug in the $submit operation for the patient and practitioner end points
When doing a batch submit using the `$submit` operation for either resource, we'll attempt to use the FHIR resource id sent in by the customer if they provide one.  This is a problem, because that field is also the primary key of the `patient` and `provider` tables, so when more than one organization sends in a patient or provider with the same id, the second one will fail, which is exactly what we're doing with the smoke tests.  To get the tests working, I removed the ids from the synthetic data we're testing with.  Just like the above bullet point, this will need to be fixed before going live.  ([DPC-4270](https://jira.cms.gov/browse/DPC-4270) created to look into this.)

## ℹ️ Context

Our smoke tests have been consistently failing for months and this was an attempt to get them working again.

## 🧪 Validation

Got three [Full Deploy and Release](https://management.dpc.cms.gov/job/DPC%20-%20Smoke%20Test/5721/) cycles in a row with passing smoke tests.


dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/service/ConsentServiceImplUnitTest.java
```diff
@@ -121,4 +121,4 @@ private Consent createTestConsent(String mbi){
         consent.setId(UUID.randomUUID().toString());
         return consent;
     }
-}
\ No newline at end of file
+}
```
jdettmannnava, line 124: My favorite!

dpc-common/src/test/java/gov/cms/dpc/fhir/configuration/FHIRClientConfigurationUnitTest.java
```diff
@@ -0,0 +1,19 @@
...
+class FHIRClientConfigurationUnitTest {
+	private FHIRClientConfiguration fhirClientConfiguration = new FHIRClientConfiguration();
+	@Test
+	public void testSettersAndGetters() {
+		TimeoutConfiguration timeouts = new TimeoutConfiguration();
+
+		fhirClientConfiguration.setTimeouts(timeouts);
+		fhirClientConfiguration.setServerBaseUrl("url");
+		assertEquals(timeouts, fhirClientConfiguration.getTimeouts());
+		assertEquals("url", fhirClientConfiguration.getServerBaseUrl());
```
jdettmannnava, line 16: I give up -- where is the configuration for this test set?

dpc-smoketest/src/main/java/gov/cms/dpc/testing/smoketests/ClientUtils.java
```diff
@@ -149,7 +163,8 @@ private static JobCompletionModel awaitExportResponse(String jobLocation, String
         boolean done = false;
 
         while (!done) {
-            Thread.sleep(1000);
+            // Our WAF rate limits us to 300 requests every 5 minutes, so don't poll too often
+            Thread.sleep(5000);
```
jdettmannnava, line 167: Heck, make it 20000! The extra 15 seconds is not going to move the needle...

jdettmannnava (approved):
LGTM - but please answer questions!

dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/service/ConsentServiceImplUnitTest.java
MEspositoE14s: I'll give the credit for this one to IntelliJ.

dpc-common/src/test/java/gov/cms/dpc/fhir/configuration/FHIRClientConfigurationUnitTest.java
MEspositoE14s: In this case, there isn't one.  I hate these types of tests.  All the class does it set and get the server's timeouts and url, so all the test does is set and get two dummy values.  Normally I wouldn't bother, but SonarQube wants a certain percentage of coverage on added code so 🤷‍♂️

PR #2268, opened 2024-09-17
PLT-639: Adding gitleaks MBI scanning & updating gitleaks pre-commit version

ildesenesence:
## 🎫 Ticket

## 🛠 Changes

Gitleaks custom config extension was added (.gitleaks.toml) allowing for a regex-based MBI filter to be run against new commits, and the gitleaks pre-commit version was updated to 8.19.x which doesn't affect us at this time.

## ℹ️ Context

Bulk API Platform Team has an ongoing goal of improving security and safeguarding PHI/PII for our members, to that end we are applying a belt-and-suspenders approach to preventing leakage of data (such as Medicare Beneficiary Identifiers) in github.

##   Validation

These changes were tested locally and automatically as they make modifications to the pre-commit functionality, and were initially tested and validated against a pregenerated file with MBI data in PLT-532.

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -1,0 +1,9 @@
+title = "DASG Standard"
+
+[extend]
+  useDefault = true
+
+[[rules]]
+  id = "mbi-detection"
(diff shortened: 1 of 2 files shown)


jdettmannnava (approved):
LGTM

PR #2269, opened 2024-09-17
[DPC-4259] Add lambda for syncing ip address set

ashley-weaver:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4259

## 🛠 Changes

Adds lamdba to keep IP sets in sync on AWS.

## ℹ️ Context

By default, the production DPC API is not open to the public network. Organizations must provide IP addresses for inclusion in DPC's network allowlist so that they can access the DPC API.

We are currently storing these IP addresses in the database, but they are not automatically synced to the AWS WAF IP Set for functional purposes.

## 🧪 Validation

main_test.go includes an integration test that updates the WAF set, verifies the update, and then resets it.


MEspositoE14s (approved):
Two quick points but otherwise looks good to me.

- Can we add a read me or something that tells people they have to setup credentials from CloudTamer before they can run `make test` locally?
- Definitely outside the scope of this ticket, but we should consider another ticket to create a shared library so we don't have duplicate code in all three of our lambdas.  The stuff we use to connect to AWS, get secrets and connect to the DB is repeated in all of them.

ashley-weaver:
> * Definitely outside the scope of this ticket, but we should consider another ticket to create a shared library so we don't have duplicate code in all three of our lambdas.  The stuff we use to connect to AWS, get secrets and connect to the DB is repeated in all of them.

Agreed on pulling the AWS/DB out into a common directory; I tried to make this work, but I wasn't able to.

PR #2270, opened 2024-09-18
Fix create bug in portal

jdettmannnava:
## 🎫 Ticket

No ticket

## 🛠 Changes

- user info passed to create user in invitations controller
- user info method in invitations spec renamed because bled into controller

## ℹ️ Context

Bug in invitations controller was masked by a method in the spec

## 🧪 Validation

Automated and manual testing


MEspositoE14s (commented):
Looks good to me 👍

MEspositoE14s (approved):
Doh... Clicked the wrong button!

PR #2273, opened 2024-09-20
DPC-4258 Run sonarqube check in GitHub actions

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4258

## 🛠 Changes

- Saved results of tests from webapps' workflows for use in sonarqube evaluation
- Added sonarqube evaluations of portal, web/admin, and app

## ℹ️ Context

We had been running sonarqube evaluation in deployment, but it makes much more sense to have code analysis/failure block merge than deployment, so we moved it here.

## 🧪 Validation

- Checked output of ci-workflow and verified that sonarqube runs for each context
- Tested dpc-app by adding class without corresponding test; failed github action
- Tested dpc-web by adding class without corresponding test; failed github action
- Tested dpc-admin by adding class without corresponding test; failed github action
- Tested dpc-portal by adding class without corresponding test; failed github action


MEspositoE14s (approved):
No more last minute SQ failures on deploy! 🥳

PR #2280, opened 2024-09-27
[DPC-4299] - Update AO job so sanctions don't affect org API access

lukey-luke:
## 🎫 Ticket

[DPC-4299](https://jira.cms.gov/browse/DPC-4299)

## 🛠 Changes

- remove code that updates ProviderOrganization verification status in [verify_ao_job.rb](https://github.com/CMSgov/dpc-app/blob/acb5c1a19f3274afbdd8aad2ce3cff4be93d0c08/dpc-portal/app/jobs/verify_ao_job.rb#L7)

## ℹ️ Context

 - an AO failing med sanction check would cause an entire organization to also be rejected. [disable_rejected()](https://github.com/CMSgov/dpc-app/blob/ca55c2499cdc2503929c10f00245ea65e76adb1d/dpc-portal/app/models/provider_organization.rb#L64) would delete the api tokens for the organization
 - We no longer want this behavior, so now only the AO user is rejected

## 🧪 Validation

- existing test cases already cover this sidekiq job in [verify_ao_job_spec.rb](erify_ao_job_spec.rb)
- failing test cases updated to reflect changes

### DIFF
--- a/dpc-portal/spec/jobs/verify_ao_job_spec.rb
+++ b/dpc-portal/spec/jobs/verify_ao_job_spec.rb
@@ -198,9 +198,6 @@ RSpec.describe VerifyAoJob, type: :job do
             expect_log_for(link, 'ao_med_sanctions')
           end
           VerifyAoJob.perform_now
-          links.each do |link|
-            expect_audits(link, also: %i[user org])
-          end
         end
         it 'neither former nor current org link updated when ao verification status rejected' do
           user = create(:user, pac_id: '900666666', verification_status: :approved)
--- a/dpc-portal/spec/jobs/verify_ao_job_spec.rb
+++ b/dpc-portal/spec/jobs/verify_ao_job_spec.rb
@@ -180,7 +180,7 @@ RSpec.describe VerifyAoJob, type: :job do
           end
         end
         # @@@
-        it "should update user, but not impact ProviderOrganization.verification_status" do
+        it 'should update user, but not impact ProviderOrganization.verification_status' do
           expect(AoOrgLink.where(last_checked_at: ..6.days.ago).count).to eq 1
           VerifyAoJob.perform_now
           links.each do |link|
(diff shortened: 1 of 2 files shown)


NavaCharlesHorowitz (commented):
If your preference is to comment out the code, could you add a comment to explain the retention/reason for disabling? Or else just delete these calls if we're done with them?  Thx!

dpc-portal/spec/jobs/verify_ao_job_spec.rb
```diff
@@ -198,11 +198,8 @@ def expect_audits(link, also: [])
             expect_log_for(link, 'ao_med_sanctions')
           end
           VerifyAoJob.perform_now
-          links.each do |link|
-            expect_audits(link, also: %i[user org])
```
ashley-weaver, line 202: I don't have all the context here, but should this still check the link and user audits, ie just remove `org` from the also params?

dpc-portal/spec/jobs/verify_ao_job_spec.rb
```diff
@@ -188,8 +189,7 @@ def expect_audits(link, also: [])
             expect(link.verification_reason).to eq 'ao_med_sanctions'
             expect(link.user.verification_status).to eq 'rejected'
             expect(link.user.verification_reason).to eq 'ao_med_sanctions'
-            expect(link.provider_organization.verification_status).to eq 'rejected'
-            expect(link.provider_organization.verification_reason).to eq 'ao_med_sanctions'
+            expect(link.provider_organization.verification_status).to eq 'approved'
```
ashley-weaver, line 192:
```suggestion
            expect(link.provider_organization.verification_status).to eq 'approved'
            expect(link.provider_organization.verification_reason).to be nil
```

ashley-weaver (approved):
LGTM, just a few nitpicks

PR #2281, opened 2024-09-30
Remove id elements from prod smoketest test data

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4302

## 🛠 Changes

Removed `id` elements from prod smoke test resource files.

## ℹ️ Context

When the smoke tests run they load the resources from these files, and then remove them when they're finished.  If the first smoke test takes a long time, then the second will sometimes start before the first is finished and try to insert the same resources.  This leads to a primary key collision and the tests fail.  Removing the ids allows the system to generate random ones on insert so the tests can run in parallel without problems.

Note: This was already done on the test files used in the other environments.

## 🧪 Validation

Smoke tests ran successfully in prod [here](https://management.dpc.cms.gov/job/DPC%20-%20Smoke%20Test/5774/).

### DIFF
--- a/src/main/resources/prod_patient_bundle-dpr.json
+++ b/src/main/resources/prod_patient_bundle-dpr.json
@@ -11,7 +11,6 @@
           {
             "resource": {
               "resourceType": "Patient",
-              "id": "728b270d-d7de-4143-82fe-d3ccd92cebe4",
               "meta": {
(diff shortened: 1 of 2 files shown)


ashley-weaver (approved):
👍

PR #2285, opened 2024-10-08
[DPC-4229] Removed given name and phone matching on cd invite

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4229

## 🛠 Changes

In the CD invitation flow we no longer match on given name and phone number.  

## ℹ️ Context

These changes were made in an attempt to streamline the CD invitation process.

## 🧪 Validation

Tests updated to verify that first name and phone number are no longer required to match.  

**Note**: We still require that the invitation has a given name and phone number, we just don't match on them.


ashley-weaver (approved):
LGTM

PR #2295, opened 2024-10-22
[DPC-4333] Me/dpc 4333 portal network connectivity

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4333

## 🛠 Changes

- Creates a new `VerifyResourceHealthJob` that periodically checks to see if the sidekiq-portal has a healthy connection to dpc-api, the CPI API gateway and the IDP, and logs its results as Cloudwatch metrics (1=healthy, 0=not healthy).
- Schedules the new job to run every 1 minute.
- Adds a healthcheck to the `DpcClient`.
- Adds a healthcheck to the `CpiApiGatewayClient`.

## ℹ️ Context

We had failures in testing due to a misconfiguration in our WAF preventing the portal from reaching dpc-api, both services were up, though, so no alarms were triggered.  This will proactively let us know if the portal loses the connection to one of its dependent services before the users start seeing errors.

This PR requires https://github.com/CMSgov/dpc-ops/pull/745.

## 🧪 Validation

Deployed to dev and verified metrics were created correctly:
image


dpc-portal/app/jobs/verify_resource_health_job.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+require 'aws-sdk-cloudwatch'
+
+# A background job that verifies that external services are up and accessible.
+class VerifyResourceHealthJob
+  include Sidekiq::Job
```
MEspositoE14s, line 7: I originally had this setup as an ActiveJob, like most of our other jobs, but there's an old issue using sidekiq with it and cron jobs.  It worked, but it logged an error on every run, and the suggested solution was to use Sidekiq::Job instead.

dpc-portal/app/jobs/verify_resource_health_job.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def dimensions
+    [
+      {
+        name: 'Type',
+        value: 'healthcheck'
+      },
+      {
+        name: 'environment',
+        value: ENVIRONMENT
```
MEspositoE14s, line 72: I hate that "Type" is capitalized and "environment" is lower case, but it matches the metrics we already have in Cloudwatch so they all get grouped together.

dpc-portal/app/services/cpi_api_gateway_client.rb
```diff
@@ -4,7 +4,7 @@
 
 # A client for requests to the CPI API Gateway
 class CpiApiGatewayClient
-  attr_accessor :access
+  attr_accessor :access, :client
```
jdettmannnava, line 7: Unclear why we need this accessor

engines/api_client/app/services/dpc_client.rb
```diff
@@ -182,16 +187,17 @@ def post_text_request(uri_string, json, query_params, token)
   def http_request(request, uri)
     http = Net::HTTP.new(uri.host, uri.port)
 
-    if use_ssl?
+    if use_ssl? && (uri.scheme == 'https')
```
jdettmannnava, line 190: Maybe move this to use_ssl?

engines/api_client/spec/services/dpc_client_spec.rb
```diff
@@ -794,6 +795,53 @@
     end
   end
 
+  describe '#get healthcheck' do
+    context 'successful api request' do
+      it 'calls healthcheck' do
+        stub_request(:get, 'http://dpc.example.com/healthcheck').with(
```
jdettmannnava, line 801: I'm not sure what the 'with' gives us.

dpc-portal/spec/jobs/verify_resource_health_job_spec.rb
```diff
@@ -0,0 +1,247 @@
...
+    describe 'everything healthy' do
+      it 'should emit healthy metrics' do
+        VerifyResourceHealthJob::IDP_HOST = 'www.idp_test.com'
+        stub_request(:get, 'https://www.idp_test.com').to_return(status: 200)
+
+        expect(mock_dpc_client).to receive(:healthcheck)
+        expect(mock_dpc_client).to receive(:response_successful?).twice.and_return(true)
+        expect(mock_cpi_client).to receive(:healthcheck).and_return(true)
+        expect(mock_cloudwatch_client).to receive(:put_metric_data).with(
+          put_metric_data_parms(
```
jdettmannnava, line 33: Can we DRY this up a little? It is hard to see what changes between tests.

dpc-portal/config/schedule.yml
```diff
@@ -2,3 +2,8 @@ verify_ao_job:
   cron: "0 1 * * 0"
   class: "VerifyAoJob"
   queue: portal
+
+verify_resource_health_job:
+  cron: "*/1 * * * *"
```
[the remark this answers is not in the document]
MEspositoE14s: I admit I pulled that number out of thin air.  AWS defaults to 30 seconds for its load balancer health checks and that seemed a little too frequent for me, so I doubled it.  I'm definitely open to suggestions, though.

dpc-portal/spec/jobs/verify_resource_health_job_spec.rb
```diff
@@ -0,0 +1,247 @@
...
+  before do
+    allow(Aws::CloudWatch::Client).to receive(:new).with(region: 'us-east-1').and_return(mock_cloudwatch_client)
+  end
+
+  let(:job) { VerifyResourceHealthJob.new }
+
+  context 'can successfully send metrics' do
+    describe 'everything healthy' do
+      it 'should emit healthy metrics' do
+        VerifyResourceHealthJob::IDP_HOST = 'www.idp_test.com'
```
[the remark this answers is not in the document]
MEspositoE14s: I'm adding it there now.  I didn't even think to look for it!

dpc-portal/app/services/cpi_api_gateway_client.rb
```diff
@@ -65,6 +65,15 @@ def org_info(npi)
     fetch_provider_info(body)
   end
 
+  # The CPI API Gateway doesn't support a healthcheck, and their suggestion was to just hit one of their
+  # end points and see if we get a response.  Don't over use this, as it counts against our rate limit.
+  def healthcheck
```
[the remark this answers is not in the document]
MEspositoE14s: Doh!  You're right.  For some reason I had it in my head that they were the same url.

dpc-portal/app/services/cpi_api_gateway_client.rb
MEspositoE14s: This was there because I was updating the client in the previous version of the health check's tests.  It's no longer needed.

engines/api_client/spec/services/dpc_client_spec.rb
MEspositoE14s: In hindsight, not much.  Removing it.

dpc-portal/app/services/cpi_api_gateway_client.rb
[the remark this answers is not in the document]
jdettmannnava: It might be a good idea to also have this check as a separate thing? Good to know if it is the CPI gateway or IDM that is down...

dpc-portal/app/services/cpi_api_gateway_client.rb
MEspositoE14s: Put this back in because it makes it easier to sub-out the OAuth client when testing getting a failed token.

engines/api_client/app/services/dpc_client.rb
MEspositoE14s: I didn't realize I could pass parameters to a method that ends in "?".  I guess you learn something new everyday!

dpc-portal/app/services/cpi_api_gateway_client.rb
```diff
@@ -58,13 +54,29 @@ def org_info(npi)
...
+    # We'll get a 400 because the npi is bad, but any response indicates they're up and we're connected.
+    org_info('fake_npi')
+    true
+  rescue StandardError
+    false
+  end
+
+  def healthy_auth?
+    # Check if we can get a token
+    @client.client_credentials.get_token(scope: 'READ').nil?
```
jdettmannnava, line 74: You could also just check if request_client.nil?

jdettmannnava (approved):
LGTM! Something you can address if you want

dpc-portal/app/services/cpi_api_gateway_client.rb
MEspositoE14s: I was thinking about that, but I was worried there might be some scenario where we use a previously cached client even though the OAuth connection is down.  Admittedly, probably not super likely.

PR #2298, opened 2024-10-25
[DPC-4213] Add error on create existing group

MEspositoE14s:
## 🎫 Ticket

https://github.com/CMSgov/dpc-app/compare/me/dpc-4213-create-duplicate-rosters?expand=1

## 🛠 Changes

Attempting to create a roster that already exists for a provider in an org will return an error.  Previously, the request would be ignored, and the already existing roster would be returned with a 200.

## ℹ️ Context

After much [discussion](https://cmsgov.slack.com/archives/CJL0FRN2E/p1722519081409839), I went with the majority and made the error a 403.  If we change our minds before this is merged it's only a two line change.

## 🧪 Validation

Ran locally and verified.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,8 +1,20 @@
 version: '3'
 
 services:
-
   db:
     ports:
       - "5432:5432"
 
+  # Open up the debugger port on each service
+  api:
+    ports:
+      - "5005:5005"
+  aggregation:
+    ports:
+      - "5006:5005"
+  attribution:
+    ports:
+      - "5007:5005"
+  consent:
+    ports:
+      - "5008:5005"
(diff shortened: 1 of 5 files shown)


docker-compose.override.yml
```diff
@@ -1,8 +1,20 @@
...
+      - "5005:5005"
+  aggregation:
+    ports:
+      - "5006:5005"
+  attribution:
+    ports:
+      - "5007:5005"
+  consent:
+    ports:
+      - "5008:5005"
```
MEspositoE14s, line 20: This was just to make debugging integration tests a little easier.

dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/GroupResourceTest.java
```diff
@@ -621,33 +621,6 @@ public void testOrgCanOnlyAddPatientsTheyManageToGroup() throws IOException, URI
         conn.disconnect();
     }
 
-    @Test
-    public void testApiDoesNotUseOrgTagSpecifiedByClient() throws GeneralSecurityException, IOException, URISyntaxException {
```
MEspositoE14s, line 625: This test is no longer possible since we can't "create" an existing roster anymore.  The functionality is replaced with a unit test.

jdettmannnava (approved):
LGTM

PR #2302, opened 2024-10-28
Remove console logging when running rspec

jdettmannnava:
## 🎫 Ticket

No ticket

## 🛠 Changes

- VerifyResourceHealthJob use Rails.logger instead of logger
- Stub IDP_HOST in spec

## ℹ️ Context

Sidekiq:Job has its own logger that pushes to STDOUT, so it logs to console when running tests. And setting the constant manually was writing warnings to the console.

## 🧪 Validation
All tests pass without writing anything to the console


ashley-weaver (approved):
just noticed this myself, thanks for fixing it!

PR #2313, opened 2024-11-01
[DPC-4246] Disable default logging in opt-out-import lambda

ashley-weaver:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4246

## 🛠 Changes

Removed default logging from AWS SDK Golang config.

## ℹ️ Context

The AWS SDK (Golang, v2) automatically logs errors to stderr in a non-JSON format that is not parseable by Splunk. This is not necessary since our application itself will catch and log errors in the standardized format, with the standardized log information.

### DIFF
--- a/lambda/opt-out-import/main.go
+++ b/lambda/opt-out-import/main.go
@@ -22,6 +22,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/aws/smithy-go/logging"
 
 	"github.com/ianlopshire/go-fixedwidth"
 )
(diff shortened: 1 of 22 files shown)


lambda/opt-out-import/main.go
```diff
@@ -194,9 +195,13 @@ func importResponseFile(bucket string, file string) (int, int, string, error) {
 }
 
 func createV2Cfg() (*awsv2.Config, error) {
-	cfg, err := config.LoadDefaultConfig(context.TODO(), config.WithRegion("us-east-1"))
-	assumeRoleArn, err := getAssumeRoleArn()
+	// disable default logging from aws-sdk-go-v2
+	cfg, err := config.LoadDefaultConfig(context.TODO(), config.WithLogger(logging.Nop{}))
```
jdettmannnava, line 199: Do we not need region?

jdettmannnava:
Any way to validate?

lambda/opt-out-import/main.go
ashley-weaver: this is to remove the default logging entirely, so region is unnecessary

jdettmannnava: Is this not the config for everything, not just logging?

ashley-weaver: this is how the docs say to configure logging
https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/logging/
https://pkg.go.dev/github.com/aws/smithy-go/logging

jdettmannnava: Sure, but I think you can leave the region in, as well?

jdettmannnava (approved):
LGTM

PR #2318, opened 2024-11-06
[DPC-3785] Mask Postgres details in application logs

lukey-luke:
## 🎫 Ticket
[DPC-3875](https://jira.cms.gov/browse/DPC-3785)

## 🛠 Changes
 - adds additional masking method for stripping out PSQL details to existing DPCJsonLayout logging layout class
 - this requires changes to both "exception" and "message" fields of incoming log messages.

## ℹ️ Context
Even when database entities (i.e. [OrganizationEntity](https://github.com/CMSgov/dpc-app/blob/99f63fd7ccbdf603c246f6c062eb874f4053cd0f/dpc-common/src/main/java/gov/cms/dpc/common/entities/OrganizationEntity.java)) are not explicitly logged,  if an exception is raised, information from the database may show up in system logs. (See validation section below for example)

## 🧪 Validation

 - manual testing done with the Organization `$submit` endpoint
 - making this API call multiple times raises a unique constraint violation
 - see also automated unit test added to `dpc-common/src/test/java/gov/cms/dpc/common/logging/DPCJsonLayoutUnitTest.java`


MEspositoE14s (approved):
Might as well throw in a test around the other pattern, `PSQL_BATCH_ENTRY_EXCEPTION_PATTERN`, too, but seems good to me.

PR #2323, opened 2024-11-17
.gitignore update to exclude project tmp/ contents

NavaCharlesHorowitz:
Based on 10/21/2024 security incident in which a logfile in project tmp/ was inadvertently committed and pushed to a branch (and caught by github secret scanning as containing a logged license key), updated .gitignore to ignore all contents in project tmp/ directory, but to create an empty directory (with .keep placeholder) to support local tmp work as necessary.

## 🎫 Ticket

This ticket provides the relevant background. No separate ticket was opened for the update of the gitignore, which was discussed in post-mortem.
https://jira.cms.gov/browse/DPC-4341

## 🛠 Changes

.gitignore was updated to ignore all tmp/ contents except tmp/.keep
tmp/.keep was added to pre-establish a tmp directory in the repo to be used as necessary without its contents being included in commits.

## ℹ️ Context

Change was made to permit typical uses of tmp directory in local development without accidentally committing contents to github

## 🧪 Validation

Verified that files added to tmp/ (except empty placeholder .keep) are ignored for staging commits.


chris-ronning-ny (approved):
Good thinking, thanks!

PR #2324, opened 2024-11-17
Docker, Makefile, and Script Updates

NavaCharlesHorowitz:
## 🎫 Ticket

This is one component of my catchall tech debt/housekeeping ticket:
https://jira.cms.gov/browse/DPC-4298

## 🛠 Changes

-Replaced dadarek containers with docker v3 dependencies and healthchecks
-Updated Makefile with better targets
-Updated test scripts to use make targets 
-Added some variables for flexibility in setting docker project -Replaced bash scripting with posix because of support variations in cicd -Replaced curl with wget in dpc-admin and dpc-web entrypoints because there is no curl in images

## ℹ️ Context

Simplify and streamline Docker  and system lifecycle work, add flexibility through configurability, reduce warning clutter

## 🧪 Validation

Ran through system lifecycle, used Makefile targets, ran scripts.


chris-ronning-ny:
I'm concerned about how much this is going to change my day to day dev workflow, and it's not quite obvious to me just how drastic a change it's going to be. I would like to request some documentation to be added to this pull request, possibly in a new section of the README.

NavaCharlesHorowitz:
Good call -- I had some nice doc updates but lost them when I blew away the original branch and PR.  I will update docs as the last commit to this PR asap after I get main merged in with no problems.

PR #2326, opened 2024-11-18
Tech debt - Maven lifecycle updates

NavaCharlesHorowitz:
## 🎫 Ticket

This is one component of my catchall tech debt/housekeeping ticket: 
[DPC-4298](https://jira.cms.gov/browse/DPC-4298)

## 🛠 Changes

Updates to project POMs to implement Failsafe for integration test rather than reusing Surefire (generally limited to unit test)
Renamed integration test files consistent with Failsafe defaults (preserving commit history on these files)
Added test display names
Added test reporter to clean up maven output during build/test (flags added to enable out/err as needed)
Knocked out shade warnings
Updated maven plugin versions
Added maven clean plugin with config because CICD filelocks are slow to release

## ℹ️ Context

Maven lifecycle activities, specifically package and test, were too cluttered with warnings and exception logging, making it difficult to understand the state of the software, quality and purpose of tests, and the results of specific development actions. 
Given that development work is typically very scoped, reducing 99% of the noise produced makes it easier to focus on evaluating specific changes made. Easier to understand and evaluate the software as a whole based on monitoring maven lifecycle.

## 🧪 Validation

Ran local dpc-test.sh to verify. Encountered sporadic issues with final mvn clean compile of dpc-smoketest which is redundant, so I've disabled it temporarily.  I believe changes made to cicd environment for csrap may have affected environment behaviors


dpc-attribution/src/test/java/gov/cms/dpc/attribution/AbstractAttributionDAOTest.java
```diff
@@ -2,6 +2,7 @@
 
 import gov.cms.dpc.common.entities.*;
 import gov.cms.dpc.testing.AbstractMultipleDAOTest;
+import org.junit.jupiter.api.DisplayName;
```
MEspositoE14s, line 5: Can we remove this import since it's not being used?  The linter in IntelliJ will flag it as an unused import and remove it the first time one of us using IntelliJ touches the file, anyway.

dpc-macaroons/src/test/java/gov/cms/dpc/macaroons/verifiers/ExpirationVerifierTest.java
```diff
@@ -8,6 +8,7 @@
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.time.temporal.ChronoUnit;
+import org.junit.jupiter.api.DisplayName;
```
MEspositoE14s, line 11: I think this is the only other one.

dpc-test.sh
```diff
@@ -61,7 +61,7 @@ docker compose -p start-v1-app up start_api_dependencies
 AUTH_DISABLED=true docker compose -p start-v1-app up start_api start_consent
 
 # Run the Postman tests
-npm install
+# npm install
```
MEspositoE14s, line 64: Was this just for testing, or do you want it turned off permanently?

MEspositoE14s (commented):
Definitely a fan of the tree-reporter (and the emojis)!

dpc-test.sh
NavaCharlesHorowitz: Good question - here was my experience: 

After cloning dpc-app and running make ci-app (or dpc-test.sh directly), everything ran ok,. But subsequent runs would result in partial failure in the final actions of the end-to-end system tests.  Don't recall how I figured that this npm install command was a culprit, but commenting this out resolved the issue.  I've noticed a couple things that I run into locally that others haven't seemed to encounter, so this may be one of those things that can affect some and not others.  I'd say if you aren't having any problems with/without the settings, we can remove it and presume that if someone follows the developer setup docs, their environment will be prepared for the s/w and don't need the built-in install command.

Your call though.  We could further explore why I was/am getting this problem if you'd rather.

dpc-attribution/src/test/java/gov/cms/dpc/attribution/AbstractAttributionDAOTest.java
NavaCharlesHorowitz: yep, totally in favor of import cleanups

dpc-test.sh
MEspositoE14s: Hmm... I uncommented and ran locally a few times and everything's coming up green.  

At least in the long term, it's probably worth uncommenting and figuring out why it's not working on your machine.  In the short term, I don't think it's a deal breaker to leave it commented out for now just to get this across the line and coming back to it later.

PR #2327, opened 2024-11-18
[DPC-4390] Log organization credential status via Sidekiq

lukey-luke:
## 🎫 Ticket
[DPC-4390](https://jira.cms.gov/browse/DPC-4390)

## 🛠 Changes
- added new sidekiq job to call api and get status of organization's api access
- log credential status for ea organization
- log aggregate stats of total organizations with/without api access

## ℹ️ Context
- this ticket was created in order to enable [DPC-4381](https://jira.cms.gov/browse/DPC-4381)

## 🧪 Validation

 - see dpc-portal/spec/jobs/log_organizations_access_spec.rb

PR #2330, opened 2024-11-20
Add exception handling to LogOrganizationsApiCredentialStatusJob

lukey-luke:
## 🎫 Ticket

This is a follow-up to https://github.com/CMSgov/dpc-app/pull/2327 for [DPC-4390](https://jira.cms.gov/browse/DPC-4390).

## 🛠 Changes
 - wrap DpcApiClient calls so if an exception is raised, the sidekiq job can continue to run for organizations with successful API calls (see context below)

## ℹ️ Context
 - following the automatic deployment to the dev environment, I observed that the sidekiq job was not emitting the expected message in the logs.
  - it appears that one of the organizations on the dev environment has an invalid UUID, causing get_client_tokens() to fail
 
Screenshot 2024-11-19 at 5 58 44 PM
Screenshot 2024-11-19 at 5 58 47 PM

### DIFF
--- a/dpc-portal/spec/jobs/log_organizations_api_credential_status_spec.rb
+++ b/dpc-portal/spec/jobs/log_organizations_api_credential_status_spec.rb
@@ -86,5 +86,29 @@ RSpec.describe LogOrganizationsApiCredentialStatusJob, type: :job do
 
       described_class.perform_now
     end
+    it 'processes next organization successfully when an exception is raised' do
+      second_organization = create(
+        :provider_organization,
+        name: 'Test2',
+        dpc_api_organization_id: 'bar',
+        terms_of_service_accepted_by: user,
+        terms_of_service_accepted_at: 1.day.ago
+      )
+      second_organization.save!
+      provider_organization.save!
+
+      allow_any_instance_of(described_class).to receive(:fetch_credential_status).with('foo').and_return(
+        num_tokens: 1, num_keys: 2, num_ips: 3
+      )
+      allow_any_instance_of(described_class).to receive(:fetch_credential_status).with('bar').and_raise(
+        StandardError.new('something failed inside fetch_credential_status!')
+      )
+
+      allow(Rails.logger).to receive(:info)
+      expect(Rails.logger).to receive(:info).with(['Organizations API credential status',
+                                                   { have_active_credentials: 1,
+                                                     have_incomplete_or_no_credentials: 0 }])
+      described_class.perform_now
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


dpc-portal/spec/jobs/log_organizations_api_credential_status_spec.rb
```diff
@@ -60,31 +60,55 @@
...
-    expect(mock_dpc_client).to receive(:get_client_tokens).and_return(mock_one_token_response).once
-    expect(mock_dpc_client).to receive(:get_public_keys).and_return({ 'count' => 2 }).once
-    expect(mock_dpc_client).to receive(:get_ip_addresses).and_return({ 'count' => 0 }).once
-    allow(Rails.logger).to receive(:info)
-    expect(Rails.logger).to receive(:info).with(['Organizations API credential status',
-                                                 { have_active_credentials: 0,
-                                                   have_incomplete_or_no_credentials: 1 }])
+      described_class.perform_now
+    end
+    it 'processes next organization successfully when an exception is raised' do
```
lukey-luke, line 89: This is the only new test.

Everything above L89 shows up in the diff because I moved the tests to the same `describe` block

dpc-portal/spec/jobs/log_organizations_api_credential_status_spec.rb
```diff
@@ -60,31 +60,55 @@
...
+      )
+      allow_any_instance_of(described_class).to receive(:fetch_credential_status).with('bar').and_raise(
+        StandardError.new('something failed inside fetch_credential_status!')
+      )
+
+      allow(Rails.logger).to receive(:info)
+      expect(Rails.logger).to receive(:info).with(['Organizations API credential status',
+                                                   { have_active_credentials: 1,
+                                                     have_incomplete_or_no_credentials: 0 }])
+      described_class.perform_now
```
jdettmannnava, line 111: Can we make sure we are logging failure?

dpc-portal/spec/jobs/log_organizations_api_credential_status_spec.rb
jdettmannnava: Hide whitespace fixes it :)

dpc-portal/spec/jobs/log_organizations_api_credential_status_spec.rb
lukey-luke: I think this is a bit beyond the scope of the original goals for building the dashboard, but it's an easy lift, so I added a new `expect()` to this test.

dpc-portal/spec/jobs/log_organizations_api_credential_status_spec.rb
lukey-luke: good to know, thank you!

PR #2334, opened 2024-11-21
Use ssl to connect to databases in lambdas

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4415

## 🛠 Changes

Conditionally force sslmode depending on environment

## ℹ️ Context
In our upgrade to postgres 16, we have to use ssl. We had ssl turned off for our lambdas so they could run against our local database (for testing), which are not set up for ssl. New code runs in sslmode except when 'testing' (which includes running locally)

## 🧪 Validation
Tested both locally (test and run-local)
Uploaded export to test and it worked.
Uploaded import to test but unsure how to test?

### DIFF
--- a/lambda/opt-out-export/db.go
+++ b/lambda/opt-out-export/db.go
@@ -11,7 +11,11 @@ import (
 var createConnection = func(dbName string, dbUser string, dbPassword string) (*sql.DB, error) {
 	var dbHost string = os.Getenv("DB_HOST")
 	var dbPort int = 5432
-	psqlInfo := fmt.Sprintf("host=%s port=%d user=%s password=%s dbname=%s sslmode=disable", dbHost, dbPort, dbUser, dbPassword, dbName)
+	var sslmode string = "require"
+	if isTesting {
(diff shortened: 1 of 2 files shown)


ashley-weaver (approved):
LGTM, though I'm not sure how to test either

ashley-weaver:
is this needed for the api-waf-sync lambda as well? IIRC it uses the same command

jdettmannnava:
> is this needed for the api-waf-sync lambda as well? IIRC it uses the same command

Probably; but does that even run on prod?

ashley-weaver:
> Probably; but does that even run on prod?

not currently, to my knowledge

NavaCharlesHorowitz:
I know this is out the door but I believe with regard to testing, you can either adjust the logging configuration on the postgresql server to log connections and their config and/or there are system tables/views that will show active connections, including whether they are using ssl, so you can do inspection of either of those things to confirm that lambdas are indeed using ssl.

PR #2338, opened 2024-11-27
Redirect to credentials tab when AO creates credential

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4375

## 🛠 Changes

- `credential_start` parameter added to show organization
- CompoundShowComponent signature simplified; credential_start parameter added
- Updates to credentials redirect with credential_start parameter

## ℹ️ Context

By default, when the AO looks at an organization, they start on the credential delegate tab. We want them to see the credentials tab if the last thing they did was create (or delete) a credential. Adding a parameter to the redirect seemed to be the easiest way to do this.

I only changed the signature of CompoundShowComponent to reduce the number of changes to the system, though now it does not match the CredentialDelegate::ListComponent. I changed the signature to address rubocop disliking signatures with too many inputs.

## 🧪 Validation

Manual and automated testing.


ashley-weaver (approved):
👍

PR #2340, opened 2024-12-02
Expired should not include accepted, cancelled, or renewed invitations

jdettmannnava:
## 🎫 Ticket

No ticket

## 🛠 Changes

Filter was added to Invitation#expired?

## ℹ️ Context

Accepted CD invitations were showing up in the expired invitations block. They shouldn't.

## 🧪 Validation

Manual and automated testing
### Before
before
### After
after


ashley-weaver (approved):
LGTM

PR #2342, opened 2024-12-02
API-WAF-SYNC use ssl in database connection if prod

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4415

## 🛠 Changes

- Configured to use ssl if not running locally.
- CGO_ENABLED set to 0 because it was failing to run
- Format of ip address updated to match aws requirements
- Database query optimized as drive-by-refactor

## ℹ️ Context

- Upgrade to postgres 16 meant we have to use a secure connection
- Bugs fixed to allow testing

## 🧪 Validation

Manually placed in AWS and it works!


lambda/api-waf-sync/db.go
```diff
@@ -36,7 +40,7 @@ var getAuthData = func(dbUser string, dbPassword string) ([]string, error) {
 	}
 	defer db.Close()
 
-	rows, queryErr := db.Query(`SELECT ip_address FROM ip_addresses`)
+	rows, queryErr := db.Query(`SELECT DISTINCT ip_address FROM ip_addresses`)
```
ashley-weaver, line 43: good catch

ashley-weaver (approved):
👍

PR #2345, opened 2024-12-04
Disable cgo in opt out lambdas because does not work

jdettmannnava:
## 🎫 Ticket

No ticket

## 🛠 Changes

Build in github actions changed to disable using CGO

## ℹ️ Context

CGO is disabled in the builds for other environments and they work. It was not enabled in dev builds, and they had 'GLIBC not found' errors when run.

## 🧪 Validation

Pushed to dev with branch github actions and export worked.


ashley-weaver (approved):
👍

PR #2348, opened 2024-12-10
[DPC-4429] Added maven-config.sh and used it in makefile

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4429

## 🛠 Changes

Updated `make maven-config` command to use new maven-config.sh script.

- All env variables should be copied from `ops/config/decrypted/local.env` to `.mvn/maven.config`.  The last one was being skipped.
- Quoted env values should be passed to maven correctly.  Instead of `-Dkey="value"` send `"-Dkey=value"`.
- Newlines should no longer be mangled in values.

## ℹ️ Context

It's been a while since most of us have had to run `make maven-config`, and in that time our encrypted environment variables file has changed and the command no longer works.  This gets everything working again and makes sure the new maven.config will run our tests.

## 🧪 Validation

Ran locally and verified the contents of maven.config.


ashley-weaver (approved):
good find

PR #2349, opened 2024-12-10
Remove deprecations in portals

jdettmannnava:
## 🎫 Ticket

No ticket

## 🛠 Changes

- Redis.current removed for mail throttling in web and admin
- enums updated
- csv gem added to web and admin Gemfiles
- rspec and rubocop configurations made quieter
- rubocop updated to new version

## ℹ️ Context
The portals contain a number of deprecated behaviors that seem worth fixing. I made some other changes just to reduce noise in the testing.

Enums were deprecated because the old format will break with next Rails upgrade.

The fix for removing Redis.current is somewhat naive; there are more complex fixes involving pooling, but instantiation is not really very expensive, and its use is for sending emails in the sandbox (which is not particularly common). So an ideal fix seems like overengineering.

I also updated rubocop, because I was getting odd errors with the older version.

## 🧪 Validation

Automated tests all pass and no deprecations in the output.


ashley-weaver (approved):
🙌

PR #2364, opened 2024-12-27
[DPC-4401] WIP: Add StatsD integration step to DPC Load Test worklow

chris-ronning-ny:
## 🎫 Ticket

### DIFF
--- a/.github/workflows/dpc-load-test.yml
+++ b/.github/workflows/dpc-load-test.yml
@@ -2,6 +2,9 @@ name: DPC Load Test
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - cr/dpc-4401
 
 jobs:
   run-load-test:
@@ -14,6 +17,16 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+
+      - name: Start StatsD Integration
+        run: |
+          docker run -d \
+            --name newrelic-statsd \
+            --network host \
+            -e NR_ACCOUNT_ID=${{ secrets.NEW_RELIC_ACCOUNT_ID }} \
+            -e NR_API_KEY=${{ secrets.NEW_RELIC_API_KEY }} \
+            -p 8125:8125/udp \
+            newrelic/nri-statsd:latest
 
       - name: Run k6 test
         env:


.github/workflows/dpc-load-test.yml
```diff
@@ -2,24 +2,52 @@ name: DPC Load Test
...
+      - name: Run K6 with StatsD output
         env:
           LOAD_TEST_CLIENT_TOKEN: ${{ secrets.LOAD_TEST_CLIENT_TOKEN }}
           LOAD_TEST_PRIVATE_KEY: ${{ secrets.LOAD_TEST_PRIVATE_KEY }}
           LOAD_TEST_PUBLIC_KEY_ID: ${{ secrets.LOAD_TEST_PUBLIC_KEY_ID }}
-        uses: grafana/run-k6-action@v1
-        with:
-          path: ./dpc-load-testing/script.js
+        run: |
+          ./k6 run \
```
chris-ronning-ny, line 42: This uses the K6 build from the previous step, which was spat out into the working directory.

.github/workflows/dpc-load-test.yml
```diff
@@ -7,19 +7,44 @@ jobs:
...
+        run: |
+          docker run -d \
+            --name newrelic-statsd \
+            -e NR_ACCOUNT_ID=${{ secrets.NEW_RELIC_ACCOUNT_ID }} \
+            -e NR_API_KEY=${{ secrets.NEW_RELIC_API_KEY }} \
+            -p 8125:8125/udp \
+            newrelic/nri-statsd:latest
+
+      # StatsD integration is deprecated in current K6 version; need to build custom K6 build,
+      # see https://grafana.com/docs/k6/latest/results-output/real-time/statsd/
```
MEspositoE14s, line 27: Reading through that issue, will we eventually have to move off StatsD for one of the other options?

MEspositoE14s (approved):
I'm far from a K6 or New Relic expert, but LGTM 👍

.github/workflows/dpc-load-test.yml
chris-ronning-ny: So the best I can understand it is: StatsD support was dropped from the official K6 release The functionality was separated into a [separate module](https://github.com/LeonAdato/xk6-output-statsd), which users need to build into a [custom K6 build](https://github.com/grafana/xk6), as I've done here. As long as that xk6-output-statsd module stays on GitHub, we should be good. Unfortunately, I wasn't able to really make heads or tails of the [Prometheus remote write](https://grafana.com/docs/k6/latest/results-output/real-time/prometheus-remote-write/), and I can't find any explanation of how to make that send its data to New Relic. 

Regardless, I think the current route will be good for a long time. Worst case scenario, we can fork the repo.

PR #2367, opened 2025-01-03
DPC-4453  - revert fetch optimization

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4453

## 🛠 Changes

Go back to fetching everything as separate step.

## ℹ️ Context

Previous fix tried to optimize fetching tags. Turns out, it doesn't work.
Runners try to do the least work possible, so it was saving previously-fetched tags. Running a different workflow forced a complete clean of the directory, so the next time this workflow was run, it didn't have any tags.

## 🧪 Validation

Successful run: https://github.com/CMSgov/dpc-app/actions/runs/12601770911/job/35123530838

### DIFF
--- a/.github/workflows/tag-for-release.yml
+++ b/.github/workflows/tag-for-release.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   tag-dpc-app:
     name: "Tag dpc-app"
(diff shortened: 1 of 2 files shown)


ashley-weaver (approved):
👍

PR #2369, opened 2025-01-07
Me/dpc 4437 add docker healthchecks

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4437

## 🛠 Changes

- Removed `wait-for-dependencies` images and replaced with docker compose health checks and wait flags.
- Updated the makefile to use the new docker compose changes.
- Added a `USE_BFD_MOCK` flag to dpc-api.
- Added _curl_ to our docker images that didn't already have it.

### Note:

Assuming you're running things through the makefile, there should be no functionality changes whatsoever.  If you call `docker compose up` manually, though, you'll have to call the services you want directly instead of through their `wait-for-dependencies` services.

## ℹ️ Context

_dadarek/wait-for-dependencies_ hasn't been updated in about seven years, and in that time Docker Compose has added better functionality natively, in the form of health checks and the `--wait` flag.

In the past, you would call _dadarek/wait-for-dependencies_ and it would start the service you wanted, then wait for it to open up a particular port indicating the service was up, then exit and return to the caller.  Now, you call `docker compose up SERVICE --wait`, and it waits for the service to start and pass its health check before returning to the caller.  This way we know the service is up and definitely ready to handle requests.  Services that depend on one another will also now wait until the services they depend on are healthy, not just running, before they start.

## 🧪 Validation

Ran all makefile targets locally.


docker-compose.override.yml
```diff
@@ -1,5 +1,3 @@
-version: '3'
```
MEspositoE14s, line 1: _version_ has been deprecated for a while now.

dpc-api/src/main/resources/application.yml
```diff
@@ -104,6 +104,7 @@ bbclient:
     requestTimeout: 5000 # ms
   serverBaseUrl: ${BFD_URL}
   resourcesCount: 100 # Max number of resource that a request to BB will return before using another request
+  useBfdMock: ${USE_BFD_MOCK:-false}
```
MEspositoE14s, line 107: Similar to above but reversed, when running in a persisted environment we should default to using the real BFD client, not the mock.

dpc-test.sh
```diff
@@ -46,19 +46,17 @@ fi
...
 
 docker compose -p start-v1-app down
 docker volume rm start-v1-app_pgdata16
-docker compose -p start-v1-app up start_core_dependencies
-docker compose -p start-v1-app up start_api_dependencies
 
 # Start the API server
-AUTH_DISABLED=true docker compose -p start-v1-app up start_api start_consent
+USE_BFD_MOCK=true docker compose -p start-v1-app up db attribution aggregation --wait
+USE_BFD_MOCK=false AUTH_DISABLED=true docker compose -p start-v1-app up api consent --wait
```
MEspositoE14s, line 59: Our end to end tests are built so that dpc-api needs to connect to the real production BFD service, but dpc-aggregation has to use our mock BFD client 🤷.  We should probably update the mock client to support what dpc-api needs and run everything that way.

src/main/resources/bb-test-data/meta.xml
```diff
@@ -0,0 +1,101 @@
+<CapabilityStatement xmlns="http://hl7.org/fhir">
```
MEspositoE14s, line 1: This file is only here so that the mockBfdClient can pass a healthcheck when run locally.

Makefile
```diff
@@ -120,31 +120,22 @@ start-load-tests: secure-envs
...
 start-dpc-debug: secure-envs
-	@mvn clean install -Pdebug -DskipTests -ntp
-	@docker compose -f docker-compose.yml -f docker-compose.portals.yml up start_core_dependencies
-	@DEBUG_MODE=true USE_BFD_MOCK=false docker compose -f docker-compose.yml -f docker-compose.portals.yml up start_api_dependencies
-	@DEBUG_MODE=true docker compose -f docker-compose.yml -f docker-compose.portals.yml up start_api
-	@docker compose -f docker-compose.yml -f docker-compose.portals.yml up start_web
-	@docker compose -f docker-compose.yml -f docker-compose.portals.yml up start_admin
-	@docker compose -f docker-compose.yml -f docker-compose.portals.yml up start_portal
+	@mvn clean install -Pci -Pdebug -DskipTests -ntp
+	@DEBUG_MODE=true docker compose -f docker-compose.yml -f docker-compose.portals.yml up aggregation api --wait
```
jdettmannnava, line 124: Unsure if you need -f docker-compose.portals.yml here?

Makefile
```diff
@@ -120,31 +120,22 @@ start-load-tests: secure-envs
...
 .PHONY: start-app-debug
 start-app-debug: secure-envs
 	@docker compose down
-	@mvn clean compile -Pdebug -DskipTests -ntp
-	@mvn package -Pci -ntp -DskipTests
-	@docker compose -f docker-compose.yml -f docker-compose.portals.yml up start_core_dependencies
-	@DEBUG_MODE=true USE_BFD_MOCK=false docker compose -f docker-compose.yml -f docker-compose.portals.yml up start_api_dependencies
-	@DEBUG_MODE=true docker compose -f docker-compose.yml -f docker-compose.portals.yml up start_api
+	@mvn clean install -Pci -Pdebug -DskipTests -ntp
+	@DEBUG_MODE=true docker compose -f docker-compose.yml -f docker-compose.portals.yml up api aggregation --wait
```
jdettmannnava, line 132: Unsure if you need -f docker-compose.portals.yml here?

jdettmannnava (commented):
It looks so pretty now! Maybe you could make the 'The "REPORT_COVERAGE" variable is not set. Defaulting to a blank string.' message go away?

Makefile
MEspositoE14s: You're right, that was a little too much copying and pasting on my part.

Makefile
MEspositoE14s: Here, too.

dpc-test.sh
MEspositoE14s: Added [DPC-4456](https://jira.cms.gov/browse/DPC-4456).

jdettmannnava (approved):
LGTM!

PR #2371, opened 2025-01-09
[DPC-4456] Update ci to use mock bfd client for all tests

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4456

## 🛠 Changes

- Update ci tests to use the mockBfdClient for dpc-api.
- Update mockBfdClient to support the synthetic patient we search for in prod BFD to load transaction time.

## ℹ️ Context

Our integration tests run as part of `make ci-app` relied on dpc-api being connected to the real production BFD system, and dpc-aggregation using our mockBfdClient.  This PR makes the whole test suite use the mockBfdClient and makes things generally more reliable.

### Note

This fixes the Jenkins build problems due to dpc-api failing its health check because it couldn't connect to BFD.

Example [here](https://management.dpc.cms.gov/job/DPC%20-%20Docker%20Build/4382/)
Passing run [here](https://management.dpc.cms.gov/job/DPC%20-%20Full%20Deploy%20and%20Release/2829/)

## 🧪 Validation

Ran locally and in Jenkins.

### DIFF
--- a/docker-compose.portals.yml
+++ b/docker-compose.portals.yml
@@ -10,7 +10,7 @@ services:
       test: ["CMD", "redis-cli", "ping"]
       interval: 5s
       timeout: 5s
-      start_period: 10s
+      start_period: 20s
       retries: 5
 
   dpc_web:
@@ -53,7 +53,7 @@ services:
       test: curl --fail http://localhost:3500/health_check || exit 1
       interval: 5s
       timeout: 5s
-      start_period: 10s
+      start_period: 20s
       retries: 5
 
   web_sidekiq:
@@ -86,7 +86,7 @@ services:
       test: curl --fail http://localhost:7433 || exit 1
       interval: 5s
       timeout: 5s
-      start_period: 10s
+      start_period: 20s
       retries: 5
 
   dpc_admin:
@@ -123,7 +123,7 @@ services:
       test: curl --fail http://localhost:3000/admin/health_check || exit 1
       interval: 5s
       timeout: 5s
-      start_period: 10s
+      start_period: 20s
       retries: 5
 
   admin_sidekiq:
@@ -155,7 +155,7 @@ services:
       test: curl --fail http://localhost:7434 || exit 1
       interval: 5s
       timeout: 5s
-      start_period: 10s
+      start_period: 20s
       retries: 5
 
   dpc_portal:
@@ -219,7 +219,7 @@ services:
       test: curl --fail http://localhost:3100/portal/health_check || exit 1
       interval: 5s
       timeout: 5s
-      start_period: 10s
+      start_period: 20s
       retries: 5
 
   portal_sidekiq:
@@ -254,5 +254,5 @@ services:
       test: curl --fail http://localhost:7435 || exit 1
       interval: 5s
       timeout: 5s
-      start_period: 10s
+      start_period: 20s
       retries: 5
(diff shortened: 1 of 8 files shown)


dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/client/MockBlueButtonClient.java
```diff
@@ -92,7 +104,7 @@ public Bundle requestCoverageFromServer(String beneId, DateRangeParam lastUpdate
     @SuppressWarnings("JdkObsolete") // Date class is used by FHIR stu3 Meta model
     public Bundle requestNextBundleFromServer(Bundle bundle, Map<String, String> headers) throws ResourceNotFoundException {
         // This is code is very specific to the bb-test-data directory and its contents
-        final var nextLink = bundle.getLink(Bundle.LINK_NEXT).getUrl();
+        final var nextLink = bundle.getLink(IBaseBundle.LINK_NEXT).getUrl();
```
MEspositoE14s, line 107: This wasn't a change I cared about, but SonarQube flagged it as a code smell so I made it.

dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/client/MockBlueButtonClient.java
```diff
@@ -142,7 +154,14 @@ private Bundle loadBundle(String pathPrefix, String beneId) {
 
         try( InputStream sampleData = loadResource(pathPrefix + beneId + fileExt) ) {
             final var bundle = parseResource(Bundle.class, sampleData, fileExt);
-            bundle.getMeta().setLastUpdated(Date.from(BFD_TRANSACTION_TIME.toInstant()));
+
+            // If this is the synthetic bene dpc-api uses to get the BFD transaction time, make sure the time returned
+            // is in the past.  If not, dpc-aggregation will fail tests when it checks transaction times.
```
jdettmannnava, line 159: As long as you are commenting, can you add a brief bit on why subtracting 300 seconds for the other transaction times make it fail?

jdettmannnava (approved):
LGTM; with one, extremely minor quibble.

docker-compose.portals.yml
```diff
@@ -8,9 +8,9 @@ services:
     image: redis:latest
     healthcheck:
       test: ["CMD", "redis-cli", "ping"]
-      interval: 5s
+      interval: 10s
       timeout: 5s
-      start_period: 10s
+      start_period: 30s
```
MEspositoE14s, line 13: Apparently things take a lot longer to start up in Jenkins than they do when running locally.

dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/client/MockBlueButtonClient.java
MEspositoE14s: I tacked on a little more info about why the synthetic patient has to show that it was loaded in the past.  300s was just an arbitrary number, though, it could've been 3,000s and would've worked fine for our purposes.  

As for why this whole process is necessary, there's a comment in the `ResourceFetcher` in dpc-aggregation that says "See BFD's RFC0004 for a discussion on why this type error may occur.", but that RFC doesn't seem to exist anymore.  Whenever we start on v2 this is probably one of the things we should ask the BFD team about.

PR #2373, opened 2025-01-10
[NO TICKET] Update new relic api key secret name

chris-ronning-ny:
## 🎫 Ticket

None

## 🛠 Changes

This changes the name of the NEW_RELIC_API_KEY secret to LOAD_TEST_NEW_RELIC_API_KEY.

## ℹ️ Context

The original name was too generic. For the Load Test use case, we specifically need an INGEST key, and as such, it is likely not useful for other use cases. By specifying the use case, others can create and add keys of their own for their own use cases.

## 🧪 Validation

- [ ] DPC Load Test GH Actions workflow succeeds when run against this branch - https://github.com/CMSgov/dpc-app/actions/runs/12716246054/job/35450142766


jdettmannnava (approved):
LGTM

PR #2393, opened 2025-01-23
DPC-4472 fix workflow names

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4472

## 🛠 Changes

Top-level names changed for multiple workflows.

## ℹ️ Context

I had left 'dev' in some workflow names, and thought standardizing naming would be useful.

## 🧪 Validation

After merge, will verify names in Actions look better.

### DIFF
--- a/.github/workflows/api-waf-sync-deploy.yml
+++ b/.github/workflows/api-waf-sync-deploy.yml
@@ -1,4 +1,4 @@
-name: api-waf-sync dev deploy
+name: API-WAF-SYNC Deploy
 
 on:
   workflow_call:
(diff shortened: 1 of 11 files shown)


chris-ronning-ny (approved):
So to check my understanding: top-level workflows take the pattern [Dpc-Service-Name Action Being Taken], and workflows that are composed into others as utilities get the pattern [utility-thing-it-does]. 

Is that understanding correct?

jdettmannnava:
@chris-ronning-ny : Yes! This way services are grouped together (so api-waf-sync deploy and integration tests are together) and the utilities are down at the bottom and out of the way.

PR #2394, opened 2025-01-23
Me/dpc 4458 bulk pat submit update

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4458

## 🛠 Changes

Bulk submits for both Patient and Practitioner resources have been upgraded:

- In dpc-api, resource validation is done in parallel instead of sequence.
- In dpc-attribution, inserts and queries are both batched before being sent to the DB.
- DB batch and query size are configurable in application.yml files.
- Customer generated ids for Patient and Practitioner resources are replaced.
- Added logging for Patient and Practitioner bundle size on bulk submission.

## ℹ️ Context

We noticed that `/Patient/$submit` was failing every time it was called by a customer, but working when called by our smoke tests.  It worked for our smoke tests because they only submit 100 patients at a time, but our customers were trying to submit more.  

After some testing, we realized that `/Patient/$submit` usually times out somewhere between 3k and 4k patients.  This new version can handle 100k in local tests.  (For reference, only one of our customers has more than 100k patients, so all but that one should now be able to submit their entire patient roster.)

### Note
The current timeout when dpc-api calls dpc-attribution is 20 seconds.  That number wasn't chosen for any customer specific reason, it was just the lowest number that worked with our smoke tests.  If we decide to raise it, at least in local tests, we should be able to handle 250k patients at a time before we start running into memory issues.

## 🧪 Validation

Ran locally and got the following results:

| Patients | Total Request Time |
|---------|---------------|
| 1000      | 2 seconds      |
| 5000      | 4 seconds      |
| 10000      | 6 seconds      |
| 50000      | 18 seconds      |
| 100000      | 30 seconds      |
| 250000      | 72 seconds (Timeout removed)      |
| 500000      | Memory error (Timeout removed)      |


dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/PatientResourceTest.java
```diff
@@ -625,6 +631,39 @@ public void testPatientPathAuthorization() throws GeneralSecurityException, IOEx
        , "Expected auth error when export another org's patient's data");
     }
 
+    @Test
+    void testBatchSubmit() throws GeneralSecurityException, IOException, URISyntaxException {
+        final int COUNT_TEST_PATIENTS = 500;
```
MEspositoE14s, line 636: If you want to test the api's performance locally without generating thousands of test patients and setting up Postman, just set the number of patients you want here and run this test.  It'll do everything for you and log the total transaction time.

jdettmannnava: Running off a local script, I am submitting 120,000...

MEspositoE14s: That's pretty similar to what I'm getting, too.  Outside of one particular customer that has ~900k patients, that should be enough for everyone else to submit their entire patient roster in one shot if they need to.

dpc-attribution/src/main/java/gov/cms/dpc/attribution/jdbi/PatientDAO.java
```diff
@@ -57,19 +62,49 @@ public List<PatientEntity> patientSearch(UUID resourceID, String patientMBI, UUI
...
         return list(query);
     }
 
+    /**
+     * Returns a list of all {@link PatientEntity}s whose id is in resourceIds.
+     * @param mbis
+     * @return List of {@link PatientEntity}s
+     */
+    public List<PatientEntity> bulkPatientSearchByMbi(UUID organizationId, List<String> mbis) {
+        List<PatientEntity> results = new ArrayList<>(mbis.size());
```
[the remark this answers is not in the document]
MEspositoE14s: That article makes sense, and the more I think about it, since this is only currently being used to search for patients we're trying to insert, maybe we should optimize for that case?  More often than not, I'd expect this query to find 0 results, because most customers aren't going to try and insert the same patient more than once, and if they do and we take a little longer it's ok because we don't have to insert those patients anyway.

PR #2398, opened 2025-01-28
DPC-4455 Smoke test from github actions

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/dpc-4455

## 🛠 Changes

- smoke-test.yml added
- Makefile adjusted for non-jenkins
- pom.xml git-commit-id plugin configuration change

## ℹ️ Context

We are migrating from Jenkins to github actions for deployment.

The smoke-test.yml workflow duplicates the [DPC - Smoke Test](https://github.com/CMSgov/dpc-ops/blob/main/jenkins_files/Jenkinsfile.smoke_test) from jenkins.

Removed jenkins-specific logic from Makefile; had to add install bzt for all.

In the pom.xml, had to disable failing if no .git directory because otherwise tests sometimes failed.

## 🧪 Validation

Successfully ran smoke test against each environment.

- [dev](https://github.com/CMSgov/dpc-app/actions/runs/13055905887)
- [test](https://github.com/CMSgov/dpc-app/actions/runs/13056125292)
- [sandbox](https://github.com/CMSgov/dpc-app/actions/runs/13056163896)
- [prod](https://github.com/CMSgov/dpc-app/actions/runs/13056167224)


.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,140 @@
...
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ inputs.env }}
+  cancel-in-progress: false
+
+jobs:
+  smoke-test:
+    name: Smoke Test
+    runs-on: self-hosted
```
ashley-weaver, line 30: should this be changed considering the TooManyRequests errors with self-hosted?

jdettmannnava: It has to be self-hosted to hit the admin endpointts

jdettmannnava: And I don't think it uses docker

MEspositoE14s (approved):
LGTM 👍

PR #2400, opened 2025-01-29
DPC-4478 self hosted ci workflow

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4478

## 🛠 Changes
- ci-workflow.yml updated to use self hosted runners
- cleanup-docker.sh added to avoid duplication
- dcp-test.sh no longer removes database volume between tests
- api_client dockerfile removed unnecessary packages

## ℹ️ Context

For security reasons, we should be using self-hosted runners instead of github runners. Moving to self-hosted runners required us to manually install python, maven, docker compose, and npm.
We also kept running out of disk space, so cleaning up docker (via the new cleanup-docker.sh) and removing unnecessary packages from the api_client were necessary.
We had to leave the database volume between tests because of some race condition making dpc-consent not start up.
There is also a drive-by fix to sonarqube, which stopped finding tests for the portals.

## 🧪 Validation

CI runs successfully

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -6,6 +6,9 @@ on:
       - .github/workflows/opt-out-*
       - lambda/**
   workflow_dispatch: # Allow manual trigger
+  push:
+    paths:
+      - .github/workflows/ci-workflow.yml
 
 env:
   VAULT_PW: ${{ secrets.VAULT_PW }}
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-workflow.yml
```diff
@@ -16,20 +16,49 @@ env:
...
           java-version: "11"
-      - name: "Set up Python 3.8.1"
-        uses: actions/setup-python@v2
-        with:
-          python-version: "3.8.1"
+      - name: "Set up Python and install Ansible"
+        run: |
+          sudo dnf -y install python3 python3-pip
+          pip install ansible
+      - name: Install Maven 3.6.3
```
MEspositoE14s, line 36: Is this the version of Maven everyone else is running locally?  I'm on 3.9.9.

dpc-test.sh
```diff
@@ -52,7 +52,6 @@ USE_BFD_MOCK=true docker compose -p start-v1-app up db attribution aggregation -
 docker compose -p start-v1-app up --exit-code-from tests tests
 
 docker compose -p start-v1-app down
-docker volume rm start-v1-app_pgdata16
```
MEspositoE14s, line 55: As long as our ITs don't leave the DB in a state that interferes with our end to end tests this should be fine, but I'd love to figure out a way to make this work again.

MEspositoE14s (approved):
Looks good to me 👍

.github/workflows/ci-workflow.yml
jdettmannnava: I am on 3.9.9 too; I think this might be what is on jenkins?

dpc-test.sh
jdettmannnava: the _finally function removes the volume at the end of the tests

.github/workflows/ci-workflow.yml
jdettmannnava: Ah, it is what is in [docker-compose](https://github.com/CMSgov/dpc-app/blob/44ee2e7b0350aacc19f3fbbb7a09fd966f82d86c/docker-compose.yml#L121)

PR #2401, opened 2025-01-29
DPC-4475 cleanup jacoco

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4475

## 🛠 Changes

dpc-test.sh removes jacocoReport directory if it is present

## ℹ️ Context

Currently, if `make ci-app` fails mid-run, JaCoCo will cause the next run to fail unpredictably, unless the generated
`./jacocoReport` directroy is manually deleted. So this automatically removes it.

## 🧪 Validation

Ran `make ci-app` and jacocoReport directory disappeared; still ran normally.


ashley-weaver (approved):
🙌

PR #2420, opened 2025-02-04
[DPC-4477] Use admin URL generate-token endpoint for auth in load test

chris-ronning-ny:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4477

## 🛠 Changes

This updates our load testing authentication to use the admin generate-token endpoint.

## ℹ️ Context

Previously, we were using the regular /Auth endpoint, which requires users to generate a JWT. That's a computationally expensive operation that caused our load tests to hang for 20+ seconds at a time. This solution avoids the need for computationally expensive operations.

## 🧪 Validation

- [ ] Load test runs successfully, and successfully fetches a bearer token using the API admin URL https://github.com/CMSgov/dpc-app/actions/runs/13162096608/job/36733157663


lukey-luke (approved):
Looks good to me!

PR #2428, opened 2025-02-06
DPC-4469 generate golden macaroon in github actions

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4469

## 🛠 Changes

Workflow for loading a golden macaroon into a given environment added.

## ℹ️ Context

We are migrating from Jenkins to github actions. This workflow duplicates [Jenkinsfile.generate_macaroon](https://github.com/CMSgov/dpc-ops/blob/main/jenkins_files/Jenkinsfile.generate_macaroon)

## 🧪 Validation

Noted dev golden macaroon in aws parameter store, called workflow with temporary caller, verified dev golden macaroon in aws parameter store had changed.


.github/workflows/generate-macaroon.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - name: AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/dpc-${{ inputs.env }}-github-actions
+      - name: Get and load Golden Macaroon
+        run: |
+          ALB_URL=$(aws elbv2 describe-load-balancers --names dpc-${{ inputs.env }}-frontend-internal | jq -r '.LoadBalancers[0].DNSName')
+          token=$(curl -X POST ${ALB_URL}:9900/tasks/generate-token)
```
chris-ronning-ny, line 40: Where does this workflow get `ALB_URL` from? I created a [new secret](https://github.com/CMSgov/dpc-app/pull/2420/files#diff-90cd840fd9e1214d3cf5f89cb8b05bdcc55a57bd4aefc487474388179f0096f3R36) for this to access this endpoint from the load test code. Maybe unnecessary?

chris-ronning-ny (approved):

.github/workflows/generate-macaroon.yml
jdettmannnava: Well, it is dynamic for every environment, so a single secret will not work.
From the line above, it is calling AWS load balancers for the frontend-internal load balancer for the given environment, then grabbing its DNSName

chris-ronning-ny: Oh I see, because the variable is set on the line above, from the result of the `aws` command. Makes sense, thank you!

PR #2429, opened 2025-02-06
[DPC-4496] use upload/download artifacts for rails docker builds

lukey-luke:
## 🎫 Ticket

[DPC-4496](https://jira.cms.gov/browse/DPC-4496)

## 🛠 Changes

 - Remove obsolete commented out code
 - remove job for building just web-portal
 - 2 new jobs
   - ea rails app will be built and then uploaded as an artifact
   - the second job will pull down that artifact and push to AWS ECR

## ℹ️ Context
 - currently experiencing issues w/ self-hosted GHA runners running out of disk space, so these jobs were separated out as opposed to a single `make ci-portals-v1` command 
 - in practice, the java apps still take a majority of the time, but this gives feedback to the developer sooner if one of the rails apps fails to build

## 🧪 Validation
 - Example CI job [here](https://github.com/CMSgov/dpc-app/actions/runs/13185239034)

### DIFF
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -42,14 +42,13 @@ jobs:
         run: |
           make ci-portal
 
-#        run: docker save dpc-$REPOSITORY:latest | gzip > ${{ runner.temp }}/dpc_$REPOSITORY_latest.tar.gz
       - name: gzip the image
         run: docker save dpc-$REPOSITORY:latest | gzip > ${{ runner.temp }}/dpc_web-portal_latest.tar.gz
       - name: upload tar artifact
         uses: actions/upload-artifact@v4
         with:
           name: dpc-$REPOSITORY
-          path: ${{ runner.temp }}/dpc_web-portal_latest.tar.gz
+          path: ${{ runner.temp }}/dpc_$REPOSITORY_latest.tar.gz.tar
 
   docker_push_rails_web_portal_image:
     runs-on: self-hosted
@@ -62,18 +61,6 @@ jobs:
         with:
           name: dpc-$REPOSITORY
           path: ${{ runner.temp }}
-
-      - name: Load image
-        run: |
-          docker load --input ${{ runner.temp }}/myimage.tar
-          docker image ls -a
-      - name: gzip the image
-        run: docker save dpc-$REPOSITORY:latest | gzip > $REPOSITORY_latest.tar.gz
-      - name: upload tar artifact
-        uses: actions/upload-artifact@v4
-        with:
-          name: dpc-$REPOSITORY
-          path: ${{ runner.temp }}/$REPOSITORY_latest.tar.gz.tar
 
       - name: Load image
         run: |
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -18,15 +18,15 @@ jobs:
     runs-on: self-hosted
     strategy:
       matrix:
-        repository: [ web-portal, web-admin, web ]
+        ecr_repository: [ web-portal, web-admin, web ]
         include:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker_build.yml
```diff
@@ -31,107 +44,54 @@ jobs:
           sudo chown root:root /usr/local/lib/docker/cli-plugins/docker-compose
           sudo chmod +x /usr/local/lib/docker/cli-plugins/docker-compose
 
-      # TODO replace w/ make ci-portals-v1 when disk space issue resolved
-      - name: Build portal
+      - name: Build specified app
+        run: $MAKE_COMMAND
```
jdettmannnava, line 48: why not `${{ matrix.make_command }}`?

.github/workflows/docker_build.yml
```diff
@@ -15,8 +14,22 @@ env:
   ENV: "github-ci"
 
 jobs:
-  docker_build_rails_web_portal:
+  docker_build_rails_apps:
     runs-on: self-hosted
+    strategy:
+      matrix:
+        repository: [ web-portal, web-admin, web ]
```
jdettmannnava, line 21: I don't really like 'repository' as a name, especially because it has meaning in git...

.github/workflows/docker_build.yml
```diff
@@ -31,107 +44,54 @@ jobs:
...
+      - name: Build specified app
+        run: $MAKE_COMMAND
+
+      - name: gzip the image
+        run: docker save dpc-${{ matrix.repository }}:latest | gzip > ${{ runner.temp }}/dpc_${{ matrix.repository }}_latest.tar.gz
+      - name: upload tar artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: dpc-${{ matrix.repository }}
+          path: ${{ runner.temp }}/dpc_${{ matrix.repository }}_latest.tar.gz
```
jdettmannnava, line 56: Probably want to set [the retention period](https://github.com/marketplace/actions/upload-a-build-artifact#retention-period) to the minimum

jdettmannnava (commented):
On to image tag!

.github/workflows/docker_build.yml
```diff
@@ -31,107 +44,54 @@ jobs:
...
+
+  docker_push_rails_apps:
```
jdettmannnava, line 58: Shouldn't all apps be pushed together?

.github/workflows/docker_build.yml
lukey-luke: No real motivation for either. Both would achieve the same thing in this scenario.
Removing Lines 33-34 and using `${{ matrix.make_command }}` would be less lines of code, so I'll go for that one

.github/workflows/docker_build.yml
lukey-luke: Is there a reason to stray away from the default 90 days?

For a given build, we might want to access the images at a later date.

lukey-luke:
Switched to ready-for-review. It appears disk space limitations have been resolved in [PLT-897](https://jira.cms.gov/browse/PLT-897).

.github/workflows/docker_build.yml
jdettmannnava: These are pretty big objects, and I cannot think of a single reason to keep them in github, as the whole point is to push them to AWS?

lukey-luke: ohhhh, you're right, they don't need to live on github and this wouldn't impact how long builds are available on ECR. Updated!

.github/workflows/docker_build.yml
lukey-luke: Yeah, breaking up the java build to use upload/download artifacts and then having a single job waiting for all should achieve this. I'll work on a follow-up PR for this

jdettmannnava (approved):
Fine for WIP

PR #2436, opened 2025-02-11
Me/dpc 4501 remove get total

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4501

## 🛠 Changes

- No longer relies on `Bundle.getTotal()` to determine if calls to BFD were successful.
- Removed the `total` element from our test bundles.

## ℹ️ Context

The `total` element isn't required in a bundle, so we shouldn't be relying on it, and the BFD team is considering a change that will remove it in the future so this makes sure we won't have any issues if they do.

Note: The large number of deleted test files were from DPC v2, which was removed a long time ago.

## 🧪 Validation

Removed total from all test data and made sure tests pass locally.


docker-compose.base.yml
```diff
@@ -1,5 +1,3 @@
-version: '3'
```
MEspositoE14s, line 1: Not really necessary, but I was getting a warning that this was deprecated while rebuilding so I got rid of it.

chris-ronning-ny (approved):
Looks good to me. I can't say I understand all the .xml changes in bb-test-data, but if it's not breaking test, I imagine we're okay. Can you maybe speak to that a bit?

MEspositoE14s:
> Looks good to me. I can't say I understand all the .xml changes in bb-test-data, but if it's not breaking test, I imagine we're okay. Can you maybe speak to that a bit?

I removed the total element from each FHIR bundle so that they match what we'll be getting from BFD in the future.  It's the same change in the .json files, too, but they're a lot easier to read.

PR #2439, opened 2025-02-12
Me/dpc 4497 remove deprecated performance tests and client

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4497

## 🛠 Changes

- Removed deprecated performance texts in `dpc-testing/performance`
- Removed dpcclient.

## ℹ️ Context

Both of these still work, but neither of them are maintained and haven't been used in 5 years since we have better options now.  A new suite of performance tests is currently being built and the dpcclient doesn't really do anything that can't be done with a few shell commands.

## 🧪 Validation

Nothing really to validate.


chris-ronning-ny (approved):
✂️

PR #2472, opened 2025-02-21
Update contact_telecoms column name

ashley-weaver:
## 🎫 Ticket

No ticket.

## 🛠 Changes

Updates column name.

## ℹ️ Context

Before upgrading to Java 17, we need to update this column name, as "value" is a protected term.

## 🧪 Validation

Tests passing, including smoke.

### DIFF
--- a/dpc-attribution/src/main/resources/migrations.xml
+++ b/dpc-attribution/src/main/resources/migrations.xml
@@ -436,6 +436,13 @@
     </changeSet>
 
     <changeSet id="increment-attributions_id_seq" author="MEsposito">
-        <alterSequence sequenceName="attributions_id_seq" incrementBy="100"></alterSequence>
+        <alterSequence sequenceName="attributions_id_seq" incrementBy="100"/>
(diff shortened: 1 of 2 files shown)


dpc-attribution/src/main/resources/migrations.xml
```diff
@@ -436,6 +436,13 @@
     </changeSet>
 
     <changeSet id="increment-attributions_id_seq" author="MEsposito">
-        <alterSequence sequenceName="attributions_id_seq" incrementBy="100"></alterSequence>
+        <alterSequence sequenceName="attributions_id_seq" incrementBy="100"/>
```
jdettmannnava, line 439: This change is probably unnecessary...

ashley-weaver: I know, IntelliJ just really wanted me to include it. I can revert it.

jdettmannnava (approved):
LGTM!

PR #2478, opened 2025-02-25
DPC-4542 Fix Concurrency

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4542

## 🛠 Changes

workflow id replaced with name of workflow.

## ℹ️ Context

We were getting blocking collisions when workflows called other workflows.

## 🧪 Validation

Promote passed.

### DIFF
--- a/.github/workflows/ecs-deploy.yml
+++ b/.github/workflows/ecs-deploy.yml
@@ -57,7 +57,7 @@ on:
         default: 'main'
 
 concurrency:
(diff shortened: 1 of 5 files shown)


lukey-luke (approved):
Should unblock smoke test from failing to run during promote build because both would use the same concurrency block when called from same workflow.

Example job:
https://github.com/CMSgov/dpc-app/actions/runs/13529826965

LGTM

PR #2480, opened 2025-02-26
DPC4473 run opt out tests manually

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4473

## 🛠 Changes

Opt out integration tests configured to run manually
Removed unnecessary checkouts
Moved opt out integration tests to self-hosted runners

## ℹ️ Context

Because the top-level job does not run when triggered manually, all other jobs with any needs require an "always" trigger.
Moved to self-hosted due to policy.
Other changes just drive-by cleanup

## 🧪 Validation

Tests run and pass when run manually:
- [Opt out import](https://github.com/CMSgov/dpc-app/actions/runs/13549970234)
- [Opt out export](https://github.com/CMSgov/dpc-app/actions/runs/13549975831)


ashley-weaver (approved):
👍

PR #2485, opened 2025-02-28
Me/dpc 4545 fix aggregation

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4545

## 🛠 Changes

- Updated `completePartialBatch` so that it updates the job_queue_batch table after each patient is processed.
- Temporarily Increased the job timeout to 15 minutes.

## ℹ️ Context

We noticed that jobs were getting stuck in the queue because the update_time in the job_queue_batch table wasn't getting set frequently enough.  Sometimes it's because of issues in our DB code and sometimes it's because patients are taking a long time to process.  This should clean up both of those.

## 🧪 Validation

- Added a test to ensure update_time is set after each patient is processed.
- Deployed to test env and watched the DB as the smoke tests ran.

I verified that the code does what I expected it to, but the only way to know for sure if this fixes our problem is to deploy it to prod and periodically check the logs to see if jobs are still getting stuck.


dpc-queue/src/main/java/gov/cms/dpc/queue/DistributedBatchQueue.java
```diff
@@ -174,7 +174,7 @@ public Optional<JobQueueBatch> claimBatch(UUID aggregatorID) {
     @SuppressWarnings("unchecked")
     private void restartStuckBatches(Session session) {
         // Find stuck batches
-        List<String> stuckBatchIDs = session.createNativeQuery("SELECT Cast(batch_id as varchar) batch_id FROM job_queue_batch WHERE status = 1 AND update_time < current_timestamp - interval '5 minutes' FOR UPDATE SKIP LOCKED")
+        List<String> stuckBatchIDs = session.createNativeQuery("SELECT Cast(batch_id as varchar) batch_id FROM job_queue_batch WHERE status = 1 AND update_time < current_timestamp - interval '15 minutes' FOR UPDATE SKIP LOCKED")
```
MEspositoE14s, line 177: We have [DPC-4548](https://jira.cms.gov/browse/DPC-4548) to figure out what time this should be set to, but I bumped it up in the meantime to prevent jobs from getting stuck until that ticket gets completed.

ashley-weaver (approved):
🙌

PR #2505, opened 2025-03-07
Me/dpc 4549 update logging

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4549

## 🛠 Changes

- Added logging when we fetch resources.  Anything under 30 seconds is logged at INFO, over at WARN.  
- Made the logging threshold configurable in dpc-aggregation's application.yml file.
- Added aggregator_id and patient_fhir_id to the MDC so it shows up in logs.
- Logged extra messaging for failed jobs.

## ℹ️ Context

We recently had a number of batches that took way too long to complete and haven't been able to definitively pinpoint why.  These changes should help narrow it down, specifically if we're having trouble loading resources from BFD.

Our default logging level is set to WARN, and any resource fetch that runs under 30 seconds is logged at INFO so they won't be clogging up our logs.  Fetches that take longer than 30 seconds are logged at WARN so they will show up in our logs.

I considered setting up something like this in the BFD client itself, but it already has a 5 second timeout so no queries can run longer than that.

## 🧪 Validation

Deployed to test and verified that the new logs are getting written and parsed correctly in Splunk:
image


dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/JobBatchProcessor.java
```diff
@@ -247,7 +248,9 @@ private Optional<Patient> fetchPatient(JobQueueBatch job, String mbi) {
 
         // If we get more than one unique Patient for an MBI then we've got some upstream problems.
         if (patients.getEntry().size() == 1) {
-            return Optional.of((Patient) patients.getEntryFirstRep().getResource());
+            Patient patient = (Patient) patients.getEntryFirstRep().getResource();
+            MDC.put(MDCConstants.PATIENT_FHIR_ID, patient.getIdPart());
```
jdettmannnava, line 252: Are we sure we want this in the log? Isn't this protected information?

MEspositoE14s: This is the id of the FHIR Patient resource itself as opposed to something like the MBI that identifies the patient directly.  It's basically the primary key of the patient table.  I've never considered it protected, but it's probably not a bad idea to ask someone from security if it's an issue.

jdettmannnava: Hrrm, well then, I think calling it fhirId is super confusing...

jdettmannnava (approved):
LGTM

PR #2506, opened 2025-03-10
[DPC-4441] Add URI to logs where we log gzip

lukey-luke:
## 🎫 Ticket

[DPC-4441](https://jira.cms.gov/browse/DPC-4441)

## 🛠 Changes

Add additional logging to include `uri` where we log gzip headers.

## ℹ️ Context

 - currently there isn't a way to view which types of requests are gzipped
 - as a maintainer of the system, I want to know if the bulkiest endpoints containing more data get gzipped


lukey-luke:
Now that greenfield work is done, environment for non-migration work is available. I was able to test this in the greenfield dev environment and confirm that URI shows up in the logs as expected
Screenshot 2025-06-17 at 11 24 34 AM

PR #2514, opened 2025-03-13
Me/dpc 2832 patient everything error fix

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-2832

## 🛠 Changes

- When a patient fails look back, dpc-aggregation returns a more descriptive `OperationOutcome` to dpc-api.
- `OperationOutcome` issue type is set to SURPRESSED.
- `OperationOutcome` location is set to the patient's fhir resource id instead of their MBI.
- On a `patient/$everything` failed look back, dpc-api throws a `ForbiddenOperationException`, which returns a 403 and the `OperationOutcome` from dpc-aggregation to the caller.

## ℹ️ Context

When calling `patient/$everything`, if the patient failed its look back check dpc-api would return a generic internal server error.  Customers were complaining and wanted something more specific.

## 🧪 Validation

Ran locally and added integration tests to check functionality.

### DIFF
--- a/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/JobBatchProcessor.java
+++ b/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/JobBatchProcessor.java
@@ -187,7 +187,7 @@ public class JobBatchProcessor {
                     Flowable.just(
                         AggregationUtils.toOperationOutcome(
                             failReason,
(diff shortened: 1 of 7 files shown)


chris-ronning-ny (approved):
Code looks good, tests look good. I questioned whether `SUPPRESSED` is the correct issue type, and looking at some documentation on [FHIR issue types](https://build.fhir.org/codesystem-issue-type.html), it seems like you landed on the right one.

PR #2522, opened 2025-03-18
Me/dpc 4552 add load test workflow

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4552

## 🛠 Changes

- Successfully makes each of the calls listed in the Jira ticket in workflow A.
- Creates new resource bodies for Provenance and Group.
- Streamlines header creation for requests.
- Removed token code that's no longer necessary with auth disabled in dpc-api.

## ℹ️ Context

Implementing these requests is the next step moving towards full load testing.

## 🧪 Validation

Ran locally and ensured all requests were successful.


Makefile
```diff
@@ -86,7 +86,7 @@ start-api: start-app
 
 start-api-load-tests: ## Start a new API env for load tests
 start-api-load-tests: secure-envs
-	USE_BFD_MOCK=true AUTH_DISABLED=true docker compose -p dpc-load-tests up api --wait
+	USE_BFD_MOCK=true AUTH_DISABLED=true docker compose -p dpc-load-tests up api aggregation --wait
```
MEspositoE14s, line 89: Added aggregation here because that's where the queue migrations are done, and we need that for Group/$export to succeed.

dpc-load-testing/script.js
```diff
@@ -1,7 +1,9 @@
-import { check, fail } from 'k6';
+import { check, fail, group } from 'k6';
 import exec from 'k6/execution'
-import tokenCache, { generateDPCToken, fetchGoldenMacaroon } from './generate-dpc-token.js';
-import { createOrganization, deleteOrganization, getOrganization } from './dpc-api-client.js';
+import { 
+    createGroup, createOrganization, createPatient, createProvider, deleteOrganization, exportGroup, getGroup, getOrganization,
```
chris-ronning-ny, line 4: Just a nit; when imports in a block get too long to reasonably declare on one line, the standard I usually see is to give each named import its own line. 

e.g.
```
import {
  createGroup,
  createOrganization,
  createPatient,
  ...etc
}
```

Might be reasonable to set up a linter for this project in the future.

dpc-load-testing/script.js
```diff
@@ -62,34 +63,67 @@ export function setup() {
...
-  const checkOutput = check(
-    orgResponse, 
-    { 'response code was 200': res => res.status === 200 }
-  )
+  // POST practitioner
+  const practitionerResponse = createProvider("1232131239", orgId);
+  if (practitionerResponse.status != 201) {
+    fail('failed to create practitioner for workflow A');
+  }
+  // There's only 1 identifier in our synthetic practitioner, so we don't have to search for npi
```
chris-ronning-ny, line 72: I'm assuming this will still apply once we're creating x number of practitioners, right?

dpc-load-testing/script.js
```diff
@@ -62,34 +63,67 @@ export function setup() {
...
+  const practitionerNpi = practitionerResponse.json().identifier[0].value;
+  const practitionerId = practitionerResponse.json().id;
 
-  if (!checkOutput) {
-    fail('Failed to get a 200 response in workflow A');
+  // POST patient
+  const patientResponse = createPatient("1S00EU8FE91", orgId);
+  if (patientResponse.status != 201) {
+    fail('failed to create patient for workflow A');
```
chris-ronning-ny, line 79: I'm wondering how this is going to scale when we're running the actual test, if a given request fails some larger number of times. 

Just a note for myself to look into ways to capture this for analytics purposes.

dpc-load-testing/script.js
MEspositoE14s: I let the IDE fill that in on its own, but I like the way you're suggesting better, too.

dpc-load-testing/script.js
MEspositoE14s: Yeah, it'll only change if we add more identifiers in `generateProviderResourceBody`

dpc-load-testing/script.js
MEspositoE14s: We've definitely seen dpc-attribution get overwhelmed with a few really big requests, and if that happens these would probably start timing out.  

I wonder if we could export the results, or output formatted logs to Splunk or something like that?

chris-ronning-ny: I imagine it's easy enough to send number of fails over to the New Relic dashboard we've got running.

chris-ronning-ny (approved):
LGTM!

PR #2523, opened 2025-03-19
DPC_4565 Run load tests with authorization on

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4565

## 🛠 Changes

- Authorization turned on when running load tests
- Utility for creating org-limited token based on golden macaroon added
- Test script updated to use token in calls
- Github action updated to use golden macaroon from AWS Parameter store
- Added unit test for Macaroon
- Add Github action that runs unit test for Macaroon

## ℹ️ Context

We were going to get scripts running with auth disabled, but this is a way around that.

## 🧪 Validation

Tests pass against authenticated service.
Successful run in github actions: https://github.com/CMSgov/dpc-app/actions/runs/13971836770/job/39115385958
Successful unit test run in github actions: https://github.com/CMSgov/dpc-app/actions/runs/13994145683


jdettmannnava:
Should I add sources of inspiration to the comments? Like that Macaroon is basically a port from ruby-macaroons?

Makefile
```diff
@@ -86,7 +86,7 @@ start-api: start-app
 
 start-api-load-tests: ## Start a new API env for load tests
 start-api-load-tests: secure-envs
-	USE_BFD_MOCK=true AUTH_DISABLED=true docker compose -p dpc-load-tests up api aggregation --wait
+	USE_BFD_MOCK=true AUTH_DISABLED=false docker compose -p dpc-load-tests up api --wait
```
chris-ronning-ny, line 89: I'm pretty sure AUTH_DISABLED is false by default. You shouldn't need to specify that here.

dpc-load-testing/generate-dpc-token.js
```diff
@@ -0,0 +1,121 @@
...
+  deserialize(buf) {
+    let index = 0;
+    while (index < buf.byteLength) {
+      const packetLength = new Uint8Array(buf, index, PACKET_PREFIX_LENGTH);
+      const len = parseInt(arrayBuffer2String(packetLength), 16);
+      const chunk = new Uint8Array(buf, index, len);
+      const packet = new Uint8Array(buf, index + PACKET_PREFIX_LENGTH, len - (PACKET_PREFIX_LENGTH + 1));
+      const keyValue = arrayBuffer2String(packet).split(" ");
+      if (keyValue[0] == 'location') {
+	this.location = keyValue[1];
```
chris-ronning-ny, line 51: Indentation from here to line 58 is awkward, please fix

dpc-load-testing/generate-dpc-token.js
```diff
@@ -0,0 +1,121 @@
...
+  return http.post(fetchGoldenMacaroonURL, {}, { headers: headers }).body
+}
+
+export function generateDPCToken(orgId, goldenMacaroon) {
+  const rawGoldenMacaroon = encoding.b64decode(goldenMacaroon, 'rawurl');
+  const macaroon = new Macaroon();
+  macaroon.deserialize(rawGoldenMacaroon);
+  macaroon.addFirstPartyCaveat('dpc_macaroon_version = 1');
+  macaroon.addFirstPartyCaveat(`organization_id = ${orgId}`);
+  macaroon.addFirstPartyCaveat(`expires = ${new Date(new Date().getTime() + 5*60000).toISOString()}`);
```
MEspositoE14s, line 27: Just to make sure I fully understand, we're skipping building a JWT and asking dpc-api to give us a client token, and instead using the golden macaroon to create it client side instead?  If that's the case, is there any reason we couldn't up the expiration date if we wanted to run a workflow that takes longer than 5 minutes?  Server side, I think this only gets enforced on creation.

MEspositoE14s (approved):
Looks good to me, and all the tests pass locally so 👍

dpc-load-testing/generate-dpc-token.js
jdettmannnava: Full understanding is an eldritch horror. We are making a bearer token (which are always macaroons) that does not have the third-party caveats baked into the client tokens. The third-party caveats in the client tokens prevent the use of the client tokens as bearer tokens. The jwt -> auth process basically adds a discharger that allows verification of the third-party caveat, which thus allows use of the bearer token for authorization.
The actual logic is based on the dpc_client used in the Rails apps for interacting with the API without going through the jwt -> token process

jdettmannnava: Yes, we can set the timeout to whatever we want! Five minutes was just arbitrary.
When we want more time, we can probably add the minutes as an input parameter, but I chose a magic number over overengineering.

.github/workflows/load-test-unit-tests.yml
```diff
@@ -0,0 +1,47 @@
...
+        working-directory: ./dpc-load-testing
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+      - name: Set up tests for dpc-load-testing utilities
+        run: yarn
+      - name: Run unit tests
+        run: yarn test
```
chris-ronning-ny, line 25: I'm surprised this step doesn't fail, given that this branch doesn't have a `test` script in package.json.

chris-ronning-ny: Oh, I see that you skipped that step in the workflow. Perfect, thank you.

chris-ronning-ny (approved):
Nice job with this Jeff!

PR #2525, opened 2025-03-20
[DPC-4551] Generate fake NPIs and MBIs

chris-ronning-ny:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4551

## 🛠 Changes

- New code generates valid NPIs and MBIs, for use in load testing. 
- Runs unit tests for these utilities in GH Actions

Info on NPI validation: https://www.eclaims.com/articles/how-to-calculate-the-npi-check-digit/
Info on MBI format: https://www.cms.gov/medicare/new-medicare-card/understanding-the-mbi-with-format.pdf

## ℹ️ Context

When we're running load tests, we'll need to generate an arbitrary number of valid identifiers when we create Providers and Patients. 

## 🧪 Validation

- [x] Unit tests pass in CI - https://github.com/CMSgov/dpc-app/actions/runs/14038652530/job/39302972844?pr=2525
- [x] Unit tests run quickly -- 10000 NPIs and 10000 MBIs are generated during unit tests, and the code runtime is 0.633 seconds.

### DIFF
--- a/.github/workflows/load-test-unit-tests.yml
+++ b/.github/workflows/load-test-unit-tests.yml
@@ -1,0 +1,22 @@
+name: Load Testing Unit Tests
+
+on:
+  pull_request:
+    paths:
+      - .github/workflows/load-test-unit-tests.yml
+      - dpc-load-testing/**
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./dpc-load-testing
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+      - name: Run unit tests for dpc-load-testing utilities
+        run: yarn test
--- a/.github/workflows/load-test-unit-tests.yml
+++ b/.github/workflows/load-test-unit-tests.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
     defaults:
       run:
         working-directory: ./dpc-load-testing
--- a/.github/workflows/load-test-unit-tests.yml
+++ b/.github/workflows/load-test-unit-tests.yml
@@ -18,6 +18,8 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 22
+      - name: Install yarn
+        run: npm install -g yarn
       - name: Set up tests for dpc-load-testing utilities
         run: yarn
       - name: Run unit tests
--- a/.github/workflows/load-test-unit-tests.yml
+++ b/.github/workflows/load-test-unit-tests.yml
@@ -19,4 +19,5 @@ jobs:
         with:
           node-version: 22
       - name: Run unit tests for dpc-load-testing utilities
-        run: yarn test
+      - run: yarn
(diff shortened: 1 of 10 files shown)


dpc-load-testing/package.json
```diff
@@ -7,6 +7,10 @@
   "license": "MIT",
   "type": "module",
   "devDependencies": {
-    "@types/k6": "^0.54.2"
+    "@types/k6": "^0.54.2",
+    "jest": "^29.7.0"
+  },
+  "scripts": {
+    "test": "NODE_OPTIONS='--experimental-vm-modules --experimental-specifier-resolution=node' jest"
```
chris-ronning-ny, line 14: If we don't do this, we'll need to set up babel and transpile it to enable ES imports for unit tests. That can be a pretty arduous task, and we're only going to ever unit test a small number of utilites. 

This is the better way.

.github/workflows/load-test-unit-tests.yml
```diff
@@ -0,0 +1,24 @@
...
+on:
+  pull_request:
+    paths:
+      - .github/workflows/load-test-unit-tests.yml
+      - dpc-load-testing/**
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
jdettmannnava, line 12: If it can't run on self-hosted, does @CMSgov/ab2d-bcda-dpc-platform know?

dpc-load-testing/utils/mbi-generator.js
```diff
@@ -0,0 +1,44 @@
+const alpha = 'ACDEFGHJKMNPQRTUVWXY';
```
jdettmannnava, line 1: I am wondering if we should limit this to letters that are never in real mbis (S, L, O, I, B, and Z)? Or maybe just for line 9?

I had to do https://jira.cms.gov/browse/DPC-3986 to make sure real MBIs weren't used in non-production situations.

jdettmannnava (commented):
Any notes in a README and/or a make to run the tests locally?

.github/workflows/load-test-unit-tests.yml
chris-ronning-ny: I don't know that it has to run on self-hosted, given that this isn't part of our deployment process. I didn't try to run it there, but I'll give it a go for completeness sake.

dpc-load-testing/utils/mbi-generator.js
chris-ronning-ny: I was a bit concerned that this would not be accepted by our API, but I tested locally and the API requests are made successfully. I'm not really sure what the rationale is, given that we're going to wipe the database, but I'll take the ticket you link as evidence that this is probably a requirement. It won't hurt our testing at all.

jdettmannnava: I meant this should go the other way, so that alpha should be 'SLOIBZ'. I think this way has only real MBIs?

chris-ronning-ny: I hadn't pushed my commit when I made that comment, but now I have.

chris-ronning-ny:
@jdettmannnava This should be all set, aside from the nokogiri dependency addressed in the other PR.

jdettmannnava (approved):
LGTM

PR #2534, opened 2025-03-24
Fix accessibility test issues

jdettmannnava:
## 🎫 Ticket
No ticket

## 🛠 Changes

- Accessibility test entrypoint adds icu-data-full before adding firefox
- Tests updated

## ℹ️ Context

For some reason, an update to the the ruby docker image made it so the icu-data-full needs to be explicitly added before firefox before it will install.
We also updated the invitation flow, which required a fix to the tests

## 🧪 Validation

Accessibility tests run locally and on GHA with only accessibility failures.

### DIFF
--- a/dpc-portal/docker/accessibility-test.sh
+++ b/dpc-portal/docker/accessibility-test.sh
@@ -1,4 +1,6 @@
 #!/usr/bin/env sh
 
+apk update
+apk add --no-cache icu-data-full
 apk add --no-cache firefox
 ACCESSIBILITY=true bundle exec rspec --tag type:system --format documentation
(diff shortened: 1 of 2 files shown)


MEspositoE14s (approved):
The tests run for me locally, but I'm getting a bunch of "moderate" and "minor" violations.  I'm assuming we really only care about anything marked "high"?

jdettmannnava:
> The tests run for me locally, but I'm getting a bunch of "moderate" and "minor" violations. I'm assuming we really only care about anything marked "high"?

The tests weren't running at all before (https://github.com/CMSgov/dpc-app/actions/runs/14042006372), but they have always been failing. The important thing is that the failures are accessibility failures. Before I updated the test, it was failing that it was not the correct page.

PR #2541, opened 2025-03-25
DPC-4534 run CI on self-hosted runners

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4534

## 🛠 Changes
- ci-workflow.yml updated to use self-hosted runners (along with necessary changes: cf docker-build.yml)
- Docker image source changed from docker hub (default) to artifactory

## ℹ️ Context

Because of SLA agreements, we should be using self-hosted runners.
Self-hosted runners seem more likely to have issues with leftover docker detritus, so we need to blow away the entire docker setup as cleanup. This means we download docker images more often. Docker hub has a limit (which we have run into), so we need to download our base images from our artifactory instance instead.

## 🧪 Validation

CI workflow passes in a reasonable amount of time.


MEspositoE14s (approved):
LGTM! 👍

PR #2554, opened 2025-04-02
[DPC-4602] Remove patient.gender from mock bfd data

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4602

## 🛠 Changes

Removed `patient.gender` from `MockBlueButtonClient` data.
Add us-core-sex extension to `MockBlueButtonClient` data.

## ℹ️ Context

On 4/30, BFD will no longer be populating `patient.gender` and will instead be sending a us-core-sex extension.  We want to make sure our test data reflects this and dpc-aggregation doesn't have any problems with it.

## 🧪 Validation

All unit and integration tests are passing.


jdettmannnava (commented):
Also dpc-api/src/test/resources/f1e518f5-4977-47c6-971b-7eeaf1b433e8-0.patient.ndjson ?
It is unnecessary for the test (JobResourceTest), but it technically addresses "BFD not sending gender"

MEspositoE14s:
> Also dpc-api/src/test/resources/f1e518f5-4977-47c6-971b-7eeaf1b433e8-0.patient.ndjson ? It is unnecessary for the test (JobResourceTest), but it technically addresses "BFD not sending gender"

That doesn't get processed through the mock blue button client, but you're right that if we want to be as technically correct as possible it should probably get removed.  And technically correct is the best kind of correct!

jdettmannnava (approved):
LGTM!
Weird that JobResourceTest just sets the values it wants to see...

PR #2565, opened 2025-04-06
[DPC-4567] Update my organizations list page

lukey-luke:
## 🎫 Ticket

[DPC-4567](https://jira.cms.gov/browse/DPC-4567)

## 🛠 Changes

 - updates existing Organization List page to use the list format instead of cards
 - see figma design spec [HERE](https://www.figma.com/design/cvJ8ru6pzsVTS5zAmMwIAC/DPC-Production-Portal?node-id=3-17042&p=f&m=dev)

## 🧪 Validation

Manual testing...

AO and CD links show up in the same list
Screenshot 2025-04-03 at 5 43 49 PM

Tested "Sign terms of service", "Configuration needed", and "Configuration complete" status icons manually


dpc-portal/app/assets/stylesheets/_organizations-list.scss
```diff
@@ -0,0 +1,49 @@
+$link-color: rgb(0, 0, 238);
```
jdettmannnava, line 1: We use the uswds design system usually; could you not find components there to use?

lukey-luke: I updated `$link-color` to `#005ea2` in order to match the `.usa-link` class.


Wouldn't we want to keep the UX consistent across visits? I.e. stay blue color even after clicking into a particular organization?

Default behavior for links is to change color when they're visited. If it was a random link at the bottom of the site, I'd reckon we'd want it to show purple after visiting.

dpc-portal/app/components/core/card/organization_card_component.html.erb
```diff
@@ -1,28 +1,21 @@
-<li class="usa-card tablet-lg:grid-col-1 widescreen:grid-col-1" style="list-style:none; visibility:visible;" data-npi="<%= @organization.npi%>" data-name="<%= @organization.name%>">
-  <div class="usa-card__container">
-    <div class="usa-card__header">
-      <h2 class="usa-card__heading">
-        <%= @organization.name %>
-      </h2>
+<tr class="organizations-list-row">
```
jdettmannnava, line 1: This should probably not be called card component

dpc-portal/app/assets/stylesheets/_organizations-list.scss
jdettmannnava: Oh, I meant why use custom css at all? I am pretty sure everything we want to do is in the system; Clara is pretty good about that.

dpc-portal/app/assets/stylesheets/application.sass.scss
```diff
@@ -17,3 +17,4 @@
  */
 @forward "uswds-theme";
 @forward "uswds";
+@import "organizations-list";
```
jdettmannnava, line 20: Unsure why you are importing this instead of requiring it like login and invitation?

dpc-portal/app/assets/stylesheets/_organizations-list.scss
```diff
@@ -0,0 +1,21 @@
+$visited-blue-link-color: #005ea2;
```
jdettmannnava, line 1: Unsure why this file begins with underscore?

dpc-portal/spec/components/page/organization/organization_list_component_spec.rb
```diff
@@ -49,47 +50,36 @@ def normalize_space(str)
...
+                             <a class="display-block maxw-full visited:text-blue text-underline truncate-text-ellipsis" href="/portal/organizations/5">
+                               name
+                             </a>
+                           </th>
+                         <td style="width: 16.67%;"> npi </td>
+                         <td style="width: 8.33%;">
+                             CD
+                         </td>
+                         <td style="width: 25%;">
+                             <div class="clearfix">
```
jdettmannnava, line 74: This doesn't really test the 'one row' concept as it truncates early

jdettmannnava (commented):
Looks good! But I still have some questions.
Also, if you could update the [lookbook preview](https://github.com/CMSgov/dpc-app/blob/main/dpc-portal/app/components/page/organization/organization_list_component_preview.rb) for more than one status, that would be nice!

dpc-portal/app/assets/stylesheets/_organizations-list.scss
lukey-luke: Reduced amount of custom css. Used more generic CSS classes, but retained some new css in order to achieve requirements. Namely table background color and ellipsis for text overflow in the table

dpc-portal/spec/components/page/organization/organization_list_component_spec.rb
lukey-luke: Added missing HTML and verified test renders the configuration needed status!

dpc-portal/config/locales/en.yml
```diff
@@ -36,13 +36,23 @@ en:
     server_error_text: We're experiencing a temporary outage due to an external system error. Please try again later. 
     manage_org: Manage your organization.
     tos_not_signed: You must sign DPC Terms of Service.
+    sign_tos: Sign terms of service
```
jdettmannnava, line 39: It looks like changing the key broke the accessibility tests, which should only fail on the fact that many pages do not meet accessibility. Please change back to tos_not_signed or update spec/system/accessibility_spec.rb
You can run the accessibility tests with `make ci-portal-accessibility`
I think you will need to find another way to make the "mixed ao cd" test in the accessibility tests to pass as well...

dpc-portal/app/components/page/organization/organization_list_component.html.erb
```diff
@@ -1,16 +1,31 @@
...
+<p>Use this portal to manage your organization's access to Data at the Point of Care (DPC). Depending on your role, you can register your organization, sign terms of service, and complete API configuration.</p>
+<div class="display-flex flex-row flex-justify">
+  <div class="flex-align-self-center">
+    <h2 class="margin-bottom-0">My organizations</h2>
   </div>
-  <p>These are organizations for which you serve as an AO. Organizations cannot access Medicare claims data without an AO agreement to follow terms of service.</p>
+</div>
+<p>You’re designated as an Authorized Official (AO) and/or Credential Delegate (CD) of the following organizations. An AO is accountable for organization-level access to the API. The AO can can assign a CD to complete configuration steps on their behalf.</p>
+
+<% if @links.length %>
```
jdettmannnava, line 11: `@links.present?` is the usual way to check this

PR #2569, opened 2025-04-07
Update Accessiblity tests to match message

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4589

## 🛠 Changes

Selector messages updated to match reality

## ℹ️ Context

We updated success messages that are used in the ci-portal-accessibilty tests, so they no longer matched. Since the accessibilty tests fail on accessibility issues, we don't run them, so we didn't notice.

## 🧪 Validation
ci-portal-accessibility fails only accessiblity


MEspositoE14s (approved):
LGTM 👍

PR #2577, opened 2025-04-09
Fix flaky AttributionFHIRTest

ashley-weaver:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4654

## 🛠 Changes

Sorts patients in test to keep order consistent.

## ℹ️ Context

AttributionFHIRTest was failing stochastically due to the member list not having a consistent order.

## 🧪 Validation

Test passing reliably.


MEspositoE14s (approved):
I am straight up shocked that that the built in `compareDeep` doesn't handle two lists being in different orders.  I always assumed it would.  Good find!

ashley-weaver:
> I am straight up shocked that that the built in `compareDeep` doesn't handle two lists being in different orders. 

you would think, and yet...
```
for(int i = 0; i < e1.size(); ++i) {
  if (!compareDeep((Base)e1.get(i), (Base)e2.get(i), allowNull)) {
    return false;
  }
}

PR #2590, opened 2025-04-16
[DPC-4635] Added load test options in GHA

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4635

## 🛠 Changes

Added test selection drop down in load test workflow.

## ℹ️ Context

We'll be adding "average" and "stress" load tests and need a way select a particular test suite.

## 🧪 Validation

Triggered GHA and made sure we could successfully run existing load test script.  See workflow run [here](https://github.com/CMSgov/dpc-app/actions/runs/14498818062).

### DIFF
--- a/.github/workflows/dpc-load-test.yml
+++ b/.github/workflows/dpc-load-test.yml
@@ -2,6 +2,16 @@ name: DPC Load Test
 
 on:
   workflow_dispatch:
+    inputs:
+      test_type:
+        description: Load test script to run
+        required: true
+        type: choice
+        options:
+          - script  # Leaving this for now until the average and stress test scripts are created
+          - average
+          - stress
+        default: script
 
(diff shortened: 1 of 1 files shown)


.github/workflows/dpc-load-test.yml
```diff
@@ -2,6 +2,16 @@ name: DPC Load Test
 
 on:
   workflow_dispatch:
+    inputs:
+      test_type:
+        description: Load test script to run
+        required: true
+        type: choice
+        options:
+          - script  # Leaving this for now until the average and stress test scripts are created
```
MEspositoE14s, line 11: We can delete this once the average and stress tests are created, but in the meantime I wanted to test my changes, so I added this to call the existing script.js

PR #2594, opened 2025-04-18
[DPC-4632] Average Load Test

chris-ronning-ny:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4632

## 🛠 Changes

- Creates the configuration for the average load test
- Rewrites the workflow code to accommodate precise number of API calls
- Removes workflowB
- Added a single-iteration config so that we can test exactly one run-through of the workflow

This PR consists of a pretty significant rewrite to the workflow logic. `http.batch` is used in numerous places to run several API calls in parallel. 

Code is written to make the following number of API calls over the course of the average load test:

- POST Patient - 700 total - 14 per iteration
- POST Practitioner - 50 total - 1 per iteration
- POST Group - 50 total - 1 per iteration
- GET Practitioner - 500 total - 10 per iteration
- GET Patient - 250 total - 5 per iteration
- GET Group - 50 total - 1 per iteration
- PUT Group - 250 total - 5 per iteration
- GET Group/$export - 250 total - 5 per iteration
- GET Job - 500 total - 10 per iteration

Over the course of the 50 iterations specified in the average-load-test config, we'll hit the correct total number for each endpoint. This will then scale for more iterations per hour when we configure the stress test. Numbers are slightly adjusted from those in the ticket to simplify counting. They all fall well within the ranges noted in the [Average Hourly Load doc](confluence.cms.gov/display/DAPC/Hourly+API+Load). 

In order to simplify the API call counts, `workflowB` was removed in favor of a single workflow. 

## ℹ️ Context

This enables us to run an Average Load Test, so that we can analyze how our system operates under the average hourly load, in a controlled environment.

## 🧪 Validation

- [x] linter and tests pass
- [x] single-iteration run passes without failed http requests - https://github.com/CMSgov/dpc-app/actions/runs/14598618165/job/40950837809


dpc-load-testing/workflows.js
```diff
@@ -70,153 +71,94 @@ export function setup() {
...
 
-  const orgId = data.orgIds[exec.vu.idInInstance];
+  const orgId = data.orgIds[1];
   const token = generateDPCToken(orgId, data.goldenMacaroon);
 
   // POST practitioner
   const practitionerResponse = createPractitioner(token, npiGenerator.iterate());
   if (practitionerResponse.status != 201) {
-    fail('failed to create practitioner for workflow A');
+    console.error('failed to create practitioner for workflow A');
```
jdettmannnava, line 88: Won't this make it keep going, even if it can't make a practitioner? Do we want this? Or maybe fail when we actually need this information?

dpc-load-testing/workflows.js
```diff
@@ -70,153 +71,94 @@ export function setup() {
...
-  if (!jobId) {
-    fail('failed to get a location to query the export job in workflow A');
-  }
-  const jobResponse = findJobById(token, jobId);
-  if (jobResponse.status != 200 && jobResponse.status != 202) {
-    fail('failed to successfully query job in workflow A');
-  }
-}
-
-export function workflowB(data) {
```
jdettmannnava, line 152: Are we just abandoning workflowB?

dpc-load-testing/workflows.js
```diff
@@ -70,153 +71,94 @@ export function setup() {
...
   return { orgIds: orgIds, goldenMacaroon: goldenMacaroon };
 }
 
-export function workflowA(data) {
+export function workflow(data) {
   const npiGenerator = npiGeneratorCache.getGenerator(exec.vu.idInInstance);
   const mbiGenerator = mbiGeneratorCache.getGenerator(exec.vu.idInInstance);
 
-  const orgId = data.orgIds[exec.vu.idInInstance];
+  const orgId = data.orgIds[1];
```
jdettmannnava, line 82: Unsure why we are limiting to one org?

dpc-load-testing/workflows.js
```diff
@@ -70,153 +71,94 @@ export function setup() {
...
   const practitionerId = practitionerResponse.json().id;
 
-  // POST patient
-  const patientResponse = createPatient(token, mbiGenerator.iterate());
-  if (patientResponse.status != 201) {
-    fail('failed to create patient for workflow A');
-  }
-  const patientId = patientResponse.json().id;
+  // POST patients
+  const patientResponses = createPatients(token, 14, mbiGenerator);
```
jdettmannnava, line 95: Can we abstract 14? Feels like a magic number

dpc-load-testing/workflows.js
```diff
@@ -70,153 +71,94 @@ export function setup() {
...
+  // GET patients by MBI
+  const findPatientsResponses = findPatientsByMbi(token, patients.map((patient) => patient.patientMbi));
+  findPatientsResponses.forEach((res) => {
+    if (res.status != 200) {
+      console.error('failed to GET patient.');
+    }
+  });
+
+  // distribute (PUT) patients into group
+  const updateGroupResponses = addPatientsToGroup(token, orgId, groupId, patients.splice(0, 5), practitionerId, practitionerNpi);
```
jdettmannnava, line 130: Maybe move 5 into a constant?

dpc-load-testing/workflows.js
```diff
@@ -70,153 +71,94 @@ export function setup() {
...
+    }
+  })
+  
 
   // GET group export
-  const getGroupExportResponse = exportGroup(token, groupId);
-  if (getGroupExportResponse.status != 202) {
-    fail('failed to export group for workflow B');
+  const jobIds = [];
+  for (let i = 0; i < 5; i++) {
```
jdettmannnava, line 140: Use the constant above here? Or a new constant?

dpc-load-testing/workflows.js
```diff
@@ -70,153 +71,94 @@ export function setup() {
...
+    const getGroupExportResponse = exportGroup(token, groupId);
+    if (getGroupExportResponse.status != 202) {
+      console.error('failed to export group for workflow A');
+    } else {
+      const jobId = getGroupExportResponse.headers['Content-Location'].split('/').pop();
```
jdettmannnava, line 145: I think this is kind of odd. Couldn't we just save the location and get job by location below?

dpc-load-testing/workflows.js
```diff
@@ -70,153 +71,94 @@ export function setup() {
...
 
-  const jobId = getGroupExportResponse.headers['Content-Location'].split('/').pop();
-  if (!jobId) {
-    fail('failed to get a location to query the export job in workflow B');
-  }
-  const jobResponse = findJobById(token, jobId);
-  if (jobResponse.status != 200 && jobResponse.status != 202) {
-    fail('failed to successfully query job in workflow B');
+  // GET job status
+  for (let i = 0; i < 2; i++) {
```
jdettmannnava, line 155: Pleas move 2 to a constant

dpc-load-testing/workflows.js
```diff
@@ -70,153 +71,94 @@ export function setup() {
...
-  }
-
-  // GET patient
-  const getPatientResponse = findPatientByMbi(token, patientMbi);
-  if (getPatientResponse.status != 200) {
-    fail('failed to get patient for workflow B');
-  }
-
   // GET group by practitioner NPI
   const getGroupResponse = findGroupByPractitionerNpi(token, practitionerNpi);
```
jdettmannnava, line 116: Shouldn't we also get all groups, and by id?

Makefile
```diff
@@ -105,7 +105,7 @@ start-portals: start-db start-web start-admin start-portal
 
 start-load-tests: ## Run DPC performance tests locally in a Docker image provided by Grafana/K6
 start-load-tests: secure-envs
-	@docker run --rm -v $(shell pwd)/dpc-load-testing:/src --env-file $(shell pwd)/ops/config/decrypted/local.env -e ENVIRONMENT=local -i grafana/k6 run /src/script.js
+	@docker run --rm -v $(shell pwd)/dpc-load-testing:/src --env-file $(shell pwd)/ops/config/decrypted/local.env -e ENVIRONMENT=local -i grafana/k6 run /src/single-iteration.js
```
jdettmannnava, line 108: Maybe add another one for average load test?

dpc-load-testing/single-iteration.js
```diff
@@ -0,0 +1,16 @@
...
+// details on this configuration object.
+export const options = {
+  scenarios: {
+    workflow_1: {
+      executor: 'per-vu-iterations',
+      vus: 1,
+      iterations: 1,
+      exec: "workflow"
+    }
+  }
```
jdettmannnava, line 13: I am not sure why we can't still have workflow B here?

dpc-load-testing/dpc-api-client.js
```diff
@@ -53,10 +53,42 @@ export function createPatient(token, mbi) {
   return res;
 }
 
+export function createPatients(token, number, mbiGenerator) {
```
jdettmannnava, line 56: I don't think this is how most people use the API, as I think they iterate rather than use multi-threaded tools. Which doesn't mean this has to change; just putting down a thought.

dpc-load-testing/workflows.js
chris-ronning-ny: Yes, they were already nearly identical, and having both of them makes calculating number of calls per iteration much more complicated. We don't really lose anything by getting rid of this.

dpc-load-testing/workflows.js
chris-ronning-ny: Previously, we had just split them so that workflowA gets 1 org and workflowB gets another one. I don't really see why we would need more than one.

dpc-load-testing/workflows.js
chris-ronning-ny: Hum the idea was to use the pre-constructed request function `findJobsById` and just plug in . I could rework that function to receive a URL.

dpc-load-testing/workflows.js
chris-ronning-ny: My rationale here was that both of these methods use the API endpoint `GET v1/Group`, and so effectively we're testing the same load even if we only use one of them. Do you think I'm missing any value by only using one of these methods?

dpc-load-testing/single-iteration.js
chris-ronning-ny: See previous comment. I think workflowB is entirely unnecessary and overcomplicates the whole thing.

dpc-load-testing/workflows.js
chris-ronning-ny: hmm. I think some of these calls don't need to end the entire iteration if they fail, but the practitioner probably does. Let me reconsider a few of these.

Makefile
chris-ronning-ny: Done

dpc-load-testing/workflows.js
jdettmannnava: Huh, because from the load test strategy, there were two separate thing we wanted to test: POST heavy and GET heavy. While they are similar in terms of ordering, I thought we wanted them separate so we could tweak different aspects of the flow?

dpc-load-testing/dpc-api-client.js
jdettmannnava: "Batch multiple HTTP requests together to issue them in parallel over multiple TCP connections." So, how is this in sequence?

chris-ronning-ny: It's not running them at the same time, it's just not waiting for one to return before moving on to the next one. I guess we're just talking semantics here. It wouldn't be hard to change this to run one http.post after the other. My only concern with that is, if API service begins to degrade, and the load test function is just holding until calls complete, we could run into some execution issues.

dpc-load-testing/workflows.js
jdettmannnava: I can't decide whether a bunch of orgs would be more or less efficient, so that might be something worth testing? Also, if we do decide on just one, then maybe make it not an array? Like http.batch, mostly marking down a concern rather than requesting a change.

dpc-load-testing/workflows.js
```diff
@@ -70,153 +78,101 @@ export function setup() {
...
+  for (let i = 0; i < requestCounts.getGroupExport; i++) {
+    const getGroupExportResponse = exportGroup(token, groupId);
+    if (getGroupExportResponse.status != 202) {
+      console.error('failed to export group for workflow A');
+    } else {
+      let findJobUrl = getGroupExportResponse.headers['Content-Location'];
+      if (__ENV.ENVIRONMENT === 'local') {
+        findJobUrl = findJobUrl.replace('localhost', 'host.docker.internal');
+      }
+      if (!findJobUrl) {
```
jdettmannnava, line 160: this should probably bock the replace

dpc-load-testing/dpc-api-client.js
```diff
@@ -105,19 +137,48 @@ export function updateGroup(token, orgId, groupId, patientId, practitionerId, pr
...
-      return res;
+    return res;
 }
 
 export function findJobById(token, jobId) {
   // Request fails with 406 status when made with Content-Type or Accept header
   return http.get(`${urlRoot}/Jobs/${jobId}`, { 'headers': { 'Authorization': `Bearer ${token}` } });
 }
 
+export function findJobsById(token, urls) {
```
jdettmannnava, line 169: Not finding by id

dpc-load-testing/workflows.js
```diff
@@ -70,153 +78,101 @@ export function setup() {
...
 
   // GET group export
-  const getGroupExportResponse = exportGroup(token, groupId);
-  if (getGroupExportResponse.status != 202) {
-    fail('failed to export group for workflow B');
+  const findJobUrls = []
+  for (let i = 0; i < requestCounts.getGroupExport; i++) {
+    const getGroupExportResponse = exportGroup(token, groupId);
+    if (getGroupExportResponse.status != 202) {
+      console.error('failed to export group for workflow A');
```
jdettmannnava, line 154: workflow A

dpc-load-testing/workflows.js
```diff
@@ -70,153 +78,101 @@ export function setup() {
...
+  });
+
+  // distribute (PUT) patients into group
+  const updateGroupResponses = addPatientsToGroup(token, orgId, groupId, patients.splice(0, requestCounts.addPatientsToGroup), practitionerId, practitionerNpi);
+  updateGroupResponses.forEach((res) => {
+    if (res.status != 200) {
+      console.error('failed to add patient to group');
+    }
+  })
+
```
jdettmannnava, line 147: Do we want to make sure of at least one success?

chris-ronning-ny:
@jdettmannnava After thinking a bit about orgs, I set it up so that we generate a pre-defined number of organizations, based on the max number of VUs we want for the scenario. We then distribute them during the workflow based on VU ID. Max number of VUs is set based on the test type by environment variable -- will configure this for stress test in the other PR.

dpc-load-testing/average-load-test.js
```diff
@@ -0,0 +1,20 @@
...
+// See https://grafana.com/docs/k6/latest/using-k6/k6-options/reference for
+// details on this configuration object.
+export const options = {
+  scenarios: {
+    workflow_a: {
+      executor: 'constant-arrival-rate',
+      duration: '1h',
+      rate: 50,
+      timeUnit: '1h',
+      preAllocatedVUs: 3,
```
jdettmannnava, line 13: unsure why this isn't constants.maxVUs

dpc-load-testing/average-load-test.js
```diff
@@ -0,0 +1,20 @@
+import { workflow, setup, teardown } from "./workflows.js";
+import { constants } from "./constants.js";
+
+// See https://grafana.com/docs/k6/latest/using-k6/k6-options/reference for
+// details on this configuration object.
+export const options = {
+  scenarios: {
+    workflow_a: {
```
jdettmannnava, line 8: workflow_a

Makefile
```diff
@@ -103,9 +103,14 @@ start-portal: secure-envs
 start-portals: ## Start all frontend services
 start-portals: start-db start-web start-admin start-portal
 
-start-load-tests: ## Run DPC performance tests locally in a Docker image provided by Grafana/K6
+start-load-tests: ## Run one iteration of DPC performance tests locally in a Docker image provided by Grafana/K6
 start-load-tests: secure-envs
-	@docker run --rm -v $(shell pwd)/dpc-load-testing:/src --env-file $(shell pwd)/ops/config/decrypted/local.env -e ENVIRONMENT=local -i grafana/k6 run /src/script.js
+	@docker run --rm -v $(shell pwd)/dpc-load-testing:/src --env-file $(shell pwd)/ops/config/decrypted/local.env -e ENVIRONMENT=local -e TEST_TYPE=single-iteration -i grafana/k6 run /src/single-iteration.js
```
jdettmannnava, line 108: The duplication here feels a little dodgy

dpc-load-testing/single-iteration.js
```diff
@@ -0,0 +1,17 @@
+import { constants } from "./constants.js";
```
jdettmannnava, line 1: Can you set the environment variable before the import?

Makefile
chris-ronning-ny: I don't really understand why? What feels dodgy?

dpc-load-testing/single-iteration.js
chris-ronning-ny: Not sure what you want here? The environment variable gets read in constants.js, and is read in the config object.

Makefile
jdettmannnava: The duplication of the environment variable and the script name. Nothing concrete, just sort of flicked my spider sense.

dpc-load-testing/average-load-test.js
chris-ronning-ny: Because they're conceptually different. The baseline for both this and the stress test will be 3 preAllocatedVUs. I suppose I can hardcode that into the constant.

Makefile
chris-ronning-ny: Updated to include a local variable for both commands.

jdettmannnava (approved):
LGTM

PR #2595, opened 2025-04-18
PLT-657:Rename the public platform github repo

christopher-maboh:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-657

## 🛠 Changes

references to ab2d-bcda-dpc-platform renamed to cdap

## ℹ️ Context

we are renaming the ab2d-bcda-dpc-platform repo to cdap

## 🧪 Validation

upon approval, ab2d-bcda-dpc-platform repo will be renamed to cdap and workflow re-ran, if successful PR merged if not ab2d-bcda-dpc-platform repo rename to cdap reverted while investigating why workflow failed


jdettmannnava:
Please don't do anything until we have finished this sprint's deploy. I will post here when it is done

christopher-maboh:
> Please don't do anything until we have finished this sprint's deploy. I will post here when it is done

ok Will wait for your go ahead

christopher-maboh:
> I don't think you'll be able to merge this until the tests pass, and the tests won't pass unless the repo is renamed first. I haven't tested this, but supposedly if you just rename the repo the existing name will be treated as an alias to the new name by Github. If that's the case, you might be good to just rename it without worrying about breaking anything. I'd definitely want to double check that first, though.

yes as indicated in validation, I want to get approvals first before renaming so once the check passes I just merge right away rather than waiting for approvals and because its already been renamed workflows start failing without the branch being merged 
so the plan is to - get approvals
-rename repo and run test workflows again, if successful merge if not successful revert the name chnage while investigating the failure. I will only merge once the rename is successful

jdettmannnava:
@christopher-maboh : deployment complete! Thanks for waiting

juliareynolds-nava (dismissed):

PR #2599, opened 2025-04-21
[DPC-4668] Use tfenv for terraform install in CI

lukey-luke:
## 🎫 Ticket

[DPC-4668](https://jira.cms.gov/browse/DPC-4668)

## 🛠 Changes

 - Rather than manually specifying terraform version used, use the terraform version located in[ .terraform-version](https://github.com/CMSgov/dpc-ops/blob/main/.terraform-version)
 - Use platform's [existing Github Action](https://github.com/CMSgov/ab2d-bcda-dpc-platform/blob/main/actions/setup-tfenv-terraform/action.yml) to handle the install

## ℹ️ Context

  - As we update our infrastructure to work with the new AWS greenfield account, we should take care to follow best practices.
  - CDAP recommended using tfenv as that's what other teams are doing.

## 🧪 Validation

 - tested deploy to test environment [HERE](https://github.com/CMSgov/dpc-app/actions/runs/14580194608/job/40895025965)

### DIFF
--- a/.github/workflows/ecs-deploy.yml
+++ b/.github/workflows/ecs-deploy.yml
@@ -128,10 +128,10 @@ jobs:
           ref: ${{ inputs.ops-ref }}
           ssh-key: ${{ secrets.DPC_OPS_SSH_KEY }}
           path: dpc-ops
-      - name: install terraform
-        uses: hashicorp/setup-terraform@v3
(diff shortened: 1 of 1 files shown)


MEspositoE14s (approved):
LGTM 👍

PR #2608, opened 2025-04-25
Me/dpc 4659 fix aggregation mdc

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4659

## 🛠 Changes

Resets MDC between batches in dpc-aggregation.

## ℹ️ Context

Previously, when an aggregation node would log that it restarted a stuck batch the MDC would still be filled in with information from the previous batch (job_id, batch_id, etc...).  This is because stuck batches are handled between processed job batches, and the MDC was only getting cleared when a new batch was started. Now it gets cleared after each batch is finished.

## 🧪 Validation

Add MDC check to aggregation unit tests.

### DIFF
--- a/dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/engine/AggregationEngineTest.java
+++ b/dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/engine/AggregationEngineTest.java
@@ -31,7 +31,6 @@ import org.assertj.core.util.Lists;
 import org.hl7.fhir.dstu3.model.Bundle;
 import org.junit.jupiter.api.*;
 import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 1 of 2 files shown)


dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/engine/AggregationEngineTest.java
```diff
@@ -715,7 +738,7 @@ void otherExceptions(Exception e) {
 
         private static List<String> getLogMessages() {
             boolean success = false;
-            int count = 0, maxTries = 20;
+            int count = 0, maxTries = 50;
```
MEspositoE14s, line 741: Not really part of this ticket, but this started failing for me locally and bumping this up got it working again.

lukey-luke (approved):
LGTM!

PR #2613, opened 2025-04-28
[DPC-4673] Fix NPI Generator

chris-ronning-ny:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4673

## 🛠 Changes

Fixes the NPI Generator by pulling in the [luhn-js](https://www.npmjs.com/package/luhn-js) library, replacing my hand-made implementation of the [Luhn algorithm](https://en.wikipedia.org/wiki/Luhn_algorithm).

## ℹ️ Context

During some testing, we found that POST v1/Practitioner was returning 400 errors after around 6 iterations. As it turns out, this was due to some errors in our NPI Generator. Pulling in the luhn-js third party library to replace the faulty logic resolves this problem. 

Because K6 does not run a Node.js runtime, we're unable to pull in NPM modules the usual way. I opted to get around this by pulling the library's main file into the `lib` directory, and import from there.

You can test this by updating `single-iteration.js` to run more than 6 total iterations, then running `make load-tests` locally. In my case, I ran 10 iterations per VU = 30 iterations.

## 🧪 Validation

- [x] Load Test runs successfully in GH Actions for `single-iteration`. https://github.com/CMSgov/dpc-app/actions/runs/14715636860/job/41298239302
- [x] Update `single-iteration.js` to have an `iterations` value of more than 6. Running `make load-tests` passes with no failed calls to either POST v1/Organization or POST v1/Practitioner. See screenshot for my own run.

Screenshot 2025-04-28 at 2 49 25 PM

### DIFF
--- a/dpc-load-testing/__tests__/npi-generator.test.js
+++ b/dpc-load-testing/__tests__/npi-generator.test.js
@@ -1,25 +1,6 @@
 import { generateNPI } from "../utils/npi-generator";
 import { isArrayUnique } from "../utils/test-utils";
-
-function luhnCheck(npi) {
-  const appendedNpi = "80840" + npi;
-  const reversed = appendedNpi.split('').reverse();
-  let total = 0;
-  for (const [index, value] of reversed.entries()) {
-    let x = parseInt(value);
-    if (index % 2 === 1) {
-      x = x * 2;
-      // if the result has two digits, add the digits together to get a single digit number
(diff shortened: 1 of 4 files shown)


lukey-luke (approved):
I was able to test locally with 10 iterations and everything's passing. Confirmed MIT license. LGTM!

PR #2632, opened 2025-05-12
[DPC-4700] use HOST_NAME for login.gov callback url

chris-ronning-ny:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4700

## 🛠 Changes

- Use HOST_NAME from environment variables to set login.gov callback url

## ℹ️ Context

We need to be able to get the correct URL when running in Greenfield. 

This depends on changes submitted in https://github.com/CMSgov/dpc-ops/pull/808/


lukey-luke (approved):
dependent PR needs to get merged first but LGTM

PR #2641, opened 2025-05-19
[DPC-4690] WIP: create Build and Deploy (Greenfield) workflow

chris-ronning-ny:
## 🎫 Ticket

### DIFF
--- a/.github/workflows/ecs-release-gf.yml
+++ b/.github/workflows/ecs-release-gf.yml
@@ -1,0 +1,109 @@
+name: 'DPC - Build and Deploy (Greenfield)'
+
+on:
+  schedule:
+    - cron: '0 6 * * 1-5'
+  push:
+    branches:
+      - main
+    paths:
+      - dpc-**
+      - engines/**
+      - .github/workflows/**
+
+  workflow_dispatch:
+    inputs:
+      env:
+        description: AWS environment to deploy to
+        required: true
+        type: 'string'
+        default: 'dev'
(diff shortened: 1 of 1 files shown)


.github/workflows/ecs-release-gf.yml
```diff
@@ -0,0 +1,113 @@
+name: 'DPC - Build and Deploy (Greenfield)'
+
+on:
+  schedule:
+    - cron: '0 6 * * 1-5'
+  push:
+    branches:
+      - main
```
chris-ronning-ny, line 8: Now that I'm thinking of it -- with this, we're going to be deploying to both legacy and greenfield dev on every merge to main. Do we want to disable the legacy workflow in the scope of this PR?

johnbeahan: I would say yes, since we only want to deploy to legacy on-demand for now, during the GFM, would be my vote

.github/workflows/ecs-release.yml
```diff
@@ -1,16 +1,6 @@
...
-  schedule:
-    - cron: '0 6 * * 1-5'
-  push:
-    branches:
-      - main
-    paths:
-      - dpc-**
-      - engines/**
-      - .github/workflows/**
-
```
chris-ronning-ny, line 13: @johnbeahan @MEspositoE14s I removed the scheduled deployment to legacy, in addition to the merge deployment that we discussed. Any objections?

PR #2646, opened 2025-05-23
[DPC-4687] migrate smoke test workflow to greenfield

lukey-luke:
## 🎫 Ticket

[DPC-4687](https://jira.cms.gov/browse/DPC-4687)

## 🛠 Changes

Add new GHA workflow to run smoketests in new greenfield aws accounts

## ℹ️ Context

  - This is part of the greenfield migration to two new AWS accounts

## 🧪 Validation

See workflow running both legacy smoketests and greenfield smoketests [HERE](https://github.com/CMSgov/dpc-app/actions/runs/15214255543).

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -91,7 +91,6 @@ jobs:
           echo "HOST_URL=$HOST/api/v1" >> "$GITHUB_OUTPUT"
           PUBLIC_HOST="https://${{ steps.fetch-admin-host.outputs.public_host }}"
           echo "PUBLIC_HOST=$PUBLIC_HOST" >> "$GITHUB_OUTPUT"
-#          echo "PUBLIC_URL=$PUBLIC_HOST/api/v1" >> "$GITHUB_OUTPUT"
       - name: Run Smoke Test
         id: run-smoke-test
         env:


.github/workflows/smoke-test-gf.yml
```diff
@@ -0,0 +1,136 @@
...
+        run: |
+          mvn -ntp -U clean
+      - name: AWS Credentials (non-prod)
+        if: ${{ inputs.env == 'dev' || inputs.env == 'test' }}
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.NON_PROD_ACCOUNT_ID }}:role/delegatedadmin/developer/dpc-${{ inputs.env }}-github-actions
+      - name: AWS Credentials (prod)
+        if: ${{ inputs.env == 'prod-sbx' || inputs.env == 'prod' }}
```
chris-ronning-ny, line 63: I know Jeff mentioned that `prod-sbx` was being abandoned in favor of `sandbox` for naming. Should we be making that update in these workflows? Probably worth a bigger discussion outside of this PR.

lukey-luke: If `prod-sbx` isn't the appropriate naming, I think it makes sense to use `sandbox`. More importantly, I think we want to be consistent with whichever decision we make.

I suggest we address this naming change in all the workflow files in a separate PR. I'm happy to submit a follow-up! I'm seeing this affect 
- ecs-deploy-gf.yml
- opt-out-export-test-integration-gf.yml
- smoke-test-gf.yml

How does that sounds to you @chris-ronning-ny?

PR #2647, opened 2025-05-23
Add new cert to keystore and update bfd url

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4728

## 🛠 Changes

- Updates bb.keystore with new public key from BFD.
- Updates BFD prod-sbx address in local.env.

## ℹ️ Context

BFD is updating their sandbox url, and we need a new cert to access it.  

Note: There will be another PR updating our persistent environments in dpc-ops once we have the cert for production BFD.

## 🧪 Validation

Ran dpc-api locally and made sure health check was still passing for api and aggregation.

If you run `make secure-envs` you can see the new value of BFD_URL in `/ops/config/decrypted/local.env` and the new keystore in `/bbcerts/bb.keystore`.  To read the contents of the keystore, run `keytool -list -keystore bb.keystore`.


bbcerts/bb.keystore
chris-ronning-ny, line 1: huh, why are these two being deleted? Just out of curiosity

PR #2652, opened 2025-05-27
Work in 'sandbox' as well as 'prod-sbx' as environment

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4732

## 🛠 Changes

- Add config files for 'sandbox'
- Check for 'sandbox' where check for 'prod-sbx'

## ℹ️ Context
Application code must work in both prod-sbx (legacy) and sandbox (greenfield) as environments

## 🧪 Validation

Passes smoke tests


dpc-admin/app/models/registered_organization.rb
```diff
@@ -49,7 +49,7 @@ def notify_users_of_sandbox_access
   end
 
   def prod_sbx?
```
lukey-luke, line 51: can we switch the function name to `sandbox?` to reflect the latest name change?

jdettmannnava: I would really rather not, as I would like to keep this refactor as small as possible.

dpc-admin/spec/controllers/registered_organizations_controller_spec.rb
```diff
@@ -29,10 +29,10 @@
       end
     end
 
-    context 'successful prod-sbx API organization new' do
-      it 'renders the blank registerd org fields with the org for prod-sbx' do
+    context 'successful sandbox API organization new' do
```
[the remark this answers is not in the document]
jdettmannnava: good idea

PR #2661, opened 2025-06-04
PLT-1031: Remove legacy GHA workflows

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1031

## 🛠 Changes

Removes legacy workflows from the repository.

## ℹ️ Context

We've now fully cut-over to the greenfield accounts, so there is not much use for us to keep these old workflows around.

## 🧪 Validation

N/a


gsf (commented):
Pretty clean. I'll defer to the DPC team to approve.

.github/workflows/opt-out-export-test-integration-gf.yml
```diff
@@ -33,6 +33,9 @@ jobs:
     uses: ./.github/workflows/opt-out-export-deploy-gf.yml
     with:
       env: test
+    permissions:
+      contents: read
+      id-token: write
```
lukey-luke, line 38: good catch

lukey-luke (approved):
I've confirmed all these workflows have already been migrated (see "gf" suffixed versions in `.github/workflows`)

PR #2669, opened 2025-06-10
[On Call] DropWizard 4.0.11 -> 4.0.13

MEspositoE14s:
## 🎫 Ticket

None, part of on call

## 🛠 Changes

- Update DropWizard from 4.0.11 -> 4.0.13
- Update location of `IAccessEvent` class in `AuthRelatedLoggingFilter`.

## ℹ️ Context

Snyk opened https://github.com/CMSgov/dpc-app/pull/2640 to upgrade `dropwizard-json-logging`, and all of DropWizard, from 4.0.11 to 4.0.13.  The PR fails with a compilation error, though, because the `IAccessEvent` class was moved to a new location in the logback library.

## 🧪 Validation

Program compiles and all tests pass.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <dropwizard.version>4.0.11</dropwizard.version>
+        <dropwizard.version>4.0.13</dropwizard.version>
         <junit.jupiter.version>5.12.1</junit.jupiter.version>
         <slf4j.version>2.0.17</slf4j.version>
         <hapi.fhir.groupID>ca.uhn.hapi.fhir</hapi.fhir.groupID>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -302,6 +297,18 @@
                 <artifactId>dropwizard-jobs-guice</artifactId>
                 <version>6.0.1</version>
             </dependency>
+            <!-- These logback dependencies are to prevent dropwizard-jobs from downgrading them -->
+            <!-- Dropwizard:4.0.14 pulls logback 1.5.18, dropwizard-jobs:6.0.1 tries to pull 1.4.14 -->
```
jdettmannnava, line 301: Is upgrading jobwizard-jobs an option?

MEspositoE14s: Interestingly enough, they released Dropwizard-jobs 6.0.2 in February of last year and that fixes the problem, but they haven't published it to the Maven central repo so we can't pull it in and use it.  

We could download the source and build it ourselves, if we really wanted to.

PR #2676, opened 2025-06-17
DPC-4663 Upgrade aws go sdk to v2

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4663

## 🛠 Changes
- Imports from aws-sdk-go changed to aws-sdk-go-v2 for all lambdas
  - Session removed; replaced with Config
  - Update to many method signatures
- Removal of tests that seemed to only test its own configuration
- Go version changed to 1.24
- Opened port 5432 on docker postgres container

## ℹ️ Context

Upgrade necessary, as aws-sdk-go will no longer be supported after July 31, 2025; and it contains a Snyk vulnerability.
Note that aside from removal of "Session" objects in favor of "Config" objects, there has been a move away from using pointers. This results in the inability to return _nil_ as a response. However, in those cases, we return an error, which is checked for error handling anyway.

Opening up port 5432 on postgres container allows all opt-out-import tests to pass.

This involves using a new library, so would like review by @SJWalter11 .

## 🧪 Validation

Integration tests pass via Github Actions

### DIFF
--- a/maven-config.sh
+++ b/maven-config.sh
@@ -5,11 +5,17 @@ mkdir -p ./.mvn
 : > ./.mvn/maven.config
 
 function process_line() {
+    # Don't process an empty line
+    if [ -z "$1" ]
+    then
+      return
+    fi
+
   KEY="$(echo "$1" | cut -d '=' -f1)"
   VALUE="$(echo "$1" | cut -d '=' -f2)"
 
   # Quoted values need special handling
-  # Before: Key="Value"
+  # Before: -DKey="Value"
   # After: "-DKey=Value"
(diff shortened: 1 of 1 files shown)


lambda/api-waf-sync/aws_test.go
```diff
@@ -1,63 +0,0 @@
-package main
```
lukey-luke, line 1: 👍

lukey-luke:
Link to integration tests here: https://github.com/CMSgov/dpc-app/actions/runs/15736273870

MEspositoE14s (approved):
I'm far from a Go expert, but it looks good to me.

PR #2687, opened 2025-06-24
[DPC-4726] Remove SigningKeyResolver

ashley-weaver:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4726

## 🛠 Changes

Removes use of deprecated class.

## ℹ️ Context

[SigningKeyResolver](https://javadoc.io/doc/io.jsonwebtoken/jjwt-api/0.12.2/io/jsonwebtoken/SigningKeyResolver.html) has been deprecated and will be removed in the 1.0 release of `io.jsonwebtoken`. This PR replaces its usage with a LocatorAdapter and direct token inspection.

## 🧪 Validation

Tests passing.


MEspositoE14s (dismissed):
Most of my comments are pretty nit-picky and not worth holding up a review for.  LGTM.

chris-ronning-ny:
I'm curious about your note here

```
Note: these changes require the issuer to be included in the header when posting requests to /Token/validate.
```

It sounds like this is a change that our users will need to implement in their auth workflows. If that's the case, we'll need to run this by product and make sure the messaging gets out. Is there any way we can avoid this requirement?

ashley-weaver:
> I'm curious about your note here
> 
> ```
> Note: these changes require the issuer to be included in the header when posting requests to /Token/validate.
> ```
> 
> It sounds like this is a change that our users will need to implement in their auth workflows. If that's the case, we'll need to run this by product and make sure the messaging gets out. Is there any way we can avoid this requirement?

That comment was leftover from a previous solution, this PR now has no implementation changes for end users.

chris-ronning-ny (dismissed):
LGTM!

dpc-api/src/main/java/gov/cms/dpc/api/auth/jwt/TokenValidator.java
```diff
@@ -104,14 +105,22 @@ void validateClaims(String issuer, Map<String, Claim> claims) {
...
-        if (!audience.equals(this.audClaim)) {
+        Object audience;
+        Object audienceClaim;
+        try {
+            audience = getClaimIfPresent("audience", claims.get("aud")).asList(String.class);
+            audienceClaim = this.audClaim;
+        } catch (JWTDecodeException e) { // not a list
+            audience = getClaimIfPresent("audience", claims.get("aud")).asString();
+            audienceClaim = this.audClaim.get(0);
+        }
```
MEspositoE14s, line 116: I hate that this is necessary.  Do we have to allow customers to send a string or a list.  The example in our [docs](https://dpc.cms.gov/docsV1#5-json-web-tokens) is only a string, can we just disallow lists?

MEspositoE14s (dismissed):

MEspositoE14s (approved):

PR #2701, opened 2025-06-27
Run CI workflow only on changes to application code

jdettmannnava:
## 🎫 Ticket

No ticket

## 🛠 Changes

CI workflow updated to run on positive change to application code, rather than running except when only certain files changed.

## ℹ️ Context

When the CI workflow was set up, the vast majority of the files in the repository was related to application code, and the amount of time it took to run CI was much less, so it made sense to run except when certain files were changed. Given the current complexity of the repository, the likelihood of additional items that need to be whitelisted (e.g. the new lambda for handling Slack messaging of CloudWatch alerts), and the relative stability of the application code, it makes more sense to only run when application code changes.

## 🧪 Validation

Changes to code still force CI -- checks ran after merging main, which had some pom changes.


ashley-weaver:
should this also include `/src` in case we update those resource files in the future?

PR #2705, opened 2025-07-01
sanitize carriage return to prevent risk of log injection attack

lukey-luke:
## 🎫 Ticket

This is a follow-up to https://github.com/CMSgov/dpc-app/pull/2506 to fix SonarQube issue [HERE](https://sonarqube.cloud.cms.gov/project/issues?resolved=false&inNewCodePeriod=true&types=VULNERABILITY&branch=snyk-upgrade-39b12342bab248147129bb69a2b480df&id=bcda-dpc-api&open=AZe2XceYfFcKu0fFjs-c)

## 🛠 Changes

 - Remove carriage return from logs if user decides to put those in the URI.

## ℹ️ Context

  - Previously, I made a change to include URI in logs in the same location where Gzip headers are logged.
  - Sonarqube has detected a log injection vulnerability in this location of the code. This PR aims to mitigate risk of a log injection attack and keep sonarqube in a healthy state.
   - Changes based on the "How can I fix it?" criteria of linked SonarQube issue.

## 🧪 Validation

 - changes deployed to dev and manually verified logs show up in splunk
Screenshot 2025-07-01 at 1 17 02 PM
 - tested locally to verify newlines don't show up in logs:
 - typical GET {{baseUrl}}/api/v1/Group
 
Screenshot 2025-07-01 at 3 05 02 PM

 - malicious GET {{baseUrl}}/api/v1/Group?user=attacker%0Aadmin=true
 
Screenshot 2025-07-01 at 3 09 32 PM

PR #2706, opened 2025-07-02
Me/dpc 4710 graceful aggregation shutdown v2

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4710

## 🛠 Changes

- When `stop()` is called on the aggregation engine it, immediately pauses a currently running batch instead of waiting for the current patient to finish processing and then pausing the batch.
- Adds `TEST_PATIENT_TIME_OUT` mbi to the MockBlueButtonClient.  It's an mbi that when requested will cause the client to block and wait forever and is used to simulate long running BFD requests.
- Updates dpc-api to use the `MockBfdClient` instead of the real BFD sandbox when running locally.  This was already the case for dpc-aggregation, and since we can't connect to BFD from our local machines anymore this was preventing us from testing.

## ℹ️ Context

We were seeing regular alerts about stuck batches getting restarted whenever dpc-aggregation scaled down and nodes were getting removed.  

The way things were supposed to work:
1. AWS sends a SIGINT to a dpc-aggregation node telling it to shut down.
2. A Jetty shutdown thread is started, and it calls the `AggregationEngine.stop()` method.
3. `stop()` sets the AggregationEngine's queue to no longer running.
4. If a batch is currently being processed, the current patient is finished and then the batch is paused (status set to QUEUED).
5. The shutdown thread finishes cleaning up other resources that implement the `Managed` interface and the node shuts down.
6. If the node hasn't shut down 30 seconds after the SIGINT, a SIGKILL is sent and it's removed.

The problem is happening at step 4.  We often have patients that take multiple minutes to process.  If one of those patients is running when the SIGINT is sent, then there isn't time for it to finish and the batch to be paused before the aggregation node is killed.  The batch is left in a running state on the queue, and after 15 minutes of no activity, another node restarts it and logs a "Restarting stuck batch..." warning.  If we get enough of these in a given time period an alert is sent to #dpc-alerts.

This PR moves the call to `pauseBatch()` into the engine's `stop()` method.  This means it's called in the Jetty shutdown thread and happens immediately, while the current patient is still being processed, instead of waiting for it to finish.

## 🧪 Validation

- Deployed to dev and smoke tests passed (link [here](https://github.com/CMSgov/dpc-app/actions/runs/16026107894/job/45218537918)).
- Created a new test that simulates shutting down a node while a patient is being processed.
- Was able to shutdown a dpc-aggregation node normally (logs [here](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/%2Faws%2Fecs%2Ffargate%2Fdpc-dev-aggregation/log-events/aggregation%2Fdpc-dev-aggregation%2F636ee76219ae4c0182c840ee71b917b6)).

## Note:
This fixes the problem described above which seems to be causing most/all of our stuck batch issues, but not every problem.  There may still occasionally be stuck batch warnings that need to be investigated.

Failing SonarQube test being addressed in https://github.com/CMSgov/dpc-app/pull/2705

### DIFF
--- a/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
+++ b/dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
@@ -99,7 +99,7 @@ public class AggregationEngine implements Runnable {
         optionalBatch.ifPresent(jobQueueBatch -> this.queue.pauseBatch(jobQueueBatch, aggregatorID));
     }
 
-    public Boolean isRunning() {
+    public boolean isRunning() {
         return queueRunning.get();
     }
 
@@ -180,7 +180,7 @@ public class AggregationEngine implements Runnable {
             }
 
             // Finish processing the batch
-            if (Boolean.TRUE.equals(this.isRunning())) {
+            if (this.isRunning()) {
                 final String jobTime = SplunkTimestamp.getSplunkTimestamp();
                 // Calculate metadata for the file (length and checksum)
                 calculateFileMetadata(job);
--- a/dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/client/MockBlueButtonClient.java
+++ b/dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/client/MockBlueButtonClient.java
@@ -82,8 +82,7 @@ public class MockBlueButtonClient implements BlueButtonClient {
 
     @Override
     public Bundle requestPatientFromServerByMbi(String mbi, Map<String, String> headers) throws ResourceNotFoundException {
-        // If the TEST_PATIENT_TIME_OUT is requested, hang forever.  SonarQube will probably flag this, but since we're
-        // purposely trying to create a bad situation, we can ignore it.
+        // If the TEST_PATIENT_TIME_OUT is requested, hang forever.
         if (Objects.equals(mbi, TEST_PATIENT_TIME_OUT)) {
             try {
                 Thread.sleep(Long.MAX_VALUE);


dpc-api/src/test/resources/test.application.yml
```diff
@@ -94,6 +94,7 @@ bbclient:
     socketTimeout: 5000 # ms
     requestTimeout: 5000 # ms
   serverBaseUrl: ${BFD_URL}
+  useBfdMock: ${USE_BFD_MOCK:-true}
```
MEspositoE14s, line 97: When run in your IDE, dpc-api was trying to connect to the real BFD sandbox.  Since that connection doesn't work anymore, tests were failing.

dpc-test.sh
```diff
@@ -58,7 +58,7 @@ docker compose -p start-v1-app down
 USE_BFD_MOCK=true docker compose -p start-v1-app up db attribution aggregation --wait
 
 # Run the integration tests
-docker compose -p start-v1-app up --exit-code-from tests tests
+USE_BFD_MOCK=true docker compose -p start-v1-app up --exit-code-from tests tests
```
MEspositoE14s, line 61: dpc-aggregation was already using the mock BFD client when running `make ci-app`, no dpc-api will use it, too.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
```diff
@@ -167,7 +180,7 @@ protected void processJobBatch(JobQueueBatch job) {
             }
 
             // Finish processing the batch
-            if (this.isRunning()) {
+            if (Boolean.TRUE.equals(this.isRunning())) {
```
jdettmannnava, line 183: Is this necessary?

dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/client/MockBlueButtonClient.java
```diff
@@ -76,6 +82,15 @@ public MockBlueButtonClient(FhirContext fhirContext) {
 
     @Override
     public Bundle requestPatientFromServerByMbi(String mbi, Map<String, String> headers) throws ResourceNotFoundException {
+        // If the TEST_PATIENT_TIME_OUT is requested, hang forever.  SonarQube will probably flag this, but since we're
```
jdettmannnava, line 85: Maybe use this: https://stackoverflow.com/questions/10971968/turning-sonar-off-for-certain-code

jdettmannnava (approved):
LGTM.
FWIW, I have just been going into SonarQube and calling it a false positive, since it seems intermittent.

dpc-aggregation/src/main/java/gov/cms/dpc/aggregation/engine/AggregationEngine.java
jdettmannnava: But this didn't change? Or was there a bug?

dpc-bluebutton/src/main/java/gov/cms/dpc/bluebutton/client/MockBlueButtonClient.java
jdettmannnava: It didn't flag it *this* time

jdettmannnava (approved):
LGTM even with the changes!

PR #2712, opened 2025-07-07
run ci workflow before upgrading packages

lukey-luke:
## 🎫 Ticket

No ticket. Previously this workflow was modified to reduce the number of builds.

## 🛠 Changes

 - include all `Gemfile`'s and `pom.xml`'s

## ℹ️ Context

  - when automated PR's upgrade packages (e.g. snyk), it's helpful to see the build succeed to ensure breaking changes are not introduced in the package upgrades

## 🧪 Validation

n/a

### DIFF
--- a/.github/workflows/ci-workflow-gf.yml
+++ b/.github/workflows/ci-workflow-gf.yml
@@ -19,6 +19,7 @@ on:
       - dpc-testing/**
       - dpc-web/**
       - engines/**
+      - pom.xml
       - '**/pom.xml'
       - '**/Gemfile'
       - src/**


.github/workflows/ci-workflow-gf.yml
```diff
@@ -19,7 +19,8 @@ on:
       - dpc-testing/**
       - dpc-web/**
       - engines/**
-      - pom.xml
+      - '**/pom.xml'
```
jdettmannnava, line 22: I'm unsure of the value of this. Is there a single pom.xml or Gemfile that is not already covered by the existing patterns?

lukey-luke: I believe this does address an issue w/ one of the snyk PR's. See slack discussion [HERE](https://cmsgov.slack.com/archives/CJL0FRN2E/p1751038961072639).

PR #2714, opened 2025-07-08
Me/dpc 4765 remove gf from workflows

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4765

## 🛠 Changes

- Removed _-gf_ and __gf_ suffixes on workflows.
- Removed _(Greenfield)_ from the names of workflows.

## ℹ️ Context

Legacy is gone, so we no longer have to differentiate between legacy and GF workflows.

## 🧪 Validation

Due to the name changes, these are all "new" workflows as far as Github is concerned, so I can't run them until after this is merged and they are created.  Once that's done, I'll do a full deploy just to be sure things are working, but this is about as minor a change as there can be, so we're probably ok.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# Default owners for everything
+ *  @CMSgov/dpc-dev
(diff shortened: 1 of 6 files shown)


ashley-weaver (approved):
LGTM

ashley-weaver (approved):

PR #2723, opened 2025-07-15
[DPC-4420] add pagination to GET /v1/Patient

lukey-luke:
## 🎫 Ticket

[DPC-4420](https://jira.cms.gov/browse/DPC-4420)

## 🛠 Changes

 - added optional query parameters: `_count` and `_page`
 - these are used to paginate results of `patientSearch()`

## ℹ️ Context

 - this is a continuation from PoC that was conducted in [DPC-4175](https://jira.cms.gov/browse/DPC-4175)
 - This aims to address an issue where customers with too many patients to return will have this request fail and cause a degradation of service.
   - Legacy behavior is still enabled, meaning the full roster will be output when `_page` is not provided

## 🧪 Validation

 - Unit tests in `PatientResourceUnitTest.java`

### DIFF
--- a/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
@@ -16,6 +16,7 @@ import gov.cms.dpc.api.auth.annotations.PathAuthorizer;
 import gov.cms.dpc.api.resources.AbstractPatientResource;
 import gov.cms.dpc.bluebutton.client.BlueButtonClient;
 import gov.cms.dpc.common.annotations.NoHtml;
+import gov.cms.dpc.common.utils.PagingUtils;
 import gov.cms.dpc.fhir.DPCIdentifierSystem;
 import gov.cms.dpc.fhir.DPCResourceType;
 import gov.cms.dpc.fhir.FHIRExtractors;
@@ -85,16 +86,20 @@ public class PatientResource extends AbstractPatientResource {
     public Bundle patientSearch(@ApiParam(hidden = true)
                                 @Auth OrganizationPrincipal organization,
                                 @ApiParam(value = "Patient MBI")
-                                @QueryParam(value = Patient.SP_IDENTIFIER) @NoHtml String patientMBI) {
+                                @QueryParam(value = Patient.SP_IDENTIFIER) @NoHtml String patientMBI,
+                                @ApiParam(value = "Patients per page")
+                                @QueryParam(value = "_limit") int limit,
+                                @ApiParam(value = "Page number")
+                                @QueryParam(value = "_page") int page) {
 
-        final var request = this.client
+        var request = this.client
                 .search()
                 .forResource(Patient.class)
                 .encodedJson()
                 .where(Patient.ORGANIZATION.hasId(organization.getOrganization().getId()))
                 .returnBundle(Bundle.class);
 
-        if (patientMBI != null && !patientMBI.equals("")) {
+        if (patientMBI != null && !patientMBI.isEmpty()) {
 
             // Handle MBI parsing
             // This should come out as part of DPC-432
@@ -104,12 +109,18 @@ public class PatientResource extends AbstractPatientResource {
             } else {
                 expandedMBI = String.format("%s|%s", DPCIdentifierSystem.MBI.getSystem(), patientMBI);
             }
-            return request
-                    .where(Patient.IDENTIFIER.exactly().identifier(expandedMBI))
-                    .execute();
+            request = request.where(Patient.IDENTIFIER.exactly().identifier(expandedMBI));
         }
 
-        return request.execute();
+        return PagingUtils.handlePaging(request, limit, page, "/v1/Patient");
+    }
+
+    public Bundle patientSearch(OrganizationPrincipal organization, String patientMBI, int page) {
+        return patientSearch(organization, patientMBI, PagingUtils.defaultLimit, page);
+    }
+
+    public Bundle patientSearch(OrganizationPrincipal organization, String patientMBI) {
+        return patientSearch(organization, patientMBI, PagingUtils.defaultLimit, 1);
     }
 
     @FHIR
--- a/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
@@ -121,7 +121,7 @@ public class PatientResource extends AbstractPatientResource {
     }
 
     public Bundle patientSearch(OrganizationPrincipal organization, String patientMBI, int page) {
-        return patientSearch(organization, patientMBI, PagingUtils.DEFAULT_LIMIT, page);
+        return patientSearch(organization, patientMBI, PagingUtils.DEFAULT_COUNT, page);
     }
 
     // passing -1 means "do not paginate" for compatiblity reasons. See default value above
--- a/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
@@ -88,7 +88,7 @@ public class PatientResource extends AbstractPatientResource {
                                 @ApiParam(value = "Patient MBI")
                                 @QueryParam(value = Patient.SP_IDENTIFIER) @NoHtml String patientMBI,
                                 @ApiParam(value = "Patients per page")
-                                @QueryParam(value = "_limit") int limit,
+                                @QueryParam(value = "_count") int count,
                                 @ApiParam(value = "Page number")
                                 @QueryParam(value = "_page") @DefaultValue("-1") int page) {
 
@@ -113,7 +113,7 @@ public class PatientResource extends AbstractPatientResource {
         }
 
         if (page >= 1) {
-            return PagingUtils.handlePaging(request, limit, page, "/v1/Patient");
+            return PagingUtils.handlePaging(request, count, page, "/v1/Patient");
         }
         else {
             return request.execute(); // deprecated - legacy behavior for clients relying on full roster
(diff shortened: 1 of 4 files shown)


dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/PatientResourceUnitTest.java
```diff
@@ -120,6 +122,190 @@ public void testPatientSearchNoIdentifier() {
...
+        assertNull(emptyBundle.getLink("next"));
+        assertEquals(emptyBundle.getLink("self").getUrl(), requestPath + hugePageNumber);
+        assertNotNull(actualResponse);
+        assertTrue(actualResponse.getEntry().isEmpty(), "Expected no entries for an out-of-bounds page");
+
+        assertEquals(1, actualResponse.getTotal());
+    }
+
+    @Test
+    public void testPageNumEqualsZero() {
```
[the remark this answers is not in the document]
lukey-luke: apologies, this is incorrect. Count=0 should be the same as summary account. Link to FHIR docs [HERE](https://build.fhir.org/search.html#_count)

> If _count has the value 0 (zero), this SHALL be treated the same as _summary=count

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
```diff
@@ -104,12 +109,24 @@ public Bundle patientSearch(@ApiParam(hidden = true)
...
-        return request.execute();
+        if (page >= 0) {
+            return PagingUtils.handlePaging(request, count, page, "/v1/Patient");
+        }
+        else {
+            return request.execute(); // deprecated - legacy behavior for clients relying on full roster
+        }
+    }
+
+    public Bundle patientSearch(OrganizationPrincipal organization, String patientMBI, int page) {
```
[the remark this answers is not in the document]
lukey-luke: These overloaded versions were helpers for unit testing. There's no reason for them to live here. Moved to `PatientResourceUnitTest.java`

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
```diff
@@ -110,12 +102,42 @@ public Bundle patientSearch(@ApiParam(hidden = true)
...
+    @Override
+    public Bundle patientSearch(@ApiParam(hidden = true)
+                                @Auth OrganizationPrincipal organization,
+                                @ApiParam(value = "Patient MBI")
+                                @QueryParam(value = Patient.SP_IDENTIFIER) @NoHtml String patientMBI,
+                                @ApiParam(value = "Patients per page")
+                                @QueryParam(value = "_count") Integer count,
+                                @ApiParam(value = "Page number") // null means "do not paginate" for compatibility reasons
+                                @QueryParam(value = "_page") Integer page) {
+        var request = this.buildPatientSearchQuery(organization.getOrganization().getId(), patientMBI);
```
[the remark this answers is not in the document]
jdettmannnava: According to the spec, we can use anything we want to make forward and back work. I think using _offset would work best, as it is what attribution expects.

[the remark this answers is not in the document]
jdettmannnava: You are absolutely correct that the _page logic would have met the spec of the ticket, so good on that! But it kind of fights with _count, which I think is better. @MEspositoE14s , do you have thoughts?

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
```diff
@@ -110,12 +102,42 @@ public Bundle patientSearch(@ApiParam(hidden = true)
...
+        if (count == null && page != null) {
+            count = this.defaultPageSize;
+        }
+
+        if (count != null && count == 0) {
+            request.count(0).summaryMode(SummaryEnum.COUNT); // count gets omitted in request
+            return request.execute();
+        }
+        else if (count !=null && page != null && page >= 1) {
```
MEspositoE14s, line 136: If they pass us a negative `_count` or `_page` can/should we throw an error and return a 4xx?

MEspositoE14s (commented):
Can you add integration tests at the API level that prove everything works together the way we're expecting? 

1. One happy path with paging.
2. One where they ask for a summary.
3. One where they send an invalid count/page.
4. One where patients are added/deleted in between pages being downloaded.
5. Maybe one where there's exactly one page of patients to be sure it handles that edge case.

lukey-luke:
Note: after further inspection, it looks like an absolute URL path is required in order to retain compatibility with `.next()` calls in PatientResource test e.g. `Bundle secondPage = client.loadPage().next(firstPage).execute();` [HERE](https://github.com/CMSgov/dpc-app/blob/55fcc7129bf8ba2c31de62d11e97a026ffb45a9a/dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/PatientResourceTest.java#L696).

I added a new provider to attribution in order to get the dpc-api url from attribution where links are generated. `String proveBaseURL()` [HERE](https://github.com/CMSgov/dpc-app/blob/bee4c1b7bad966706b8236472f2708570addbfc0/dpc-attribution/src/main/java/gov/cms/dpc/attribution/AttributionAppModule.java#L78C1-L79C1)


I resolved all the threads related to page count as we're now using `_offset` as discussed above.

lukey-luke:
I've updated existing tests to reflect use of `_offset` instead of a page number. I've also added additional tests for `PatientDAO` and integration tests in `dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/PatientResourceTest.java`.

When you have time, please have another look and let me know what you think! @jdettmannnava @MEspositoE14s !!

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
```diff
@@ -336,4 +339,26 @@ private static Optional<WebApplicationException> validateAndAddOrg(Patient patie
...
+
+    private IQuery<Bundle> buildPatientSearchQuery(String orgId, @Nullable String patientMBI) {
+        IQuery<Bundle> query = this.client
+                .search()
+                .forResource(Patient.class)
+                .encodedJson()
+                .where(Patient.ORGANIZATION.hasId(orgId))
+                .returnBundle(Bundle.class);
+        if (patientMBI != null && !patientMBI.isEmpty()) {
+            // Handle MBI parsing
```
jdettmannnava, line 351: I think these comments can disappear. DPC-432 no longer exists.

dpc-attribution/src/main/java/gov/cms/dpc/attribution/resources/v1/PatientResource.java
```diff
@@ -33,44 +37,71 @@ public class PatientResource extends AbstractPatientResource {
...
+            daoSearchQuery.setPatientMBI(patientIdentifier.getValue());
+        }
+
+        if (summary != null && summary.equals("count")) {
+            daoSearchQuery.setCount(0);
+            int totalPatients = this.dao.countMatchingPatients(daoSearchQuery);
+            return this.pagingService.convertToSummaryBundle(totalPatients);
+        }
+        if (count == null) {
+            // Legacy behavior - before _page parameter was introduced
```
jdettmannnava, line 85: _count

dpc-attribution/src/main/resources/application.yml
```diff
@@ -38,6 +38,9 @@ server:
             version: ${APP_VERSION:-"unknown_version"}
             application: ${APPLICATION:-"dpc-attribution"}
 
+# The root URL at which the application is accessible, if necessary, include the port, do not include the application version
+publicURL: ${PUBLIC_URL:-"http://localhost:3002"}
```
jdettmannnava, line 42: Will need dpc-ops update

jdettmannnava (commented):
Should we be including total in the paged response?

lukey-luke:
> Should we be including total in the paged response?

@jdettmannnava I specifically avoided included it in the response, because AFAIK this would require an additional database query. My thinking was that given the nature of the issue, focus on performance. If there's a client demand and we can make the concession, definitely possible

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
```diff
@@ -336,4 +344,46 @@ private static Optional<WebApplicationException> validateAndAddOrg(Patient patie
...
+                expandedMBI = String.format("%s|%s", DPCIdentifierSystem.MBI.getSystem(), patientMBI);
+            }
+            query = query.where(Patient.IDENTIFIER.exactly().identifier(expandedMBI));
+        }
+
+        return query;
+    }
+
+    private void handlePagingLinkPathing(Bundle bundle) {
+        if (bundle == null || bundle.getLink() == null) {
```
MEspositoE14s, line 369:
```suggestion
        if (bundle == null || bundle.hasLink() {
```

HAPI has a bunch of methods like this you can use to simplify null and empty checks.

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
```diff
@@ -336,4 +344,46 @@ private static Optional<WebApplicationException> validateAndAddOrg(Patient patie
...
+            return;
+        }
+
+        for (Bundle.BundleLinkComponent link : bundle.getLink()) {
+            String relation = link.getRelation();
+            if (relation == null) {
```
MEspositoE14s, line 375: Same as above:

```suggestion
            if (link.hasRelation()) {
```

dpc-attribution/src/main/resources/application.yml
lukey-luke: The domain part of the URL is now handled in [handlePagingLinkPathing() in dpc-api](https://github.com/CMSgov/dpc-app/blob/a91e0bcd60eace4dfb67c123fc6e9e3cbd3db78f/dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java#L368), removed!

dpc-api/src/main/java/gov/cms/dpc/api/resources/v1/PatientResource.java
lukey-luke: updated to use `!bundle.hasLink()`

MEspositoE14s (approved):
In hindsight, we probably should've broken this up into multiple tickets, but good job getting through it all!

jdettmannnava (approved):
LGTM

PR #2727, opened 2025-07-17
Address dependabot alerts

lukey-luke:
## 🎫 Ticket

No Ticket. This aims to address the following dependabot alerts: 

1. https://github.com/CMSgov/dpc-app/security/dependabot/633
2. https://github.com/CMSgov/dpc-app/security/dependabot/698
3. https://github.com/CMSgov/dpc-app/security/dependabot/699

## 🛠 Changes

 - checked in a new version of dpc-admin/package-lock.json that reflects the current packages when running `npm install`

## ℹ️ Context

 - dependabot is configured to scan for vulnerabilities. Find exhaustive list here: https://github.com/CMSgov/dpc-app/security/dependabot?q=is%3Aopen+severity%3Acritical

## 🧪 Validation

 - ran tests locally and the make commands still work
 - 
Screenshot 2025-07-17 at 10 49 51 AM

 - pipelines are passing EXCEPT snyk, which seems to be failing on `dpc-admin/package.json` for an unrelated issue


jdettmannnava (commented):
This doesn't seem to update the package-lock.json, so I think we will keep getting the security error.
Running locally (even on main), the package-lock.json file produced in dpc-admin is wildly different from the one currently in the repo. It doesn't even have elliptic as a dependency.
So maybe just get the accurate package-lock.json in the repo, and see where we are security-wise after that?
The **best** solution would update package-lock.json on the file system when it changes as part of the build process.

lukey-luke:
> This doesn't seem to update the package-lock.json, so I think we will keep getting the security error. Running locally (even on main), the package-lock.json file produced in dpc-admin is wildly different from the one currently in the repo. It doesn't even have elliptic as a dependency. So maybe just get the accurate package-lock.json in the repo, and see where we are security-wise after that? The **best** solution would update package-lock.json on the file system when it changes as part of the build process.

@jdettmannnava Good catch. Looks like this file hasn't been updated in over a year. I reinstalled the dpc-admin node dependencies.
```
rm -rf node_modules package-lock.json
npm install
```
It looks like the new package-lock.json should address the other 2 critical vulnerabilities too 

1. https://github.com/CMSgov/dpc-app/security/dependabot/699
2. https://github.com/CMSgov/dpc-app/security/dependabot/698

jdettmannnava (dismissed):
LGTM. You'll probably want to update the description for your second reviewer ;)

PR #2728, opened 2025-07-17
bugfixing snyk failure

lukey-luke:
## 🎫 Ticket

Don't review this yet. I changed package.json in a different PR and even after removing those changes, snyk seems to fail on that file. Running again here to see if this runs into snyk issues


lukey-luke:
Screenshot 2025-07-17 at 1 08 19 PM
still results in the same snyk failure, closing

PR #2746, opened 2025-07-23
Portal works even if ALLOW_INVALID_SSL_CERT not set

jdettmannnava:
## 🎫 Ticket

No ticket

## 🛠 Changes

Updated fetch of ALLOW_INVALID_SSL_CERT to be false unless set

## ℹ️ Context

Running local was failing because ALLOW_INVALID_SSL_CERT was not set. We could always force it to be set, or assume false if it wasn't set. The latter made more sense to me.

## 🧪 Validation

Ran dpc-portal locally with the new client and successfully added a public key.

### DIFF
--- a/engines/api_client/README.md
+++ b/engines/api_client/README.md
@@ -33,7 +33,7 @@ Run the tests until they pass
 $ make test
 ```
 
-Jumb into the docker shell for iterative development
+Jump into the docker shell for iterative development
 ```
 make run
 ```
--- a/engines/api_client/README.md
+++ b/engines/api_client/README.md
@@ -25,11 +25,15 @@ Build the docker image
 
 In the api_client directory
 ```bash
-$ docker build . -t api_client
+$ make build
 ```
 
 Run the tests until they pass
 ```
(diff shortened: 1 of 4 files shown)


engines/api_client/README.md
```diff
@@ -25,11 +25,15 @@ Build the docker image
...
 ```
 
 Run the tests until they pass
 ```
-$ docker run --rm -v ${PWD}:/api-client -w /api-client api_client bundle exec rspec
-$ docker run --rm -v ${PWD}:/api-client -w /api-client api_client bundle exec rubocop
+$ make test
 ```
+
+Jumb into the docker shell for iterative development
```
ashley-weaver, line 36: jump?

lukey-luke (approved):
Thank you for submitting a fix!

I was able to confirm that I could instantiate the DpcClient locally after the fallback value was set. 
Screenshot 2025-07-23 at 2 17 43 PM

This unblocks create public key being broken locally!

ashley-weaver (approved):
🙌

PR #2747, opened 2025-07-25
[DPC-4812] Moved foreman to Docker image

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4812

## 🛠 Changes

Moved Foreman gem into the portal Docker image.

## ℹ️ Context

We had been installing Foreman on a per environment basis when the portal starts up, but as part of https://github.com/CMSgov/dpc-ops/pull/853 we won't be able to write to our containers anymore.  That means we can no longer install anything, so we're adding Foreman to the Docker image ahead of time instead.  The trade off is it'll get installed regardless of whether we actually intend to use it, but I don't think that's a big deal.

## 🧪 Validation

Full build and deploy with passing smoke tests [https://github.com/CMSgov/dpc-app/actions/runs/16526473477](https://github.com/CMSgov/dpc-app/compare/main...me/here).

### DIFF
--- a/dpc-portal/Dockerfile
+++ b/dpc-portal/Dockerfile
@@ -39,6 +39,9 @@ RUN rm -rf /bin/yarn
 RUN npm run build:css
 RUN RAILS_ENV=production SECRET_KEY_BASE=dummy bundle exec rake assets:precompile --trace
 
+# Install foreman
+RUN gem install foreman
+
 # Clean up from the build
 RUN rm -rf /usr/local/bundle/cache/*.gem && \
     find /usr/local/bundle/gems/ -name "*.c" -delete && \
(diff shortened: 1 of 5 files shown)


dpc-portal/Dockerfile
```diff
@@ -39,6 +39,9 @@ RUN rm -rf /bin/yarn
 RUN npm run build:css
 RUN RAILS_ENV=production SECRET_KEY_BASE=dummy bundle exec rake assets:precompile --trace
 
+# Install foreman
```
jdettmannnava, line 42: Can we move this above the audit (line 32)

MEspositoE14s: In hindsight, that makes total sense.

jdettmannnava: To be fair, your code replicated what was being done

lukey-luke (approved):
This looks good to me! I was curious if we should lock down a version for foreman here, but it doesn't look like this package is being updated (latest release in 2020).

PR #2755, opened 2025-08-05
Quick release workflow

jdettmannnava:
## 🎫 Ticket

No ticket

## 🛠 Changes

Added quick-release.yml to github actions.

## ℹ️ Context

Sometimes developers need to test code iteratively on AWS, and waiting for a complete build cycle is extremely time consuming. This workflow builds all the apps without testing them, and it deploys only to dev or test. Images created by these releases are not pushed to the production account.

## 🧪 Validation

Successfully pushed with temp caller.

### DIFF
--- a/.github/workflows/quick-release.yml
+++ b/.github/workflows/quick-release.yml
@@ -1,0 +1,249 @@
+name: 'DPC - Quick Release'
+
+on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


ashley-weaver:
do you have a failing run when env is set to something other than test or dev?

PR #2756, opened 2025-08-12
Me/dpc 4817 use iam for db access

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4817

## 🛠 Changes

- In the test environment only, use IAM authentication to get to the DB in dpc-api.
- Remove old prod-sbx env files.

## ℹ️ Context

For security reasons, we're switching to using IAM auth to access the new RDS.  This is currently setup in the test environment, but not the others, so they still need to use the older  method.

### Note:

When we're ready to port the DB changes to our other environments, it should just be a matter of updating each services ENV.application.env file.

## 🧪 Validation

Full deploy and smoke test to dev (old auth): https://github.com/CMSgov/dpc-app/actions/runs/16914573407

Full deploy and smoke test to test (new auth): https://github.com/CMSgov/dpc-app/actions/runs/16914587330 (Smoke test passed the [second](https://github.com/CMSgov/dpc-app/actions/runs/16916276852) time)

Ran full build locally and all services can still connect to local DB.

### DIFF
--- a/dpc-portal/app/services/public_key_manager.rb
+++ b/dpc-portal/app/services/public_key_manager.rb
@@ -4,6 +4,8 @@
 class PublicKeyManager
   include CredentialManager
 
+  INVALID_KEY = 'Invalid public key.'
+
   def create_public_key(public_key:, label:, snippet_signature:)
     public_key = strip_carriage_returns(public_key)
     snippet_signature = strip_carriage_returns(snippet_signature)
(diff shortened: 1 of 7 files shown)


jdettmannnava (approved):
Looks good! I noticed you added the DB_DRIVER_CLASS to the local env files (except for consent), but not the other environments using the default. I guessed that it was so we wouldn't flip the default to aws and forget that local still needs postgresql, and that you hoped dpc-consent would be gone before that?

PR #2763, opened 2025-08-18
DPC-4862 fix alert background

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4862

## 🛠 Changes

Alert component case statement switches on both symbols and strings.

## ℹ️ Context

The flash attribute seems to have changed the keys from symbols (or whatever entered) to strings, so our conversion of alert status to appropriate uswds alert css class failed.
Unsure whether to switch on key.to_s or to include both symbol and string in case statement, but went with the latter.

## 🧪 Validation

Manual check:
alert


ashley-weaver (approved):
👍

PR #2771, opened 2025-09-04
Me/dpc 4872 remove attribution db from api

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4872

## 🛠 Changes

- Removed the attribution db from dpc-api, as it wasn't used.
- Update `@UnitOfWork` annotation to specify which DB to use in dpc-api.

## ℹ️ Context

We noticed that dpc-api had a connection to the attribution DB, but wasn't using it, and when we tried to remove it dpc-api started failing.

The issue was that our endpoints in dpc-api rely on the `@UnitOfWork` annotation to open and close hibernate transactions.  In a DropWizard service that uses multiple DBs, though, you have to specify which DB the annotation should open those transactions on, and we weren't.  In that case, it defaults to looking for a DB named "hibernate".  Coincidentally, the attribution DB that we weren't actually using was named "hibernate", so all transactions were being opened there even though the actual work was being done in the auth and queue DBs.  Fortunately this never caused any problems, most likely because the endpoints that hit the auth DB aren't in heavy usage.

This also meant that when the attribution DB was removed we started getting errors in dpc-api saying it couldn't find the "hibernate" bundle and tests wouldn't pass.

## 🧪 Validation

All integration and unit tests are passing now that the attribution DB has removed.

### DIFF
--- a/dpc-api/src/main/java/gov/cms/dpc/api/DPCAPIConfiguration.java
+++ b/dpc-api/src/main/java/gov/cms/dpc/api/DPCAPIConfiguration.java
@@ -4,7 +4,6 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import gov.cms.dpc.bluebutton.config.BBClientConfiguration;
 import gov.cms.dpc.bluebutton.config.BlueButtonBundleConfiguration;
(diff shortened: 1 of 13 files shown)


jdettmannnava (approved):
LGTM!
To be fair, the UnitOfWork things were _probably_ put in before the attribution service was set up?

PR #2775, opened 2025-09-09
DPC-4818: port postman ci tests to k6

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4818

## 🛠 Changes

- Addition of ci-app.js in dpc-load-testing. Some additions to utils within same directory.
- Removal of postman ci tests, package(-lock).json
- Update of dpc-test.sh to test apis via ci-app.js

## ℹ️ Context

Newman (the node library that runs postman tests) is not compatible with the latest version of npm. This ticket was supposed to change the ci-workflow GHA to ensure that npm was never up to date. Preventing things from being up to date seems like an anti-pattern, and EndToEndRequestTest.postman_collection.json is a giant blob of json that is hard to maintain (it is already missing several api additions), so it seems more sensible to port the tests.

Using k6 seems the appropriate choice, as it has extensive testing capabilities. Furthermore, it uses authentication, so we don't need to restart the api services with auth disabled.

These tests are not necessarily optimal; they are a direct port. The order comes directly from EndToEndRequestTest to make sure are dependencies (side effects) are met. 

### Differences include:

- Removal of consent tests, as consent is going away and these tests would just have to be removed.
- Testing for gzip moved from Request Partial Range to its own test. Newman gzips responses by default, so the EndToEndRequestTest could test this anywhere.
- We do not run tests if preconditions are not met. For example, we don't try to create a group if we can't create patients.

## 🧪 Validation

Tests pass. Tests fail on fail.
- Create Organization -> checkCreateOrganization (in setup())
- Create consent record -> ignored
- Register providers -> checkPractitionerResponse
- Register patients -> checkPatientsResponse
- Submit roster -> checkCreateGroupResponse
- Find patient by MBI -> checkPatientByMbiRequest
- Find provider Group -> checkPractitionerByNpiResponse
- Remove patient from roster -> checkRemovePatientResponse
- Add unknown patient to Roster -> checkAddBadPatient
- Try export after submitting roster -> checkGetGroupExportResponse
- Job response -> checkJobResponse
- Patient data -> checkPatientDataResponse
- Explanation of Benefit data -> checkEobDataResponse
- Coverage data -> checkCoverageDataResponse
- Operation Outcome data -> checkOperationOutcomeDataResponse
- Request Partial Range -> checkPartialEobDataResponse, checkGetOrganizationGzippedResponse
- Request Modified Since -> checkJobResponseWithSince
- Try export with _since -> checkGetGroupExportResponseWithSince
- _since job response -> checkJobResponseWithSince
- All data about a patient ($everything) -> checkPatientEverythingResponse
- Check update invalid content type -> checkUpdateOrganizationResponseBadCT
- Update organization -> checkUpdateOrganizationResponse
- Update consent record -> ignored
- Find Practitioner by NPI -> checkPractitionerByNpiResponse
- Delete patient -> checkDeletePatientResponse
- Patient should be missing from group -> checkGetGroupAfterDeletionResponse
- Delete Practitioner -> checkDeletePractitionerResponse
- Provider group should be missing -> checkGetGroupAfterPractitionerDeleteResponseetGroupResponse


MEspositoE14s (approved):
Everything passes on my end, and is generally just easier to read and follow than the Postman version.

PR #2784, opened 2025-09-18
[DPC-4874] Move pagination helper methods BundleHandler class

lukey-luke:
## 🎫 Ticket

[DPC-4874](https://jira.cms.gov/browse/DPC-4874)

## 🛠 Changes

 - move pagination methods for creating `Bundle`'s to existing BundleHandler class.
 - make `convertToBundle()` more generic to work with other Domain Resource types

## ℹ️ Context

  - This is a follow-up to [DPC-4420](https://jira.cms.gov/browse/DPC-4420) when pagination was introduced.
  - This aims to improve developer experience and keep `Bundle` logic centralized to 1 specific location in the codebase.

## 🧪 Validation

Unit tests passing


dpc-common/src/main/java/gov/cms/dpc/fhir/dropwizard/handlers/BundleHandler.java
```diff
@@ -83,4 +84,20 @@ private Bundle generateBaseBundle(Annotation[] annotations, int entryCount) {
...
+                bundle.addEntry(new Bundle.BundleEntryComponent().setResource(r))
+        );
+        return bundle;
+    }
+
+    public static Bundle convertToSummaryBundle(int total) {
+        Bundle bundle = new Bundle();
+        bundle.setType(Bundle.BundleType.SEARCHSET);
+        bundle.setTotal(total);
+        return bundle;
```
MEspositoE14s, line 101:
```suggestion
 return generateBaseBundle(new Annotation[0], total);
```

You'd have to make `generateBaseBundle` static, too, but might as well DRY things up!

You might be able to do something similar with `convertToBundle` and `writeTo`, but it would probably require pulling out some shared code.  Your call if it's worth the effort.

lukey-luke: Thanks for your feedback! That's a good point! I broke out logic from `writeTo()` into a new method `setBundleEntry()`. Let me know what you think!

MEspositoE14s (approved):
LGTM 👍

PR #2789, opened 2025-09-23
Use partials

ashley-weaver:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4656

## 🛠 Changes

Updates partials to use `@use`.

## ℹ️ Context

The use of `@import` for partials is deprecated.

## 🧪 Validation

Build passing.


jdettmannnava:
Don't we also need changes to dpc-portal?

lukey-luke (approved):
I pulled this down locally and detected no negative side effects from changing the way .scss files are imported.

Screenshot 2025-09-24 at 12 33 52 PM


LGTM!

lukey-luke:
> > > Don't we also need changes to dpc-portal?
> > 
> > 
> > portal appears to already be using `@use`
> 
> I thought we were also supposed to get rid of the require tag @lukey-luke ?

I think require is a javascript thing, like in `dpc-admin/postcss.config.js` [HERE](https://github.com/CMSgov/dpc-app/blob/1b000e2723ce6f9fca5060def6c2d9ce2cd0c2aa/dpc-admin/postcss.config.js#L3). We might be able to switch to `import` across dpc-app repository. I'm not sure, but I drafted a new ticket to look into it: [DPC-4908](https://jira.cms.gov/browse/DPC-4908)

If we're not talking about the same `require` here, please do chime in and let's get on the same page! 😀

PR #2790, opened 2025-09-23
DPC-4905: Use solid queue instead of sidekiq

jdettmannnava:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4905

## 🛠 Changes

- solid_queue added to Gemfiles
- solid_queue configuration files added (queue.yml, recurring.yml)
- solid_queue tables added to db
- puma configured to start solid_queue runners
- sidekiq-cron removed from dpc-portal (along with config - schedule.yml)
- sidekiq-cron schedule.yml ported to recurring.yml (dpc-portal only) 
- Sidekiq removed from routes
- Sidekiqs removed from docker compose
- Deployment no longer checks for sidekiq health
- documentation regarding sidekiq removed

## ℹ️ Context

The overall plan is to eliminate our reliance on Redis. This is stage one, in which we rely on [solid_queue](https://github.com/rails/solid_queue), which is the current [rails default backend](https://guides.rubyonrails.org/active_job_basics.html#default-backend-solid-queue) for async job processing, instead of sidekiq.

We will eventually remove sidekiq, but we are leaving the gems and initialization in until we remove the sidekiq containers from terraform (which depends on this ticket, but this ticket does not depend on that change).

## 🧪 Validation

Running locally (without sidekiq containers), processed mail delivery in each application, ran jobs in dpc-portal and dpc-web, and verified recurring jobs ran in dpc-portal.

Running portal on AWS processed jobs.

### DIFF
--- a/dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/engine/AggregationEngineTest.java
+++ b/dpc-aggregation/src/test/java/gov/cms/dpc/aggregation/engine/AggregationEngineTest.java
@@ -767,7 +767,7 @@ class AggregationEngineTest {
 
         private static List<String> getLogMessages() {
             boolean success = false;
-            int count = 0, maxTries = 50;
+            int count = 0, maxTries =100;
             List<String> logMessages = new ArrayList<>();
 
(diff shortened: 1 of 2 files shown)


MEspositoE14s (approved):
👍

PR #2794, opened 2025-09-30
Me/dpc 4906 increase count

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-4906

## 🛠 Changes

- Increases _count_ parameter when calling BFD to 1000.
- In the case when the call times out, reduces _count_ by half and tries again.
- If the count gets below 100, it treats it as a true timeout and throws an exception.

## ℹ️ Context

We've been downloading FHIR resources from BFD in calls of 100 at a time, and after meeting with the BFD team and doing some [research](https://confluence.cms.gov/spaces/DAPC/pages/1372712944/Count+Parameter+Analysis), discovered that bumping that up to 1000 could get us a 4x performance increase in download speed.  

We wanted to make sure that the increased download size wouldn't lead to timeouts, so we added retry logic that decreased the size after a failed attempt.

## 🧪 Validation

- Added tests against a mock server that tested repeated and one-off timeouts against the `fetchBundle` and `requestNextBundleFromServer` methods.
- Deployed to dev and passed smoke tests [here](https://github.com/CMSgov/dpc-app/actions/runs/18135683033).

### DIFF
--- a/dpc-api/src/test/java/gov/cms/dpc/api/tasks/GenerateClientTokenTests.java
+++ b/dpc-api/src/test/java/gov/cms/dpc/api/tasks/GenerateClientTokenTests.java
@@ -2,7 +2,6 @@ package gov.cms.dpc.api.tasks;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.nitram509.jmacaroons.MacaroonsBuilder;
-import edu.emory.mathcs.backport.java.util.Collections;
 import gov.cms.dpc.api.auth.OrganizationPrincipal;
 import gov.cms.dpc.api.entities.PublicKeyEntity;
 import gov.cms.dpc.api.entities.TokenEntity;
--- a/pom.xml
+++ b/pom.xml
@@ -46,7 +46,7 @@
         <jacoco.version>0.8.12</jacoco.version>
         <jooq.version>3.19.19</jooq.version>
         <bouncey.version>1.81</bouncey.version>
-        <pitest.version>1.5.1</pitest.version>
+        <pitest.version>1.20.3</pitest.version>
         <newrelic.agent.version>8.21.0</newrelic.agent.version>
         <newrelic.agent.type>zip</newrelic.agent.type>
         <h2.version>2.3.232</h2.version>


ashley-weaver:
are we concerned about the backoff strategy leading to timeouts taking longer than before, or is that an acceptable tradeoff for reducing the number of timeouts in general?

dpc-aggregation/src/test/resources/test.application.yml
```diff
@@ -88,7 +88,8 @@ bbclient:
     requestTimeout: 5000 # ms
 
   serverBaseUrl: ${BFD_URL}
-  resourcesCount: 100 # Max number of resource that a request to BB will return before using another request
+  maxResourcesCount: 1000 # Max number of resources that a request to BFD will return before using another request
```
jdettmannnava, line 91: Do we want the same in testing, or should we force it lower?

dpc-bluebutton/src/test/java/gov/cms/dpc/bluebutton/client/BlueButtonClientTest.java
```diff
@@ -301,13 +334,209 @@ private static Config getTestConfig() {
...
+    // Create mock scenario for EoB requests that should time out
+    void createFetchBundleTimeOutScenario() {
+        // All should timeout
+        createMockServerExpectation(
+            "/v1/fhir/ExplanationOfBenefit",
+            HttpStatus.OK_200,
+            "Not needed",
+            List.of(
+                Parameter.param("patient", TEST_EOB_TIMEOUT_PATIENT_ID),
+                Parameter.param("_count", "10"),
```
jdettmannnava, line 438: I wonder if these "10"s (and the ones in the actual tests) should be replaced with a constant? And maybe grab the constant from the config?

dpc-bluebutton/src/test/java/gov/cms/dpc/bluebutton/client/BlueButtonClientTest.java
```diff
@@ -268,23 +288,36 @@ void shouldThrowExceptionWhenResourceNotFound() {
...
+    /**
+     * Helper method that configures the mock server to respond to a given GET request
+     *
+     * @param path          The path segment of the URL that would be received by BlueButton
+     * @param respCode      The desired HTTP response code
+     * @param payload       The data that the mock server should return in response to this GET request
+     * @param qStringParams The query string parameters that must be present to generate this response
+     * @param delay         How long the server should wait to respond.  Useful for testing a timeout.
+     */
+    private static void createMockServerExpectation(String path, int respCode, String payload, List<Parameter> qStringParams, int delay) {
```
jdettmannnava, line 303: This feels a bit unnecessarily precise -- so maybe a "timeout" boolean?

dpc-bluebutton/src/test/java/gov/cms/dpc/bluebutton/client/BlueButtonClientTest.java
```diff
@@ -301,13 +334,209 @@ private static Config getTestConfig() {
...
+        createMockServerExpectation(
+            "/v1/fhir/ExplanationOfBenefit",
+            HttpStatus.OK_200,
+            getRawXML(SAMPLE_EOB_PATH_PREFIX + TEST_EOB_TIMEOUT_PATIENT_ID + ".xml"),
+            List.of(
+                Parameter.param("patient", TEST_EOB_TIMEOUT_PATIENT_ID),
+                Parameter.param("_count", "5"),
+                Parameter.param("excludeSAMHSA", "true")
+            ),
+            1
```
jdettmannnava, line 481: Not sure if this last param is necessary -- whether it makes it easier to read or not.

dpc-aggregation/src/test/resources/test.application.yml
MEspositoE14s: Our mock client should ignore it and return whatever we have hardcoded.  In the case where we're running against the real BFD, though, keeping it the same makes the tests as accurate as possible.  Not a hill I'm willing to die on if you feel strongly about it, though.

jdettmannnava (approved):
LGTM!

PR #2883, opened 2026-01-20
DPC-5127 update portal to use multiple OIDC

jdettmannnava:
NOT FOR MERGE UNTIL CMS APPROVAL

## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5127

## 🛠 Changes

- Devise gems removed
- Login-dot-gov provider renamed from openid_connect to login_dot_gov
- Devise functionality manually restored
- Tests updated

## ℹ️ Context

We are moving to a multi-IDP environment. These IDPs use the OIDC protocol. Because of complexity, the Devise authentication gem only supports one openid_connect provider via the omniauth_openid_connect gem. Fortunately, we use very little Devise functionality, so removing the gem and replacing the necessary parts was not very extensive.

## 🧪 Validation

Automated tests updated to handle a world without devise.

Successfully logged in and out from localhost.
Successfully failed invitation flow.
Successfully passed invitation flow.

## Post-Deploy Actions
After deployment, the following SQL will need to run in dpc-portal. This should not be a problem, as we are not in upper environments.
`UPDATE users set provider = 'login_dot_gov';`


lukey-luke (dismissed):

MEspositoE14s (dismissed):
Makes sense to me!

PR #2896, opened 2026-02-02
DPC-5159 multi csp user POC

jdettmannnava:
Not for merge.

## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5159

## 🛠 Changes

- New model with migration: IdpUid to store foreign keys for CSPs
- Updated login flow to use IdpUid
- Updated user-creation flow in invitations controller to use IdpUid

## ℹ️ Context

We need to support the ability of each user to log in to the portal with multiple CSPs.
Note: because of the way we fake the CPI API Gateway, most Authrorized Officials share the same PacId. Therefore, unlike in production, where each user will have their own PacId, we cannot bind multiple CSPs to the same user by PacId in local, dev, test, and sandbox environments. That is why we use the email address to deduplicate all users in the lower environments. We do want to test this flow, which is why we also bind AOs on PacId while running automated tests.

## 🧪 Validation

Updated Manual tests.
Logged in as same user using multiple IdPs.


MEspositoE14s (approved):
LGTM as far as I understand the way this is supposed to work 👍

ashley-weaver (approved):
Makes sense to me

PR #3004, opened 2026-05-11
[DPC-5384] Me/dpc 5384 dw upgrade

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5384

## 🛠 Changes

Bumped DropWizard from 4.x to 5.0.1 and updated code where necessary.

## ℹ️ Context

- DropWizard 4.x had reached end of life.
- Jetty 11 was accumulating Snyk alerts that are only fixed in 12.x, and Jetty 12 is only supported in DropWizard 5.

## 🧪 Validation

CI passes and services can start and run.


lukey-luke:
Looks good to me! Rails dependencies seem to be causing issues with the pipelines still, but all java workflows are passing ✅

dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/DataResourceTest.java
```diff
@@ -60,7 +60,7 @@ static Stream<Arguments> downloadArgs() {
 	@ParameterizedTest
 	@MethodSource("downloadArgs")
 	void canDownloadFiles(boolean compressFile, Map<String, String> requestHeaders) throws IOException {
-		String testData = "test data".repeat(500);
+		String testData = "test data".repeat(50);
```
MEspositoE14s, line 63: I reduced this because the test is still occasionally failing due to resource limitations.

dpc-api/src/main/resources/application.yml
```diff
@@ -140,7 +140,7 @@ fhir:
   validation:
     enabled: true
     schemaValidation: false
-    schematronValidation: false
+    schematronValidation: false # Schematron has been deprecated, we should remove this in the future.
```
ashley-weaver: do we have a ticket for this?

ashley-weaver (dismissed):
🙌

dpc-api/src/main/resources/application.yml
MEspositoE14s: Just created [DPC-5450](https://jira.cms.gov/browse/DPC-5450).

ashley-weaver (approved):

dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/DataResourceTest.java
Jose-verdance, line 63: Hey @MEspositoE14s, instead of using `.repeat(50)` here, why not just pass an example thats closer to what the data would look like?

dpc-common/src/main/java/gov/cms/dpc/fhir/validations/dropwizard/FHIRValidationModule.java
```diff
@@ -41,11 +54,9 @@ protected void configure() {
         constraintBinder.addBinding().to(ProfileValidator.class);
 
         bind(ConstraintValidatorFactory.class).to(InjectingConstraintValidatorFactory.class);
-        bind(ValidatorFactory.class).toProvider(ValidatorFactoryProvider.class);
         bind(ConfiguredValidator.class).to(InjectingConfiguredValidator.class);
 
         bind(DPCProfileSupport.class).in(Scopes.SINGLETON);
-        bind(FhirValidator.class).toProvider(FHIRValidatorProvider.class);
```
Jose-verdance, line 48: How come we no longer need `ValidatorFactory.class` and `FhirValidator.class`?

dpc-api/src/test/java/gov/cms/dpc/api/resources/v1/DataResourceTest.java
MEspositoE14s: We could have, and I don't feel particularly strongly about it either way so I'm open to switching it, but from the point of view of the DataResource endpoint it doesn't care what's in the file.  It doesn't even attempt to read it, just compress and uncompress it.  I just wanted some junk data and this was an easy way to get it with one line of code.

dpc-common/src/main/java/gov/cms/dpc/fhir/validations/dropwizard/FHIRValidationModule.java
MEspositoE14s: These were replaced with the `@Provides` methods later in the same class.  

- ValidatorFactory provideValidatorFactory(...)
- FhirValidator provideFhirValidator(...)

Doing it this way lets us configure the instances that get bound to `ValidatorFactory` and `FhirValidator`.

Jose-verdance (approved):
LG!

PR #3014, opened 2026-05-22
[DPC-5401] PoC Direct to/from CLEAR for Portal

lukey-luke:
## 🎫 Ticket


dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -79,9 +77,36 @@ def login
     Rails.logger.info(['User began login flow',
                        { actionContext: LoggingConstants::ActionContext::Registration,
                          actionType: LoggingConstants::ActionType::BeginLogin,
-                         invitation: @invitation.id,
-                         **csp_log_context }])
-    csp_login_actions(params[:provider])
+                         invitation: @invitation.id }])
+    claims = {
```
manojwadhwa81, line 81: A claim's nil value vs true value have different meanings in OIDC specifications. true value signals to the CSP that this claim is critical for the application to function. The authorization server should make every effort to return it, and may prompt the user specifically for consent. However, null ( nil in ruby will translate to null ) indicates that the claim is voluntarily requested. If available and consented to, it will be returned; if not, the flow will not break.
We should test the following by setting true instead of nil

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -79,9 +77,36 @@ def login
...
+    }.to_json
+    csp_config = CspConfig.for(:clear)
+    authorization_uri = URI(csp_config.authorization_endpoint)
+    url = URI::HTTPS.build(host: authorization_uri.host,
+                           path: authorization_uri.path,
+                           query: { client_id: csp_config.identifier,
+                                    redirect_uri: "#{my_protocol_host}#{csp_config.redirect_path}",
+                                    response_type: 'code',
+                                    scope: 'openid',
+                                    claims:,
```
manojwadhwa81, line 105: Other than CLEAR, these claims are set at integration level. Can we conditionally merge claims to query hash only if we are redirecting to CLEAR.

dpc-portal/config/csp.yml
```diff
@@ -32,8 +32,8 @@ development: &development
...
     # user_info_endpoint: <%= "https://#{ENV['CLEAR_IDP_HOST']}/integrations/userinfo" %>
-    # user_info_endpoint: <%= "https://#{ENV['CLEAR_IDP_HOST']}/integrations/userinfo" %>?claims={"id_token":{"ssn9":null,"email":null,"email_verified":null},"userinfo":{"ssn9":null,"email":null,"email_verified":null,"given_name":null,"family_name":null}}"
-    user_info_endpoint: "<%= "https://#{ENV['CLEAR_IDP_HOST']}/integrations/userinfo" %>?claims=%7B%22id_token%22%3A%7B%22ssn9%22%3Anull%2C%22email%22%3Anull%2C%22email_verified%22%3Anull%7D%2C%22userinfo%22%3A%7B%22ssn9%22%3Anull%2C%22email%22%3Anull%2C%22email_verified%22%3Anull%2C%22given_name%22%3Anull%2C%22family_name%22%3Anull%7D%7D"
+    # includes URL-encoded claims parameter for specifying required fields to verify
+    user_info_endpoint: "<%= "https://#{ENV['CLEAR_IDP_HOST']}/integrations/userinfo" %>?claims=%7B%22id_token%22%3A%7B%22ssn9%22%3Anull%2C%22email%22%3Anull%2C%22email_verified%22%3Anull%2C%22given_name%22%3Anull%2C%22family_name%22%3Anull%7D%2C%22userinfo%22%3A%7B%22ssn9%22%3Anull%2C%22email%22%3Anull%2C%22email_verified%22%3Anull%2C%22given_name%22%3Anull%2C%22family_name%22%3Anull%7D%7D"
```
manojwadhwa81, line 36: Using the same logic as for the comment in invitations_controller.rb, we might want to pass true instead of null in the user info endpont

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -79,9 +77,36 @@ def login
...
+      },
+      userinfo: {
+        ssn9: nil,
+        email: nil,
+        email_verified: nil,
+        given_name: nil,
+        family_name: nil
+      }
+    }.to_json
+    csp_config = CspConfig.for(:clear)
```
manojwadhwa81, line 97: I guess this is merge issue.. We need to use the parameter value to decide on csp

manojwadhwa81 (commented):
Found some nitty gritty details about claims attribute as used in oidc and sharing relevant comments.

lukey-luke:
This was covered in https://github.com/CMSgov/dpc-app/pull/3058, closing

PR #3018, opened 2026-05-28
Patch gem for id.me. Fix unit, system and integration tests to align with new data model changes.

manojwadhwa81:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5371

## 🛠 Changes

Added new query methods to User model to centralize logic sprinkled elsewhere in code
New helper methods added to LoginSupport
Update test cases to a) Use new helper methods b) Set-up data as needed to synch up with new changes in InvitationController and LoginDotGov Controller
Add email / deactivate email in registration flow.

## ℹ️ Context

This PR covers the area at the intersection of multiple stores e.g. DPC-5371, DPC-5368, DPC-5372.
It updates all the test cases that were previously left untouched due to overlaps of the work done by different team members.

## 🧪 Validation

Updated all the test cases to a) synch with new models and b) test new conditions


dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -100,8 +100,9 @@ def renew
   end
 
   def set_idp_token
```
ashley-weaver: should this method take the csp as a param?

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -211,7 +212,12 @@ def user
     user_info = UserInfoService.new.user_info(session)
     find_or_create_user(user_info)
     csp = Csp.find_by(name: @user.provider)
-    CspUser.find_or_create_by!(user: @user, csp: csp, uuid: user_info['sub'])
+    csp_user = CspUser.find_or_create_by!(user: @user, csp: csp, uuid: user_info['sub'])
+
+    # Update emails based upon the latest information in user info.
+    new_emails = user_info['all_emails'] || user_info['emails'] || user_info['emails_confirmed']
```
ashley-weaver, line 232: can this be a separate method? like

```
def fetch_user_emails(user_info)
    user_info['all_emails'] || user_info['emails'] || user_info['emails_confirmed']
end
```

dpc-portal/app/controllers/login_dot_gov_controller.rb
```diff
@@ -122,17 +123,18 @@ def activate_email(user_email)
   end
 
   def ial_2_actions(user, auth)
-    data = auth.extra.raw_info
-
-    return unless data.ial == 'http://idmanagement.gov/ns/assurance/ial/2'
```
ashley-weaver, line 131: are the only possible values 1 and 2?

manojwadhwa81: Yes, based upon what i know. IAL1 is primarily self attestation, and with the right configuration, we can ask only IAL2 responses to be sent back to us and others will land up in the error queue.

dpc-portal/app/controllers/invitations_controller.rb
manojwadhwa81: Yes, but i'd like to take a step further and normalize responses into one structure e.g DPCUserInfo, so as to centralize translation and consumer access. This would be an extension of the work we discussed in today's call for creating DPCSession/CSPSession.

dpc-portal/app/controllers/application_controller.rb
```diff
@@ -27,13 +24,16 @@ def authenticate_user!
...
-  def sign_in(user)
+  def sign_in(user, csp: 'login_dot_gov')
     session['user'] = user.id
+    session[:csp] = csp.to_s
   end
 
   private
 
   def check_user_verification
+    # puts current_user.inspect
```
MEspositoE14s: SonarQube might flag commented out lines as a code smell.  On the other hand, it might not be smart enough to find them in Ruby, so 🤷

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -78,13 +78,13 @@ def login
...
+    url = URI::HTTPS.build(host: csp_config.host,
+                           path: '/oauth/authorize',
+                           query: { client_id: csp_config.identifier,
+                                    redirect_uri: "#{my_protocol_host}/auth/id_me/callback",
                                     response_type: 'code',
-                                    scope: 'openid email all_emails profile social_security_number',
+                                    scope: 'openid http://idmanagement.gov/ns/assurance/ial/2/aal/2',
                                     nonce: @nonce,
                                     state: @state }.to_query)
     redirect_to url, allow_other_host: true
```
Jose-verdance: @manojwadhwa81 how come we are defaulting this to id.me instead of the existing login.gov?

dpc-portal/app/models/csp_config.rb
```diff
@@ -0,0 +1,56 @@
...
+    case code.to_s
+    when 'login_dot_gov' then LOGIN_DOT_GOV
+    when 'id_me' then ID_ME
+    # when 'clear' then CLEAR
+    else raise ArgumentError, "Unknown CSP code: #{code}"
+    end
+  end
+
+  def self.[](code)
+    from(code)
```
MEspositoE14s: I don't think this is used anywhere so it might be easier to just delete it, but I'm pretty sure this will fail if you call it.
```suggestion
    self.for(code)
```

dpc-portal/app/controllers/invitations_controller.rb
manojwadhwa81: The initial approach taken was to make changes to switch to id.me and enable logging with id.me. As we add new buttons to show in invitation flow, next step will be enhance this to redirect the user to the user selected csp.

dpc-portal/app/controllers/login_dot_gov_controller.rb
MEspositoE14s, line 90: Can we use the new methods in CspUser here?

dpc-portal/app/models/csp_user.rb
```diff
@@ -5,4 +5,39 @@ class CspUser < ApplicationRecord
...
+        UserEmail.create!(csp_user: self, email: new_email, active: true)
+      else
+        # Potentially activate this email
+        activate_email(existing_email)
+      end
+    end
+  end
+
+  def deactivate_old_email(new_emails)
+    # Don't deactivate existing emails if new_emails is empty
```
MEspositoE14s, line 27: new_emails should always include their primary email, right?  I think that's the case with Login.gov, but I don't know about the other two.  If that's the case, it might be worth logging an error.

dpc-portal/config/initializers/omniauth.rb
```diff
@@ -6,30 +6,61 @@
...
+    jwks_uri:      ID_ME_CONFIG[:jwks_uri],
+    userinfo_signed_response_alg: 'RS256',
+    id_token_signed_response_alg: 'RS256'
+  }
+}.freeze
+
+LOGIN_DOT_GOV_CLIENT_CONFIG = {
+  name: :login_dot_gov,
+  issuer: "https://#{LOGIN_DOT_GOV_CONFIG[:host]}/",
+  discovery: false,
```
MEspositoE14s: Is this for performance reasons, or is there something else I'm missing?

dpc-portal/app/services/user_info_service.rb
```diff
@@ -17,14 +15,46 @@ def auth_header(token)
...
-    start_tracking
-    response = Net::HTTP.get_response(USER_INFO_URI, auth_header(token))
+  def parsed_response(response)
+    return if response.body.blank?
+
+    body = response.body.to_s.strip
+    if response.content_type.to_s.strip.downcase == 'application/jwt' || looks_like_jwt?(body)
+      decode_jwt(body)
+    else
+      JSON.parse(body).with_indifferent_access
```
MEspositoE14s, line 40: I didn't know about with_indifferent_access, that's handy!

dpc-portal/spec/system/accessibility_spec.rb
```diff
@@ -320,7 +320,12 @@ def sign_in
               page.fill_in 'invited_email', with: 'john@beatles.com'
               page.fill_in 'invited_email_confirmation', with: 'john@beatles.com'
               page.find_button(value: 'Send invite').click
-              expect(page).to_not have_text("can't be blank")
+              expect(page).to_not have_text(/can't be blank/i)
+              # expect(page).to have_selector('#verify-modal', visible: true, wait: 10)
+
+              # within('#verify-modal') do
+              #   click_button 'Yes, I acknowledge'
+              # end
```
MEspositoE14s, line 328: I've been playing with this test since we talked about it, and I think I've got a fix to turn it back on and get it working.

1.  In `config/environments/test.rb`, remove the line `config.action_controller.asset_host = "file://#{::Rails.root}/public" if ENV['ACCESSIBILITY'] == 'true'`.

2.  In `dpc-portal/docker/accessibility-test.sh` remove the `ACCESSIBILITY=true` bit on line 7.

I think this config was originally put in to help an old version of Axe find our static assets.  Today, though, CORS is preventing us from loading our javascript from a "file://" url and it breaks the test.  If you remove it everything passes.

dpc-portal/config/initializers/omniauth.rb
manojwadhwa81: Nope, while setting config for ID.me, we discovered that few explicit settings like the id_token_signed_response_alg and userinfo_signed_response_alg could only be applied by disabling the discovery endpoint, so had to turn it off.

dpc-portal/app/models/csp_user.rb
manojwadhwa81: That's right, but this logic was added to be on the safer side, as there seems to be a config on csp side that controls the claims we get back, and any temporary misconfig might wipe data from our email table, so put this guard in place after comparing the before and after responses.

dpc-portal/app/controllers/login_dot_gov_controller.rb
manojwadhwa81: Yes, this will be part of refactoring after we stabilize the system and have it deployed in dev/test with core functionality working. Three pieces we need to work upon: a) this duplicate functionality b) duplicates across user_info_service and patch c) Normalized user info across csps d) DPCSession object for higher level apis

dpc-portal/app/controllers/invitations_controller.rb
lukey-luke: This is fixed in my PR to link across all CSP's:
https://github.com/CMSgov/dpc-app/pull/3021

dpc-portal/app/controllers/application_controller.rb
manojwadhwa81: Removed comments.

dpc-portal/app/services/user_info_service.rb
manojwadhwa81: Rails magic!!

Jose-verdance (approved):
LG!

PR #3027, opened 2026-06-10
Me/dpc 5254 dd in java apps

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5254

## 🛠 Changes

- Updates Java services to use APM in Datadog, both locally and in our persisted environments.
- Updates dpc-aggregation to use custom spans in DataDog that contain the job, batch, patient and aggregator IDs.
- Removes New Relic from our Java services.
- Updates maven build caching to no longer scan your entire Maven cache for changes and only look in specific directories.

## ℹ️ Context

We're switching to DataDog from New Relic, and that requires a number of changes.  

- Getting DD to work.
  - We now have to download the dd agent and pass it to Java when we start our services.
  - The DD agent needs to write to the /tmp dir, so we need to make it writeable in our persisted environments.
  - Running locally, we need to spin up the DD agent container in our docker-compose.yml.
  - DD is pretty good about defining spans and traces for apps that handle web requests (dpc-api and dpc-attribution), but not batch processing apps like dpc-aggregation, so we define a new trace around each job batch and a new span around each patient.
  - We don't want to see dozens of traces per minute for every health check and ping to our services, so we filter those out.
- Once we started downloading the DD agent jar, the Maven build cache began choking on it and failing because it's expecting text and the jar is a binary.
  - The build cache and the jib plugin were working together to scan the entire maven cache on our local machines, which was excessive (in my case, 180k files). 
  - Now, it only scans the `src/` and `docker/` directories of each sub-project for changes.  In addition to fixing the DD problem, builds are a lot faster now.

### Note: 
This requires https://github.com/CMSgov/dpc-ops/pull/957 from dpc-ops.

## 🧪 Validation

Full deploy [here](https://github.com/CMSgov/dpc-app/actions/runs/27299732095).

Ran both locally, in dev and in test and verified results:
image


dpc-api/pom.xml
```diff
@@ -311,6 +305,10 @@
                         <environment>
                             <DB_MIGRATION>1</DB_MIGRATION>
                         </environment>
+                        <!-- Writing to /tmp is tricky because it's read only on Fargate.  We need a volume over top of it -->
+                        <volumes>
+                            <volume>/tmp</volume>
+                        </volumes>
```
Jose-verdance, line 311: Hey @MEspositoE14s, what I am a little confused by this comment, what this /tmp being used for exactly?

Jose-verdance (approved):
Just have one clarifying question, looks good otherwise.

dpc-api/pom.xml
MEspositoE14s: The dd-java-agent.jar that we pull in when starting the service writes a few files to `/tmp` and `/tmp/ddprof_nobody`.  Our Fargate containers are read only, though, and our services run under `nobody`, so we have to jump through some hoops to make it work.

For most directories like this we mount a writable drive in the task definition in dpc-ops, then we assign ownership to `nobody` in the entry point script and that makes everything work, but `/tmp` is special.  Since it's a core system dir, Fargate forces the owner to `root` and the `chown` to assign it to `nobody` in the entry point script fails.  Fargate recognizes the` /tmp` coming from the Docker image, though, separates it from the base operating system where it's locked down and lets us take control of it.  

I may not be explaining this perfectly, but the core of it is that Fargate treats `/tmp` different than most other directories so we have to tweak some stuff so the DD agent can write to it.

ashley-weaver (approved):
👍

lukey-luke (approved):

PR #3046, opened 2026-06-30
New abstraction for CSP session data storage AND Use alert component to show email mismatch error in registration flow

manojwadhwa81:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5452
https://jira.cms.gov/browse/DPC-5497

## 🛠 Changes
**DPC-5452**
a) Fix the logout issue and session clean-up as part of redirect call from Csp post logout
b) New alert when email used to sign up doesn't match with the email used for invitation during registration flow
c) Added idempotency to the seeding functionality to avoid duplicated data.
d) Used sequence to generate ids in factory to avoid random test failures
**DPC-5497**
e) Refactored the csp data in session and wrote test cases for the same.

## ℹ️ Context

User can now login using his primary email to complete registration as long as his invitation email matches one of the verified emails. However, when there is no match, use the alert component to display the appropriate error message and provide option to logout of active CSP session.

## 🧪 Validation

Updated system tests to check for the new error message and tested out logout functionality is working as expected.


docker-compose.portals.yml
```diff
@@ -144,7 +144,6 @@ services:
       - CMS_IDM_OAUTH_URL=http://localhost:4567/
       - IDP_LOGIN_DOT_GOV_HOST=idp.int.identitysandbox.gov
       - IDP_ID_ME_HOST=api.idmelabs.com
-      - IDP_ID_ME_CLIENT_ID=925bb2985ccf623114359caa76228919
```
ashley-weaver: does this need to be moved to encrypted/local.env? the ID.me flow is failing locally for me with `'identifier' required`.

manojwadhwa81: Added the same to local.env file. Can you please check now?

dpc-portal/app/components/page/utility/error_component.rb
```diff
@@ -19,6 +19,8 @@ def initialize(invitation, reason, csp: :login_dot_gov)
         @reason = resolve_reason(reason)
         @status = "verification.#{@reason}_status"
         @text = "verification.#{@reason}_text"
+        @show_alert = true if @reason.in?(%i[email_mismatch])
```
ashley-weaver:
```suggestion
        @show_alert = @reason.in?(%i[email_mismatch])
```

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -352,12 +350,7 @@ def check_for_token
   end
 
   def valid_csp_token?
-    csp = session[:csp]
-    return false if csp.blank?
-
-    session["#{csp}_token"].present? &&
-      session["#{csp}_token_exp"].present? &&
-      session["#{csp}_token_exp"] > Time.now
+    csp_session.active?
```
ashley-weaver: `valid_csp_token?` is only used in one place, so you could replace it directly in `check_for_token`

dpc-portal/app/lib/csp_session.rb
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  # Mark an already-known CSP as current without changing its token.
+  def activate(csp)
+    @session[CURRENT_KEY] = csp.to_s
+  end
+
+  # CSP session could result in a user being logged in, so we store the user in the session as well.
+  def store_user(user)
+    @session[USER_KEY] = user
```
ashley-weaver: can you name this param user_id for clarity?

dpc-portal/app/lib/csp_session.rb
```diff
@@ -0,0 +1,121 @@
...
+    data = sessions
+    data[code] = { 'token' => token, 'token_exp' => token_exp }
+    @session[SESSIONS_KEY] = data
+    @session[CURRENT_KEY] = code
+    code
+  end
+
+  # Mark an already-known CSP as current without changing its token.
+  def activate(csp)
+    @session[CURRENT_KEY] = csp.to_s
```
ashley-weaver, line 39: should this return/raise an error if `stored?(csp)` is false?

manojwadhwa81: I think that would be the right thing to do and i intended to do the same thing to avoid inconsistency in this object, but came across one instance where activate(csp) equivalent ( storing csp name in session ) was called even before the callback phase. If you look into the login_session() in InvitationsController, it's being called through POST /organizations/<>/invitations/<>/login endpoint when we don't have a valid csp session and before we redirect user to csp's portal.

.github/workflows/ci-workflow.yml
```diff
@@ -223,7 +227,7 @@ jobs:
       - name: Install docker compose manually
         run: |
           mkdir -p /usr/local/lib/docker/cli-plugins
-          curl -SL https://github.com/docker/compose/releases/download/v2.32.4/docker-compose-linux-aarch64 -o /usr/local/lib/docker/cli-plugins/docker-compose
+          curl -fSL https://github.com/docker/compose/releases/download/v2.32.4/docker-compose-linux-aarch64 -o /usr/local/lib/docker/cli-plugins/docker-compose
```
lukey-luke, line 230: out of scope

dpc-portal/app/controllers/concerns/csp_logout.rb
```diff
@@ -32,7 +32,7 @@ def url_for_id_me_logout
     URI::HTTPS.build(host: csp_config.host,
                      path: csp_config.log_out_path,
                      query: { client_id: csp_config.identifier,
-                              redirect_uri: "#{root_url}oauth/logged_out" }.to_query)
+                              redirect_uri: "#{root_url}auth/logged_out" }.to_query)
```
lukey-luke, line 36: 👍 I noticed this in my branch as well - [HERE](https://github.com/CMSgov/dpc-app/pull/3058/changes#diff-608deffffe38796d2f2554b569d0f528b0e6bc2e6e3cf6e59bb38cf836e1b26dR36)

We should get these changes into `main` and use that to avoid re-introducing conflicts

dpc-portal/app/controllers/users/sessions_controller.rb
```diff
@@ -6,19 +6,24 @@ class SessionsController < ApplicationController
...
     def logged_out
+      clean_session
       redirect_to session.delete(:user_return_to) || sign_in_path
     end
+
+    def clean_session
+      session.delete('user')
+      csp = session.delete(:csp)
+      session.delete("#{csp}_token") if csp
+      session.delete("#{csp}_token_exp") if csp
```
lukey-luke: I'm also storing id token under `{csp}_id_token` - example [HERE](https://github.com/CMSgov/dpc-app/pull/3058/changes#diff-ab3e3fcbe908670b6fc89a121a2c2d34db10596fbb624e13bdc6da7fb00d80d0R10). In my testing, this is required for CLEAR logout to pass.

I'd love if we could just do a full `.reset_session()`, but that would break the "last used csp" feature. Instead, we could greedily call `session.delete()` here. It won't raise an exception.

dpc-portal/spec/components/core/icon/uswds_component_spec.rb
```diff
@@ -13,7 +17,7 @@
     let(:expected_html) do
       <<~HTML
         <svg class="usa-icon" aria-hidden="true">
-          <use xlink:href=/assets/@uswds/uswds/dist/img/sprite-9865eea7b251e43137fb770626d6cd51c474a3a436678a6e66cafce50968076f.svg#lock></use>
+          <use xlink:href=#{sprite_path}#lock></use>
```
lukey-luke, line 20: I believe this is also out-of-scope

ops/config/encrypted/local.env
```diff
@@ -1,113 +1,113 @@
...
+36336163626135363239303435653361663165383064633061303634663533396338363135353235
+30313364666630323837306333386265343062396137623162353731373265356561303237346232
+32363562333735666336313338383334666639633232646338623665303939343338323865326566
+33313732653861353930343735363062376662303332326238373634396630653339356534353166
+31646430373633393038643131616338646566613837353564333361373063343939363032316463
+31653234626664613234646330383864303339313336653930316531333338393439333235333537
+32663961333535303934313762353432373934613139643437346465323064656535646237636161
+32326430313962316333303863333966333332383066333539363731663634643934353230626339
+65633765613331643835643731653732316133313763386136313033623463396162333932663335
+61303330636236366535
```
lukey-luke: confirmed that this includes `IDP_ID_ME_CLIENT_ID` and `docker-compose.portals.yml` should be good 👍 

This will conflict with local.env changes in PR [HERE](https://github.com/CMSgov/dpc-app/pull/3058/changes#diff-d8afde9de66aba1078598f5c5acf43ff19c05556b16cbc8a78a5e2714c2ff5a8)

.github/workflows/ci-workflow.yml
manojwadhwa81: Agreed that it's not covered by the tickets mentioned in PR. I encountered few repetitive CI failures and narrowed it down to this change to fail fast and avoid misleading messages related to 'invalid flags' for docker command. Too little of a change to warrant a new branch followed by merge process.

dpc-portal/spec/components/core/icon/uswds_component_spec.rb
manojwadhwa81: We are clobbering assets before running rspec and this seems to hard-code the digest in spec file. Had to fix this to make build go through successfully.

dpc-portal/app/controllers/users/sessions_controller.rb
manojwadhwa81: I didn't know that some Idps need id token for logout.. let me check out.

manojwadhwa81: @lukey-luke By full reset, do you mean reset in csp_session that will call both clear_user and clear_all?

dpc-portal/app/controllers/users/sessions_controller.rb
```diff
@@ -6,19 +6,24 @@ class SessionsController < ApplicationController
...
+    end
   end
 end
```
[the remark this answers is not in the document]
manojwadhwa81: Done.. Didn't rely upon the suggested legacy_csp logic as we won't have both versions of app running and we don't need two ways to clear sessions. The suggestion is great as it presents a solution for a rolling-update deployment strategy.

docker-compose.portals.yml
manojwadhwa81: @ashley-weaver Marking this as resolved. Feel free to reopen if you face problem

dpc-portal/app/components/page/utility/error_component.html.erb
```diff
@@ -1,10 +1,22 @@
 <div>
-  <h1><%= t(key=@status, csp_display_name: @csp_display_name) %></h1>
-  <div>
-    <p>
-      <%=raw t(key=@text, org_name: @org_name, ao_full_name: @ao_full_name, ao_email: @ao_email, csp_display_name: @csp_display_name) %>
-    </p>
-  </div>
+   <% if @show_alert %>
```
ashley-weaver, line 2: can you update error_component.spec to cover these changes?

dpc-portal/app/components/page/utility/error_component.rb
```diff
@@ -19,6 +19,8 @@ def initialize(invitation, reason, csp: '')
         @reason = AoVerificationService::SERVER_ERRORS.include?(reason) ? :server_error : reason.to_sym
         @status = "verification.#{@reason}_status"
         @text = "verification.#{@reason}_text"
+        @show_alert = @reason.in?(%i[email_mismatch])
+        @alert_type = 'error' if @reason.in?(%i[email_mismatch])
```
ashley-weaver: could you have a private method like 
```
def show_alert?
    @reason.in?(%i[email_mismatch])
end
```
and
```
@show_alert = show_alert?
@alert_type = 'error' if show_alert?
```
then if we ever want to change which reasons show alerts, we only have to update one list.

dpc-portal/app/lib/csp_session.rb
```diff
@@ -0,0 +1,121 @@
...
+  def active?(csp = current)
+    inactive_reason(csp).nil?
+  end
+
+  # Returns a symbol indicating why the CSP session is not active, or nil if it is active.
+  def inactive_reason(csp = current)
+    return :no_session if csp.blank?
+    return :no_token unless token(csp).present?
+    return :no_token_exp unless token_exp(csp).present?
+    return :expired_token unless token_exp(csp) > Time.now
```
ashley-weaver, line 83: is token_exp stored as a string or a Time? if it's a string, I think this would fail for a comparison mismatch.

dpc-portal/app/lib/csp_session.rb
```diff
@@ -0,0 +1,121 @@
...
+  def clear_all
+    @session.delete(SESSIONS_KEY)
+    @session.delete(CURRENT_KEY)
+    nil
+  end
+
+  private
+
+  def sessions
+    @session[SESSIONS_KEY] ||= {}
```
ashley-weaver, line 127: I think we should separate this method into a read and a write, so we're not writing an empty hash into the session every time we check if a CSP is active.

```
def sessions
  @session[SESSIONS_KEY] || {}
end

def sessions!
  @session[SESSIONS_KEY] ||= {}
end
```
and use sessions! in store and clear

ashley-weaver (commented):
a few nitpicks

dpc-portal/spec/factories/users.rb
```diff
@@ -4,6 +4,6 @@
   factory :user, aliases: %i[invited_by] do
     sequence(:uid) { |n| n }
     provider { :id_me }
-    email { "user#{rand(0..100_000)}@example.com" }
+    sequence(:email) { |n| "user#{n}@example.com" }
```
Jose-verdance, line 7: Is this needed? This will be removed as part of my PR.

dpc-portal/spec/lib/csp_session_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+  describe '#activate' do
+    it 'switches the current CSP without changing tokens' do
+      csp_session.store(csp: :id_me, token: 'a', token_exp: future_exp)
+      csp_session.store(csp: :login_dot_gov, token: 'b', token_exp: future_exp)
+      csp_session.activate(:id_me)
+      expect(csp_session.current).to eq 'id_me'
+      expect(csp_session.token).to eq 'a'
+    end
+    # Consider: what if you activate an unstored CSP? Current impl silently accepts.
+    # Add an example that pins the behavior you decide on.
```
Jose-verdance: What is this comment for? This can probably removed.

dpc-portal/spec/factories/users.rb
manojwadhwa81: I have seen sporadic failures where test failures leaves trail behind and subsequent test runs fail. Feel free to remove it when you get rid of email as part of your feature.

dpc-portal/app/lib/csp_session.rb
manojwadhwa81: I think we don't even need session! call when calling clear.. refactored and pushed changes.

dpc-portal/app/lib/csp_session.rb
manojwadhwa81: its going to be Time object, so we are good.. here is how we pass it..
token_exp: auth.credentials.expires_in.seconds.from_now

.github/workflows/ci-workflow.yml
```diff
@@ -151,9 +151,13 @@ jobs:
...
           chown root:root /usr/local/lib/docker/cli-plugins/docker-compose
           chmod +x /usr/local/lib/docker/cli-plugins/docker-compose
+          # Fail fast if the plugin didn't install correctly (e.g. bad download or arch mismatch),
+          # otherwise `docker compose -p ...` silently falls back to `docker` and errors on -p.
+          file /usr/local/lib/docker/cli-plugins/docker-compose
```
MEspositoE14s, line 159: You might as well copy this to the other jobs that have "Install docker compose" steps. Either that, or spin it off into a shell script that can be reused. Probably out of scope, though.

dpc-portal/app/controllers/users/sessions_controller.rb
lukey-luke: By full reset, I'm referring to the session cookie API having a method `reset_session()` - doc [HERE](https://guides.rubyonrails.org/action_controller_overview.html#session).
Should be able to address the CLEAR issue by adding `session.delete("#{csp}_id_token")`

.github/workflows/ci-workflow.yml
manojwadhwa81: Sure.. Will handle it separately as part of the image optimization story.

dpc-portal/app/controllers/users/sessions_controller.rb
manojwadhwa81: I think i get what you are be. Here is how it should workout:
a) id token be stored in csp_session. Optional, as only CLEAR needs it : This will need changes from our current state and can be done as part of the 
b) reset_session should only clear out the session cookie and the last used csp cookie will remain intact.
c) You won't actually need the call to session.delete("#{csp}_id_token"), as it would be too atomic, see the latest version of sessions_controller.rb#clean_session.
Let connect if you have any questions.

lukey-luke (dismissed):

Jose-verdance (dismissed):

dpc-portal/app/controllers/users/sessions_controller.rb
lukey-luke: Yes, the `clean_session()` method is a cleaner approach. I flagged that [on slack ](https://cmsgov.slack.com/archives/CJL0FRN2E/p1783992398057759?thread_ts=1783969967.445689&cid=CJL0FRN2E) and marked PR approved. Please let me know if you need help resolving merge conflicts

Jose-verdance (approved):
LGTM!

lukey-luke (approved):
Verified CLEAR logout functions as intended and all Ansible keys are updated per latest updates! Thank you!

PR #3056, opened 2026-07-07
Ja/dpc-5522 user model update to remove unused columns

Jose-verdance:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5522

## 🛠 Changes

These changes updates existing code and associated tests to reference uuid, email, and provider from the new csp_user and user_email tables. This change also removes encrypted_password, reset_password_token, reset_password_sent_at since we are no longer using these for user login. 

## ℹ️ Context

  
  We created new tables to move CSP specific fields ( e.g. UUID, email ) but kept these in the Users table to ease migration effort. As a result, we needed to clean up the model, update migration script and test cases to sync with the new model changes.
  
 Created the following ticket to address additional changes outside scope of ticket:
 https://jira.cms.gov/browse/DPC-5564
 https://jira.cms.gov/browse/DPC-5566
 https://jira.cms.gov/browse/DPC-5567
https://jira.cms.gov/browse/DPC-5570

## 🧪 Validation

Tested changes and ensured all test pass, make ci-portals, make ci-portal-accessibility, and local smoke-test

Successfully Deployed to Test and confirmed migration was successful along with rollback:
Current branch with migration
https://github.com/CMSgov/dpc-app/actions/runs/30011962491
Main branch after rollback
https://github.com/CMSgov/dpc-app/actions/runs/30034201112


dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -97,6 +97,12 @@ def set_idp_token
 
   private
 
+  def find_user_by_email(email)
+    return nil if email.blank?
+
+    UserEmail.find_by(email:)&.csp_user&.user
+  end
```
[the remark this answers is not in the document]
Jose-verdance: Addressed suggestion and added test for changes.

dpc-portal/app/models/current_attributes.rb
```diff
@@ -22,7 +22,7 @@ def save_user_attributes(user)
 
     CurrentAttributes.current_user = {
       id: user.id,
-      external_id: user.uid,
+      external_id: user.csp_users.first&.uuid,
       pac_id: user.pac_id
     }
```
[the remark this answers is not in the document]
Jose-verdance: Addressed suggestion and added test for changes.

dpc-portal/spec/models/user_spec.rb
```diff
@@ -166,7 +166,7 @@
   describe :audits do
     let(:user) { create(:user) }
     it 'should not audit email' do
-      user.update(email: 'new_email@test.com')
+      # user.update(email: 'new_email@test.com')
       expect(user.audits.count).to eq 0
     end
```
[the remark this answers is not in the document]
Jose-verdance: Addressed

dpc-portal/Dockerfile
```diff
@@ -34,7 +34,10 @@ RUN gem install bundler --no-document && \
 RUN gem install foreman
 
 # Run bundler audit
-RUN bundle exec bundle audit update && bundle exec bundle audit check
+# CVE-2026-38969 is for webrick, which is only used in our fake_cpi_gateway client and never exposed to the outside
+# world.  There's currently no fix, and the recommended solution is to remove or disable it.  That would break
+# the fake_cpi_gateway, though, so we can't do that.
+RUN bundle exec bundle audit update && bundle exec bundle audit check --ignore CVE-2026-38969
```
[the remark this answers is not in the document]
Jose-verdance: This was pulled from main changes to get build working. This will be address in future ticket to keep within scope.

dpc-portal/db/migrate/20260701200404_remove_unused_columns_from_users.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      csp_user = CspUser.find_by(user: user)
+      next if csp_user.nil?
+
+      # Create the primary email record if it doesn't already exist
+      UserEmail.find_or_create_by!(csp_user: csp_user, email: user.email) do |user_email|
+        user_email.active    = true
+        user_email.primary   = true
+      end
+    end
```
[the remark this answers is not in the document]
Jose-verdance: Addressed issue and tested migration

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -260,37 +268,66 @@ def find_or_create_user(user_info)
...
 
-    return matching_users.first if matching_users.present?
+  def find_existing_cd_user(user_info)
+    find_cd_user_by_uuid(user_info) ||
+      find_user_by_email(user_info['email']) ||
+      find_user_by_ssn(user_info)
+  end
 
+  def find_cd_user_by_uuid(user_info)
```
manojwadhwa81: We are finding a user with no restriction on AO or CD user per logic, put the method name seem to indicate otherwise. We could have utilized the existing User.find_by_csp_uid method in User model to reuse what we already have.

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -260,37 +268,66 @@ def find_or_create_user(user_info)
...
+    csp = Csp.find_by(name: session[:csp])
+    CspUser.find_by(uuid: user_info['sub'], csp:)&.user
+  end
+
+  def find_user_by_ssn(user_info)
```
manojwadhwa81: We aren't going to be storing ssn in pac_id field, so this logic seems incorrect to me. I guess we do have something in fake api gateway that returns pac id value as ssn but that's only for testing.

manojwadhwa81: If we end up writing this way, we can make the contract clearer to only pass the attribute from user info object as needed for match.

dpc-portal/app/models/current_attributes.rb
```diff
@@ -60,4 +60,15 @@ def to_log_hash
       organization: CurrentAttributes.organization
     }
   end
+
+  private
+
+  def resolve_external_id(user, csp_name)
+    return user.csp_users.first&.uuid if csp_name.blank?
+
+    user.csp_users
```
manojwadhwa81: csp_user_for in user.rb serve the same purpose. Once we have csp_user for a user we can extract the uuid out of it. Also csp_name should never be blanck

dpc-portal/spec/support/login_support.rb
```diff
@@ -63,9 +74,12 @@ def id_me_auth_hash(user)
   end
 
   def clear_auth_hash(user)
-    all_emails = user.csp_user_for('clear')&.user_emails&.map(&:email).presence || [user.email]
-    { uid: user.csp_user_for('clear')&.uuid || user.uid,
-      info: { email: user.email },
+    csp_user   = fetch_csp_user!(user, 'login_dot_gov')
```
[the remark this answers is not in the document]
Jose-verdance: This will be addressed with additional PR for CLEAR

dpc-portal/app/controllers/invitations_controller.rb
Jose-verdance: Good point! This might be a leftover from multiple refactors lol

dpc-portal/app/controllers/invitations_controller.rb
Jose-verdance: Yeah the fake api gateway that returns pac id value as ssn is what threw me off. Lets sync on this! 🙂

Jose-verdance:
Converting back to draft until https://github.com/CMSgov/dpc-app/pull/3058 and https://github.com/CMSgov/dpc-app/pull/3046 have been merged in to reduce duplicate merge conflict and unit test fixes.

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -257,23 +252,65 @@ def confirmed_email?(user_info)
...
+      find_user_by_email(user_info['email']) ||
+      find_user_by_ssn(user_info)
+  end
+
+  def find_user_by_uuid(user_info)
+    User.find_by_csp_uid(name: csp_session.current, csp_uid: user_info['sub'])
+  end
+
+  def find_user_by_ssn(user_info)
+    ssn = user_info['social_security_number']&.tr('-', '')
```
ashley-weaver: is the SSN stored as 'social_security_number' for each CSP, or does this need to account for `user_info['ssn']` etc. also?

Jose-verdance: ah yes I think it should account for `user_info['ssn']`

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -257,23 +252,65 @@ def confirmed_email?(user_info)
...
   def find_or_create_user(user_info)
-    # Unique PacIds only available in prod
     @user = if @invitation.authorized_official? && (ENV['ENV'] == 'prod' || Rails.env.test?)
               ao_user(user_info)
             else
-              User.find_or_create_by!(email: @invitation.invited_email) do |user_to_create|
-                assign_user_attributes(user_to_create, user_info)
-                log_create_user
-              end
+              find_or_create_new_user(user_info)
```
lukey-luke, line 259: The branching is a little hard to follow, but I believe the first conditional on L256 will fire for AO invites and second will fire for CD's.

Could we rename it to `find_or_create_ao_user()` and `find_or_create_cd_user()`?

dpc-portal/app/models/user.rb
```diff
@@ -55,4 +55,15 @@ def ao?(organization)
   def cd?(organization)
     CdOrgLink.where(user: self, provider_organization: organization, disabled_at: nil).exists?
   end
+
+  # Returns the primary email across all CSPs
+  def email
+    user_emails.find_by(primary: true)&.email ||
+      user_emails.first&.email
```
lukey-luke: I believe this can result in non-deterministic results:
- if user signs up via login.gov with email1@gmail.com, email1 will be stored as primary for login.gov CspUser
- if same user signs in via ID.me and has email2@gmail.com as primary and email1@gmail.com under `"emails_confirmed"` for `/userinfo`, then they will have email2 as primary for ID.me
- the end result means that email1 or email2 could be returned by `def email` here.

Should we add a database constraint for this?

dpc-portal/spec/requests/invitations_spec.rb
```diff
@@ -301,11 +302,12 @@
   end
 
   LoginSupport::CSP_MAP.each do |provider, display_name|
+    # before { skip 'Temporarily disabled' }
     describe "with #{display_name}" do
-      let(:csp) { Csp.find_by(name: provider) || create(:csp, name: provider) }
-      let(:other_csp_name) { LoginSupport::CSP_MAP.keys.reject { |k| k == provider }.sample }
-      let(:other_csp) { Csp.find_by(name: other_csp_name) || create(:csp, name: other_csp_name) }
-      let(:provider_params) { { provider: provider } }
+      let!(:csp) { Csp.find_by(name: provider) || create(:csp, name: provider) }
```
lukey-luke, line 306: I didn't know you could force the value to be eagerly loaded like this - cool! 👍

dpc-portal/app/controllers/application_controller.rb
```diff
@@ -66,7 +66,7 @@ def set_current_request_attributes
     Datadog::Tracing.active_span&.set_tag('usr.id', current_user.id) if current_user
 
     CurrentAttributes.save_request_attributes(request)
-    CurrentAttributes.save_user_attributes(current_user)
+    CurrentAttributes.save_user_attributes(current_user, csp_name: session[:csp])
```
manojwadhwa81: I think this should be refactored to use csp_session.current

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -257,23 +252,65 @@ def confirmed_email?(user_info)
...
+      find_user_by_email(user_info['email'])
+    ].compact.uniq
+  end
+
+  def find_user_by_pac_id
+    User.find_by(pac_id: session[:user_pac_id]) if session[:user_pac_id].present?
+  end
 
+  # Queries through user_emails table, raises on multiple matches
+  def find_user_by_email(email)
```
manojwadhwa81, line 300: Moving this to user.rb file will make this usable across app.

dpc-portal/app/models/user.rb
Jose-verdance: Created the following ticket based on our discussions. https://jira.cms.gov/browse/DPC-5564

lukey-luke (dismissed):

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -258,22 +253,67 @@ def confirmed_email?(user_info)
...
-              end
+              find_or_create_new_user(user_info)
             end
   end
 
-  def ao_user(user_info)
-    matching_users = User.where('email = ? OR pac_id = ?', @invitation.invited_email, session[:user_pac_id])
-    raise MultiUserMatchError, "too many matching users | pac_id: #{session[:user_pac_id]}" if matching_users.size > 1
+  def fake_cpi_gateway?
+    ENV['ENV'] == 'prod' || Rails.env.test?
```
manojwadhwa81: This seems incorrect to me. If we are in prod environment, ENV will be prod, and there won't be any fake_cpi_gateway.

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -258,22 +253,67 @@ def confirmed_email?(user_info)
...
+      find_user_by_ssn(user_info)
+  end
+
+  def find_user_by_uuid(user_info)
+    User.find_by_csp_uid(name: csp_session.current, csp_uid: user_info['sub'])
+  end
+
+  def find_user_by_ssn(user_info)
+    ssn = user_info['social_security_number']&.tr('-', '') || user_info['SSN']&.tr('-', '')
+    User.find_by(pac_id: ssn) if ssn.present?
```
manojwadhwa81: Based upon the if condition in caller method, this method will be executed in the the CD flow in prod. The search condition implies we are storing SSN in pac_id in prod, which is incorrect IMO

dpc-portal/app/models/user.rb
```diff
@@ -55,4 +55,21 @@ def ao?(organization)
...
+    user_emails.find_by(primary: true, active: true)&.email ||
+      user_emails.find_by(active: true)&.email
+  end
+
+  # Returns all emails across all CSPs
+  def all_emails
+    user_emails.map(&:email)
+  end
+
+  def self.find_by_email_in_user_emails(email)
```
manojwadhwa81, line 70: Low priority: We don't need to have _in_user_emails in method name as that's the only way to find user and user_emails as data source will be implicit

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -258,22 +253,67 @@ def confirmed_email?(user_info)
...
+    find_user_by_uuid(user_info) ||
+      find_user_by_email(user_info['email']) ||
+      find_user_by_ssn(user_info)
+  end
+
+  def find_user_by_uuid(user_info)
+    User.find_by_csp_uid(name: csp_session.current, csp_uid: user_info['sub'])
+  end
+
+  def find_user_by_ssn(user_info)
```
manojwadhwa81: This method is essentially find_user_by_pac_id but account for the non prod sceneario where the ssn would be returned as pac_id. But irrespective of whether we are calling the real gateway or the fake gateway, the pac_id returned will be stored in session[:user_pac_id], and if that's the case, we can do the search only on the basis of presence of the same and shouldn't care if it holds a pac_id or a SSN. Going down that path, we'll have the same logic as we already have in find_user_by_pac_id and will execute only for AOs as only AOs will have session[:user_pac_id] ( coz it's set in verify_user_is_ao which is called by confirm action, CDs follow the confirm_cd path )

Comparing the logic with what existed in the main branch before refactoring, the pac_id was utilized only for AO flow and in prod env and rest of the matches relied upon just email match.

dpc-portal/spec/requests/login_dot_gov_spec.rb
```diff
@@ -354,11 +354,12 @@
   end
 
   describe 'CSP inactive' do
+    let!(:csp) { Csp.find_by(name: 'login_dot_gov') || create(:csp, :id_me) }
```
ashley-weaver: nit:
```suggestion
    let!(:csp) { Csp.find_by(name: 'login_dot_gov') || create(:csp, :login_dot_gov) }
```

dpc-portal/app/models/current_attributes.rb
```diff
@@ -17,12 +17,12 @@ def save_request_attributes(request)
     CurrentAttributes.path = request.path
   end
 
-  def save_user_attributes(user)
+  def save_user_attributes(user, csp_name: nil)
```
manojwadhwa81, line 20: Low priority: csp_name should not be defaulted to nil. If a user exists in session, we should also have a corresponding csp as used by user to login. The corresponding test cases where we are testing the behavior when it's nil or non-existent seem to me testing conditions that shouldn't even exist in base code.  Feel free to create a separate ticket to track this.

manojwadhwa81 (approved):

dpc-portal/app/models/current_attributes.rb
lukey-luke: `csp_name` is used to retrieve `CSPUser` - we shouldn't have any instances of `User` without 1+ related `CSPUser` objects as that's tracked across all forms of login. 

- updating this to be non-optional and require a non-nil value would required additional test updates
- probably good idea to scope this to a new ticket, so we can work on findings from this ticket in parallel:
 - https://jira.cms.gov/browse/DPC-5566
 - https://jira.cms.gov/browse/DPC-5567

lukey-luke (approved):
1. Additional controller updates should be made in [DPC-5566](https://jira.cms.gov/browse/DPC-5566).
2. Issues related to `User.email` edge cases are covered in [DPC-5564](https://jira.cms.gov/browse/DPC-5564).
3. Can probably break off a new ticket for any necessary updates to `current_attributes.rb` so all three of these items can be tackled in parallel.

Per original ticket a/c for [DPC-5522](https://jira.cms.gov/browse/DPC-5522), this PR covers the model changes and we can scope new ticket to include the additional test cases (3.)

dpc-portal/app/models/current_attributes.rb
Jose-verdance: Thanks Luke and Manoj. I have created the following ticket to address issue. https://jira.cms.gov/browse/DPC-5570

PR #3065, opened 2026-07-14
Me/dpc 5547 dataresource test fix

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5547

## 🛠 Changes

- Fixed (hopefully) failing `DataResource` integration test.
- Saves `dpc-api` logs as an artifact when running CI in Github Actions.
- Renamed DD services to match names used in persisted environments. 

## ℹ️ Context

The DataResource integration tests have been intermittently failing with a vague error that looks like:

`tests-1  | [INFO] org.opentest4j.AssertionFailedError: expected: <test datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest datatest data> but was: <{"code":500,"message":"There was an error processing your request. It has been logged (ID 8698c184944ea24f)."}>`
(Example [here](https://github.com/CMSgov/dpc-app/actions/runs/29105919061/job/86406326219?pr=3046))

I think the issue might be that after we stage the job and write its test file, the file handle occasionally remains open for a split second and prevents `dpc-api` from reading it.  This new version of the test will attempt to download the file's metadata and wait for it to be successful before proceeding with the rest of the test.

The reason I only "think" this might be the issue is because we don't save the logs from `dpc-api` when running integration tests in CI, so we have no way to verify this.  That's the next thing this PR fixes.  From now on, the `dpc-api` logs are saved as an artifact in each run, so if this happens again we'll be able to determine the true cause.  They're not saved now because of a quirk in the way we run ITs.  We get the logs from the integration test, but not from `dpc-api` since it's running in the DropWizard testing framework instead of its own container.  That's why we get the 500 response above, which came from the IT, but none of the logs that explain what actually caused it, which should come from the `dpc-api` service the IT calls.

Last but not least, this PR renames the DD local services to match the format we're going to be using in our persisted environments.  We go from dpc-{env}-{service} to dpc-{service} and rely on the environment tag for filtering.

## 🧪 Validation

- Full deploy [here](https://github.com/CMSgov/dpc-app/actions/runs/29337957176).
- CI GHA with new artifact [here](https://github.com/CMSgov/dpc-app/actions/runs/29262898977)

image


ashley-weaver (approved):
👍

lukey-luke (approved):
Didn't know about `Awaitility.await()` - sweeeeet!

PR #3068, opened 2026-07-17
Fix error encountered during login flow when using login.gov as CSP

manojwadhwa81:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5561

## 🛠 Changes

Replace literal slash and n characters with a new line character to fix the issue with Login.gov

## ℹ️ Context

Wrong key was causing "Client assertion Signature verification failed" error as the client assertion was getting signed with wrong key due to misrepresentation of new line character in the env variable used to store private key.

## 🧪 Validation
Verified the cause by comparing the length of env variable before and after in and outside of container to ensure it works irrespective of the mechanism used to inject env variables in environment through docker compose's env_file vs bash and account for the differences in handling of \n characters.


Jose-verdance (approved):

MEspositoE14s (approved):
Thanks!

PR #3083, opened 2026-07-27
[DPC-5552] Consolidate duplicated auth unit tests across CSP's

lukey-luke:
## 🎫 Ticket

[DPC-5552](https://jira.cms.gov/browse/DPC-5552)

## 🛠 Changes

 - majority of tests from  `dpc-portal/spec/requests/{csp}.rb` moved to `shared_examples`
 - tests that are agnostic of CSP moved to `csp_spec.rb`

## ℹ️ Context

  - When working on multi-csp effort last sprint, I noticed that we had multiple spec files that were essentially doing the exact same thing. I wanted to make sure that as we expand test coverage, It would cover all of our CSP's whenever possible
  - tests related to `UserEmail` are left as-is, because responses differ across CSP's and only Login and ID.me provide multiple emails
  - config objects used to make it easier to see what parameters are required for auth tests
    - NOTE: all params are the same except `expected_id_token`, because of the additional requirement for CLEAR logout

## 🧪 Validation

 - `bundle exec rspec` locally

### DIFF
--- a/dpc-portal/spec/requests/login_dot_gov_spec.rb
+++ b/dpc-portal/spec/requests/login_dot_gov_spec.rb
@@ -2,58 +2,12 @@
 
 require 'rails_helper'
 require 'securerandom'
+require 'support/csp_auth_shared_examples'
 
 RSpec.describe 'LoginDotGov', type: :request do
   let(:uuid) { SecureRandom.uuid }
   describe 'POST /auth/login_dot_gov' do
     let!(:csp) { Csp.find_by(name: 'login_dot_gov') || create(:csp, :login_dot_gov) }
-    RSpec.shared_examples 'a login.gov client' do
-      context 'user exists' do
-        before do
-          user = create(:user, email: 'bob1@example.com', provider: :login_dot_gov)
-          create(:csp_user, user:, uuid:, csp:)
-        end
-        it 'should sign in a user' do
-          post '/auth/login_dot_gov'
-          follow_redirect!
-          expect(response.location).to eq organizations_url
-          expect(response).to be_redirect
-          follow_redirect!
-          expect(response).to be_ok
-        end
-        it 'should log on successful sign in' do
-          allow(Rails.logger).to receive(:info)
-          expect(Rails.logger).to receive(:info).with(['User logged in',
-                                                       { actionContext: LoggingConstants::ActionContext::Authentication,
-                                                         actionType: LoggingConstants::ActionType::UserLoggedIn,
-                                                         csp: 'login_dot_gov' }])
-          post '/auth/login_dot_gov'
-          follow_redirect!
-        end
-        it 'should write a cookie with the last used csp' do
-          post '/auth/login_dot_gov'
-          follow_redirect!
-          expect(cookies[:last_used_csp]).to eq 'login_dot_gov'
-        end
-        it 'should not add another user credential' do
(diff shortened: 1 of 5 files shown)


dpc-portal/spec/support/csp_auth_shared_examples.rb
```diff
@@ -0,0 +1,240 @@
...
+      # id_token_hint for CLEAR csp only
+      expect(response.location).to include("id_token_hint=#{expected_id_token}") if expected_id_token
+      expect(request.session[:user_return_to]).to be_nil
+    end
+
+    it 'should set return to invitation flow if invitation sent' do
+      invitation = create(:invitation, :ao)
+      delete "/logout?invitation_id=#{invitation.id}"
+      expect(request.session[:user_return_to]).to eq organization_invitation_url(invitation.provider_organization.id,
+                                                                                 invitation.id)
```
lukey-luke, line 238: Note: this is a good thing that we cover, but there's at least one edge case where an error during invite flow doesn't land back on the `organization_invitation_url`

Unit test coverage should be expanded to cover `fail_to_proof` error when this work is completed: [DPC-5569](https://jira.cms.gov/browse/DPC-5569)

lukey-luke:
Flaky test is failing [CI pipeline](https://github.com/CMSgov/dpc-app/actions/runs/30294112902/job/90070740486?pr=3083#step:9:5454). This is a known issue: [DPC-5567](https://jira.cms.gov/browse/DPC-5567)
 - Will try re-running

MEspositoE14s (dismissed):
I like this version a LOT better, those tests were getting a little out of hand.  
This is much more "do not repeat yourself"!

dpc-portal/spec/requests/login_dot_gov_spec.rb
```diff
@@ -2,196 +2,40 @@
...
-      end
-    end
+    login_dot_gov_config = {
+      provider: :login_dot_gov,
+      auth_endpoint: '/auth/login_dot_gov',
+      display_name: 'Login.gov',
+      logout_host: ENV.fetch('IDP_LOGIN_DOT_GOV_HOST'),
+      ial1_auth_response: lambda {
+        {
+          uid: uuid,
```
andresen-nava: non-blocking question: Have you considered using `SecureRandom.uuid` for the uuid in the configs? I think that would allow us to remove the `lambda` in the configs and `instance_exec` in the `csp_auth_shared_examples.rb`. It might make reading and maintaining the configs a bit easier.

Jose-verdance (dismissed):

dpc-portal/spec/requests/login_dot_gov_spec.rb
lukey-luke: Agreed that `lambda` / `instance_exec` is not the most readable!
 - rather than calling `SecureRandom.uuid` twice, I assigned it to a new variable and made the same uid available on the ial1 response
- this is most consistent with the way `auth.uid` will only be unique based on each person [HERE](https://github.com/CMSgov/dpc-app/blob/cde5d8b4d11fd2a63145c5f413826f75553045ed/dpc-portal/app/controllers/csp_controller.rb#L44)

andresen-nava (approved):

ashley-weaver (approved):

PR #3087, opened 2026-08-04
[DPC-5453] Implement flow to add email

ashley-weaver:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5453

## 🛠 Changes

Adds screen to prompt users to add new email to CSP account.

## ℹ️ Context

If a user attempts to sign in with an email that is not on record, we direct them to add the email to their account.

## 🧪 Validation

image


dpc-portal/app/components/page/existing_account/add_email_component.rb
```diff
@@ -0,0 +1,30 @@
...
+      def initialize(orig_email, orig_csp)
+        super()
+        @masked_email = EmailMask.masked(orig_email)
+        @orig_csp_code = orig_csp
+        @orig_csp_display = display_name(orig_csp)
+      end
+
+      private
+
+      def display_name(csp_code)
```
lukey-luke: FYSA - this code will be needed for account linking as well and we should make it reusable across both

dpc-portal/app/controllers/csp_controller.rb
```diff
@@ -43,12 +43,31 @@ def logout
...
+
+  def email_match?(csp_user, auth)
+    csp_user.user_emails.empty? || csp_user.user_emails.map(&:email).include?(primary_email(auth))
+  end
+
+  def render_add_email(email, csp)
+    Rails.logger.info(['User has existing account associated with different email',
+                       { actionContext: LoggingConstants::ActionContext::Authentication,
+                         actionType: LoggingConstants::ActionType::MergeUserAccountEmail,
+                         **csp_log_context }])
```
lukey-luke, line 64: This log message is emitted before the email is "added". This just means that the user was prompted to sign back in with the first email. I think it should probably be emitted after `sync_csp_emails()` fires so that way these logs only indicate when the emails are actually updated in DPC portal

dpc-portal/app/controllers/csp_controller.rb
```diff
@@ -43,12 +43,31 @@ def logout
   def user_actions(auth, csp)
     csp_user = CspUser.find_by(uuid: auth.uid, csp:)
     user = csp_user&.user
+
+    return render_add_email(user.email, csp.name) if user && !email_match?(csp_user, auth)
```
lukey-luke: I'm not able to reproduce this on ID.me with a new test account because it won't let me add additional emails?

I was able to verify that the add email pages shows up successfully by manually editing a `UserEmail` in my database, but keep the `CspUser.uid` the same.

We definitely do track the existence of multiple emails for ID.me and I can see that in the associated `user_info.json` [HERE](https://github.com/CMSgov/dpc-app/blob/05578fd20513773fcf4da8b9530bdb30b61df70d/dpc-portal/spec/fixtures/csps/id_me/user_info.json#L7), but I don't know how the data was generated. Could you please advise on creating an account with multiple emails for testing the email actually getting updated?

Image

dpc-portal/app/controllers/csp_controller.rb
ashley-weaver: should logging be added to the CspEmailSync methods then?

dpc-portal/app/controllers/csp_controller.rb
Jose-verdance: Hey @ashley-weaver, I was able to trigger that new add email page with login.gov, but I am running into an error locally when clicking the "Add new email" button. Happy to sync on what I am seeing.

 
Screenshot 2026-08-05 at 2 22 58 PM

dpc-portal/app/components/page/existing_account/add_email_component_preview.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Page
+  module ExistingAccount
+    # Render the screen to add a new email to an existing account.
+    class AddEmailComponentPreview < ViewComponent::Preview
+      def default
+        org = ProviderOrganization.new(name: 'Health Hut', npi: '1111111111', id: 2)
```
Jose-verdance, line 8: Is this for a test case, how come we are hard coding these values?

dpc-portal/app/controllers/csp_controller.rb
```diff
@@ -43,12 +43,27 @@ def logout
...
+
     sign_in_and_log(user, csp.name)
     sync_csp_emails(csp_user, all_emails(auth), primary_email(auth))
     ial_2_actions(user, auth)
     redirect_to path(user, auth)
   end
 
+  def email_match?(csp_user, auth)
+    csp_user.user_emails.empty? || csp_user.user_emails.map(&:email).include?(primary_email(auth))
```
Jose-verdance, line 57: What would cause an existing user to not have any emails?

ashley-weaver: this was a holdover from debugging some failing tests, I think

dpc-portal/app/components/page/existing_account/add_email_component_preview.rb
ashley-weaver: this is for the lookbook preview

lukey-luke:
I was able to successfully link multiple email accounts to sandbox Login.gov and trigger the add email page by signing in with the new Login.gov email.

When I click add email, it routes back to homepage, which is good, but it's falling back on the 
`"You need to sign in or sign up before continuing."` message, because DPC doesn't recognize the account as logged in.

- This is because the `sign_in_and_log()` method isn't called.  
- I think it might be easier to call render for email after signing in, that way DPC is "signed in" the same time Login.gov is "signed in"

e.g. 
```
  def user_actions(auth, csp)
    csp_user = CspUser.find_by(uuid: auth.uid, csp:)
    user = csp_user&.user
    user_path = path(user, auth)


    sign_in_and_log(user, csp.name)
    ial_2_actions(user, auth)
    
    return render_add_email(user.email, csp.name, user_path) if user && !email_match?(csp_user, auth)
    
    sync_csp_emails(csp_user, all_emails(auth), primary_email(auth)) # moved to not update `UserEmail` yet
    redirect_to user_path
  end
  ```
  and then the accept button for new email component would need a new route instead of going to the homepage, it needs to call new controller method that will `sync_csp_emails()` and then redirect them to the homepage ("/")

File: dpc-portal/spec/integration/public_keys_spec.rb [hunk 6]
```diff
@@ -16,7 +16,7 @@
     let!(:org) { create(:provider_organization, dpc_api_organization_id:, name: 'Health Hut') }
     let(:label) { 'New Public Key' }
 
-    LoginSupport::CSP_MAP.each do |display_name, provider|
+    CspUtils::CODES_TO_DISPLAY.each do |display_name, provider|
```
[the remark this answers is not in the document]
ashley-weaver: Valid.

dpc-portal/spec/support/csp_auth_shared_examples.rb
```diff
@@ -94,7 +79,42 @@
       end
     end
 
-    context :user_does_not_exist do
+    context 'user exists with different email' do
+      before do
+        user = create(:user)
+        csp_user = create(:csp_user, user:, uuid:, csp:)
+        create(:user_email, csp_user:, email: 'original@example.com', active: true)
+      end
```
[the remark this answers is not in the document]
ashley-weaver: Valid.

dpc-portal/app/controllers/csp_controller.rb
```diff
@@ -43,12 +43,32 @@ def logout
...
+  end
+
+  def render_add_email(csp_user, auth)
+    Rails.logger.info(['User has existing account associated with different email',
+                       { actionContext: LoggingConstants::ActionContext::Authentication,
+                         actionType: LoggingConstants::ActionType::MergeUserAccountEmail,
+                         **csp_log_context }])
+    render(Page::ExistingAccount::AddEmailComponent.new(csp_user.user.email, csp_user.csp.name,
```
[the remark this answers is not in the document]
ashley-weaver: user model has email method, not valid

dpc-portal/app/controllers/concerns/csp_email_sync.rb
```diff
@@ -2,6 +2,14 @@
 
 # Handles CSP emails
 module CspEmailSync
+  # Alternative to login for add email flow
+  def update
+    csp_user = CspUser.find_by(id: params[:id], csp: params[:csp])
+    sync_csp_emails(csp_user, params[:all_emails], params[:primary_email])
+
+    redirect_to root_url
+  end
```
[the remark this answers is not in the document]
ashley-weaver: Valid

dpc-portal/app/concerns/csp_utils.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  CODES_TO_DISPLAY = {
+    login_dot_gov: 'Login.gov',
+    id_me: 'ID.me',
+    clear: 'CLEAR'
+  }.freeze
+
+  def self.display_name(csp_code)
+    CODES_TO_DISPLAY.fetch(csp_code.to_sym, 'CSP')
+  end
```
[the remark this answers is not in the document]
ashley-weaver: Edited to remove default instead.

lukey-luke:
I'm seeing a forbidden error when I try to add a new email. It **does** sign my user in, but fails to `sync_csp_emails()` so manually navigating back to the homepage shows logged in and no new `UserEmail` objects.

https://github.com/user-attachments/assets/ebe859ca-0935-4184-9f30-30e3b02c6522

This guard is firing when it should have my CspUser info
```
    return render plain: :forbidden, status: :forbidden unless csp_user&.user == current_user
```

lukey-luke:
> I'm seeing a forbidden error when I try to add a new email. It **does** sign my user in, but fails to `sync_csp_emails()` so manually navigating back to the homepage shows logged in and no new `UserEmail` objects.
> 403_for_add_email.mov
> 
> This guard is firing when it should have my CspUser info
> 
> ```
>     return render plain: :forbidden, status: :forbidden unless csp_user&.user == current_user
> ```

added additional logging and it looks like `current_user` is `nil` when I reach this part of the code

dpc-portal/spec/support/csp_auth_shared_examples.rb
```diff
@@ -238,4 +260,50 @@
...
+      let(:params) do
+        { id: csp_user.id, csp: csp.id, all_emails: %w[alice1@example.com alice2@example.com],
+          primary_email: 'alice1@example.com' }
+      end
+
+      context 'not logged in' do
+        it 'returns forbidden' do
+          post '/update', params: params
+          expect(response.status).to eq 403
+        end
```
lukey-luke, line 277: Is this reproducible by real users?

After update to `current_user`, I don't know how this code would be reachable?

dpc-portal/spec/support/csp_auth_shared_examples.rb
```diff
@@ -94,7 +79,43 @@
...
-    context :user_does_not_exist do
+    context 'user exists with different email' do
+      before do
+        user = create(:user)
+        csp = Csp.find_by(name: csp_name) || create(:csp, name: csp_name)
+        csp_user = create(:csp_user, user:, uuid:, csp:)
+        create(:user_email, csp_user:, email: 'original@example.com', active: true)
+      end
+
+      it 'renders the add email component' do
```
lukey-luke, line 90: 👍

dpc-portal/spec/support/csp_auth_shared_examples.rb
ashley-weaver: this is probably overkill, but I figured better safe than sorry

lukey-luke (approved):
LGTM

dpc-portal/app/controllers/csp_controller.rb
manojwadhwa81: We have to match against all emails ( primary and verified/confirmed) on auth object to see if we get a match. Current logic only checks on primary email

MEspositoE14s (approved):
I can see why this was a 5pt ticket! 👍

PR #3093, opened 2026-08-07
Me/dpc 5620 force https flag on portal

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5620

## 🛠 Changes

- Enforces https usage on the portal.
- Fixes a bug that would've started the fake CPI gateway in prod.

## ℹ️ Context

Setting the ssl flag was required to meet security standards.  

In production, `$ENV` is set to "prod", not "production".  If we deployed the old version of the entrypoint.sh script to prod, the conditional that starts the portal would've fallen through and called the nonprod script, which would've also started the fake CPI gateway.  In prod we want to connect to the real thing.

## 🧪 Validation

Deployed to dev [here](https://github.com/CMSgov/dpc-app/actions/runs/31184971511).

### DIFF
--- a/dpc-portal/docker/entrypoint.sh
+++ b/dpc-portal/docker/entrypoint.sh
@@ -30,7 +30,7 @@ if [ "$1" == "portal" ]; then
   fi
 elif [ "$1" == "async" ]; then
   echo "Starting Solid Queue..."
-  if [[ "$ENV" == "production" ]]; then
+  if [[ "$ENV" == "prod" ]]; then
     echo "Starting in production"
     bundle exec rails solid_queue:start
   # For local, SolidQueue starts in same container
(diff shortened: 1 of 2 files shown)


dpc-portal/docker/entrypoint.sh
```diff
@@ -18,7 +18,7 @@ if [ "$1" == "portal" ]; then
   echo "Migrating the database..."
   bundle exec rails db:migrate db:seed
 
-  if [[ "$ENV" == "production" ]]; then
+  if [[ "$ENV" == "prod" ]]; then
```
lukey-luke, line 21: great catch! ⚾

lukey-luke (approved):

ashley-weaver (approved):

PR #3124, opened 2026-08-27
[DPC-5633] Random token as path suffix for each of the URL in the invitation flow

manojwadhwa81:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5633

## 🛠 Changes

Add random token as path suffix for each of the URL in the invitation flow for both CD and AO

## ℹ️ Context

To make invitation links difficult to guess thus preventing leak of metadata

## 🧪 Validation
A valid invitation id with a non-matching token 404s a generic message.

Updated invitation model and request spec to ensure:
a) Tokens are generated for any new invitation
b) Tokens are matched in before action as part of the invitation load
c) Resend link regenerates the token
d) Tokens are unique
e) Mailers include new token.


dpc-portal/config/routes.rb
```diff
@@ -28,14 +28,19 @@
...
-      get 'accept', on: :member
-      post 'confirm', on: :member
-      post 'register', on: :member
-      post 'login', on: :member
-      post 'renew', on: :member
-      get 'confirm_cd', on: :member
-      get 'set_idp_token', on: :member
+    resources :invitations, only: [] do
+      member do
+        scope ':token', constraints: { token: /[A-Za-z0-9]{24}/ } do
```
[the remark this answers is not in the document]
manojwadhwa81: I had to revert back the suggestion as anchor elements aren't allowed in routes config regex.

andresen-nava:
@manojwadhwa81 Have you considered using the token instead of the ID in the url [by overriding to_param](https://guides.rubyonrails.org/routing.html#renaming-default-route-parameter-id)? This should allow us to avoid changes to the routes, reduce nesting in the url, and allow us to pass fewer parameters to the path helpers. We would probably want to add a [uniqueness validation and constraint](https://guides.rubyonrails.org/active_record_validations.html#uniqueness) for the token if we switched to that approach.

manojwadhwa81:
@andresen-nava to_param is new to me. On exploring the doc for it, it looks like it kinda replaces the id in the url for the show method with the output of this method and that ActiveRecord's default implementation is to return a models' id for it. By using the approach I have, we keep the semantics of a unique identifier for an invitation, while adding a random string that need to be available to the user to get to use the invite. So the developers can still talk and refer to invitation using invitation id. Most of the change are in the specs to update them to use the new URL syntax, and the invitation_url_for method in InvitationUrlSupport module kind of plays the same role as to_param to help generate the url for specs. Lets synch if you think I'm missing out anything about to_param.

We have a unique constraint on token enforced via an index. 
 uniqueness: true  in model class
and 
  add_index :invitations, :token, unique: true in migration file
So we are covered there too..

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -357,13 +365,22 @@ def update_user(user_info)
...
   def load_invitation
-    @invitation = Invitation.find(params[:id])
-    if @organization != @invitation.provider_organization
-      invalid_status = @invitation.credential_delegate? ? 'cd_invalid' : 'ao_invalid'
-      render(Page::Utility::ErrorComponent.new(@invitation, invalid_status), status: :not_found)
-    end
-  rescue ActiveRecord::RecordNotFound
-    render(Page::Utility::ErrorComponent.new(@invitation, 'ao_invalid'), status: :not_found)
+    @invitation = Invitation.find_by(id: params[:id], token: params[:token])
+    return render_invitation_not_found if @invitation.nil?
```
Jose-verdance, line 369: Why not also include the organization id in this lookup?

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -357,13 +365,22 @@ def update_user(user_info)
...
+    invalid_status = @invitation.credential_delegate? ? 'cd_invalid' : 'ao_invalid'
+    render(Page::Utility::ErrorComponent.new(@invitation, invalid_status), status: :not_found)
+  end
+
+  # No invitation matched the id/token pair, so we cannot say anything about the invitation itself.
+  def render_invitation_not_found
+    log_event(:info, 'Invitation not found',
+              action_context: LoggingConstants::ActionContext::Registration,
+              action_type: LoggingConstants::ActionType::InvalidInvitation,
+              invitation: params[:id])
```
Jose-verdance, line 382: Should we also log more info like the invitation token, organization id, and invitation type??

dpc-portal/app/controllers/invitations_controller.rb
```diff
@@ -398,14 +415,14 @@ def invitation_log_data(reason)
...
+    redirect_to organization_invitation_url(*invitation_url_args) unless @invitation.authorized_official?
   end
 
   def verify_cd_invitation
-    redirect_to organization_invitation_url(@organization, @invitation) unless @invitation.credential_delegate?
+    redirect_to organization_invitation_url(*invitation_url_args) unless @invitation.credential_delegate?
   end
 
-  def check_for_token
+  def check_for_csp_token
```
Jose-verdance, line 425: I see why you renamed this, it would have gotten confusing. 👍

dpc-portal/db/migrate/20260824150000_add_token_to_invitations.rb
```diff
@@ -0,0 +1,30 @@
...
+
+# Add a random token to Invitation URLs, it will be required by every route in the flow.
+class AddTokenToInvitations < ActiveRecord::Migration[8.0]
+  # Local model so the backfill is not affected by later changes to Invitation.
+  class MigrationInvitation < ActiveRecord::Base
+    self.table_name = 'invitations'
+  end
+
+  def up
+    add_column :invitations, :token, :string
```
Jose-verdance: We should limit the size of the column values to 24 characters as well so its not unbounded.

Jose-verdance:
Hey Manoj, how come I am seeing this error when entering a longer or shorter token value?

Screenshot 2026-08-28 at 3 17 13 PM

dpc-portal/app/controllers/invitations_controller.rb
manojwadhwa81: yep, now token term is overloaded.

dpc-portal/app/controllers/invitations_controller.rb
manojwadhwa81: Invitation id should serve the debug purpose as it uniquely identifies an invitation. Not logging invitation token also limits the attack surface as now just having access to logs won't enable anyone to guess full url.

dpc-portal/app/controllers/invitations_controller.rb
manojwadhwa81: The sample value space for token ( 2^24) is far bigger than the limited number of orgs ( few thousands may be ) we'll be sending invite too. So adding org id won't help here.

manojwadhwa81:
Hi @Jose-verdance This will happen only in dev environment, and is the default Rails provide. This helps with debugging. There is another ticket that Becky created to have custom page for catch all error instead of rails provided default pages but it's marked as low priority now.
You should be able to get to this error for any invalid route you try outside of invitation flow too.

manojwadhwa81:
Made a change to the migration after the initial commit. In order to apply the changes, run the following commands in sequence:
rails db:rollback
rails db:migrate

ashley-weaver (approved):
👍

Jose-verdance (approved):

PR #3126, opened 2026-09-01
Me/dpc 5653 update portal healthcheck alarms

MEspositoE14s:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5653

## 🛠 Changes

- Update `verify_resource_health_job` to differentiate between the health checks for each CSP.
- Switched CSP health checks to use the OIDC discover endpoint of each CSP.
- Renamed IDP -> CSP.
- Updated `csp` model and `csp_config` with `discovery_uri`.
- Fixed bug where CSP health checks were executing every minute instead of every hour.
- Added option in makefile to shell into portal container without starting the portal.

## ℹ️ Context

Our CSP health checks were written when we only had one CSP and have never really worked since we moved to three.  The login.gov health check has been passing, while id.me and clear have consistently failed.  Those failures were hidden because the Cloudwatch alarms they trigger were also written when we only had one CSP, and weren't alerting.

There is an associated dpc-ops PR: https://github.com/CMSgov/dpc-ops/pull/975

## 🧪 Validation

Successful deploy [here](https://github.com/CMSgov/dpc-app/actions/runs/33429408131).

New dimensions in Cloudwatch:
image

New metrics in Cloudwatch:
image

### DIFF
--- a/dpc-portal/config/recurring.yml
+++ b/dpc-portal/config/recurring.yml
@@ -23,13 +23,19 @@ production:
     schedule: "*/1 * * * *"
     class: "VerifyResourceHealthJob"
     queue: portal
-    args: [ { check_dpc: true, check_cpi: false, check_idp: false } ]
+    args:
+      check_dpc: true
+      check_cpi: false
+      check_idp: false
 
   verify_non_dpc_health_job:
     schedule: "0 * * * *"
     class: "VerifyResourceHealthJob"
     queue: portal
-    args: [ { check_dpc: false, check_cpi: true, check_idp: true } ]
+    args:
+      check_dpc: false
+      check_cpi: true
+      check_idp: true
 
   log_organizations_api_credential_status_job:
     schedule: "0 0 * * *"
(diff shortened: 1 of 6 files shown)


dpc-portal/config/recurring.yml
```diff
@@ -23,13 +23,19 @@ production:
     schedule: "*/1 * * * *"
```
lukey-luke, line 23: The PR description says that verify health should be updated from every 1 minute to every hour, should we update the `schedule` cron here?

```
  verify_dpc_health_job:
    schedule: "0 * * * *"
```

MEspositoE14s: The way it was originally written, the `verify_dpc_health_job` that checks dpc-api should run every minute, and the `verify_non_dpc_health_job` that checks the CSPs should run every hour.  

In the original version there was a problem where both of them were being called every minute, though, because the way we were passing parameters to the `VerifyResourceHealthJob` caused them all evaluate as true.  When the dpc-api job ran at the start of every minute it was also running the CSP checks that should only run every hour.  This PR should get them working the way they're supposed to.

lukey-luke (approved):
verified metrics are emitted for `dev` and `test ✅ 
verified `check_csp=false` for checks that fire every minute (cron expression same) ✅

Jose-verdance (approved):

PR #3139, opened 2026-09-10
[DPC-5694]: Verify user info token from CSP using published jwks

manojwadhwa81:
Verify the token as returned in the user info endpoint by utilizing the public key information as available through the jwks url published via .well-known/openid-configuration endpoint
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5684

## 🛠 Changes

Updated the patch we created earlier to not just decode the token but verify the signature in the token through the kid and public key. Added additional constraint to verify the host as used for configuring strategy is one of the three host configured in csp.yml.
Accounted for the fact that the convention for using /well-known/openid-configuration might not be followed, and in that case the discovery_uri should be configured in csp.yml and that would trump the default well known open id configuration derived by adding /well-known/openid-configuration  to host.

## ℹ️ Context

Using JWT for user info respnose payload is uncommon, but since one of the CSP utilizes this mechanism, we'd like to verify the token before deciding to use it.

## 🧪 Validation

Added tests to ensure the following:
a) We hit the /well-known/openid-configuration  to retrieve JWKS uri.
b) We hit the JWKs uri to download the public key
c) Tempered JWTs are rejected

### DIFF
--- a/dpc-portal/spec/config/initializers/omniauth_openid_connect_patch_spec.rb
+++ b/dpc-portal/spec/config/initializers/omniauth_openid_connect_patch_spec.rb
@@ -3,9 +3,25 @@
 require 'rails_helper'
 
 describe OmniAuth::Strategies::OpenIDConnect do
-  let(:strategy) { described_class.new(nil) }
+  let(:idp_host) { 'idp.example.com' }
+  let(:strategy) do
+    described_class.new(nil, client_options: { host: idp_host, userinfo_endpoint: '/userinfo' })
+  end
   let(:payload) do
     { 'sub' => '12345', 'email' => 'test@example.com' }
+  end
+  let(:rsa_key) { OpenSSL::PKey::RSA.generate(2048) }
+  let(:jwk) { JSON::JWK.new(rsa_key) }
+
+  before do
+    stub_const('OmniAuth::Strategies::OpenIDConnect::ALLOWED_IDP_HOSTS_DISCOVERY_URL_MAP',
+               { idp_host => "https://#{idp_host}/.well-known/openid-configuration" })
+  end
+
+  def signed_jwt(claims, key: rsa_key, kid: jwk[:kid])
+    token = JSON::JWT.new(claims)
+    token.kid = kid
+    token.sign(key, :RS256).to_s
   end
 
   describe '#user_info' do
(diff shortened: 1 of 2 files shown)


dpc-portal/spec/config/initializers/omniauth_openid_connect_patch_spec.rb
```diff
@@ -57,4 +73,132 @@
...
+      # expect(OpenIDConnect.http_client).to have_received(:get).with(well_known_uri).once
+      # expect(OpenIDConnect.http_client).to have_received(:get).with(jwks_uri).once
+    end
+  end
+
+  describe '#well_known_configuration_uri' do
+    it 'uses the discovery endpoint as configured in csp.yml file' do
+      expect(strategy.send(:well_known_configuration_uri))
+        .to eq("https://#{idp_host}/.well-known/openid-configuration")
+    end
```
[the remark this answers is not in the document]
manojwadhwa81: After refactoring, removed the fallback upon the standard '/.well-known/openid-configuration' path in the absence of explicit setting discovery_uri in csp.yml and relying upon the discovery url as configured in csp.yml

dpc-portal/app/services/oidc_jwks_verifier.rb
```diff
@@ -0,0 +1,86 @@
...
+      nil
+    end
+  ).to_set.freeze
+
+  class << self
+    # Decodes +jwt_string+, verifying its signature against +host+'s
+    # published JWKS. If the JWT's kid isn't found in our cached key set,
+    # the JWKS is refreshed once before giving up, so that a key rotated in
+    # by the provider but missing in the stale JWKS doesn't cause failures
+    def decode_and_verify(jwt_string, host:)
```
ashley-weaver: this method verifies the signature, but should it also validate the other claims (iss, aud, and/or sub) match the expected values? 

> The sub Claim in the UserInfo Response MUST be verified to exactly match the sub Claim in the ID Token; if they do not match, the UserInfo Response values MUST NOT be used.
https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse

dpc-portal/app/services/oidc_jwks_verifier.rb
```diff
@@ -0,0 +1,86 @@
...
+      JSON::JWT.decode(jwt_string, jwk_set_for(host)).to_h
```
ashley-weaver: the documentation for ruby-jwt says the algorithm should be set explicitly to avoid an attack vulnerability

> For safe cryptographic signing, you need to specify the algorithm in the options hash whenever you call JWT.decode to ensure that an attacker [cannot bypass the algorithm verification step](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/). It is strongly recommended that you hard code the algorithm, as you may leave yourself vulnerable by dynamically picking the algorithm

https://github.com/jwt/ruby-jwt

dpc-portal/app/services/oidc_jwks_verifier.rb
ashley-weaver: we should probably also validate the JWT isn't expired with the exp claim

manojwadhwa81: Added validations for issuer, audience and exp claims..  This class is handling user info and won't have access to the id token, nor are we storing id token in session, so can 't match sub claim.

dpc-portal/app/services/oidc_jwks_verifier.rb
manojwadhwa81: Hardcoded RS256 after inspecting the token, typically this function uses the one available in header to verify signature, but that leaves us vulnerable as you mentioned, but with the new parameter passed, we should be good.

lukey-luke:
I could not find one definitive answer on best practices for key rotation. This accepted SO answer [HERE](https://stackoverflow.com/a/61661802) lists 1 hour - 1 week as best practice, so limiting cache entries to one hour seems like a reasonable timeframe.

I confirmed on my end for the CLEAR csp that the portal does properly retrieve keys from `https://verified.clearme.com/integrations/.well-known/jwks.json`. I'm assuming that the second key is for when keys are rotated out, they will do it one at a time, but there is a safeguard for `KidNotFound`

PR #3141, opened 2026-09-14
[DPC-5720] Fix errors in tofu getting swallowed and not failing CI

lukey-luke:
## 🎫 Ticket

[DPC-5720](https://jira.cms.gov/browse/DPC-5720)

## 🛠 Changes

 - explicitly set bash shell for all workflows that use `config-stdout-to-cw-logs` so that failures are captured in CI output

## ℹ️ Context

  - CI workflows were recently updated to log infra related CI outputs to cloudwatch instead of GitHub as a security measure ([DPC-5374](https://jira.cms.gov/browse/DPC-5374))
  - It was recently discovered that this mechanism would swallow errors that are piped to cloudwatch (e.g. tofu step passes [HERE](https://github.com/CMSgov/dpc-app/actions/runs/34399642709))
  - Digging into Github docs, looks like we can ensure `pipefail` is set and ensure workflows fail as soon as possible. Setting this should fix the issue.
    - Shell parameter in GHA docs [HERE](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsshell) specifies that `pipefail` will be set for `bash`
    - See also [Exit codes and error action preference](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#exit-codes-and-error-action-preference)

## 🧪 Validation

- example of error getting swallowed when CI should not pass: https://github.com/CMSgov/dpc-app/actions/runs/34903900270/job/104175997384
- example of the same error failing the CI: https://github.com/CMSgov/dpc-app/actions/runs/34904191541
- example of failed tofu command failing the CI: https://github.com/CMSgov/dpc-app/actions/runs/34905189393


Jose-verdance (approved):
LGTM!

MEspositoE14s (approved):
This is better than what I was thinking of for this 👍

PR #3144, opened 2026-09-16
[DPC-5748] Update error handling in CSP auth

ashley-weaver:
## 🎫 Ticket

https://jira.cms.gov/browse/DPC-5748

## 🛠 Changes

Failure in the sign-in path now redirects the user back to the same sign-in page they entered at.

## ℹ️ Context

This unblocks [DPC-5714](https://jira.cms.gov/browse/DPC-5714) and [DPC-5715](https://jira.cms.gov/browse/DPC-5715), allowing these tickets to be worked on in parallel.

## 🧪 Validation

Tests updated.


dpc-portal/app/controllers/concerns/csp_error_handling.rb
```diff
@@ -20,48 +20,56 @@ def csp_param
...
+
+  private
+
+  def handle_csp_error(level:, alert_text:, action_type:, invitation:, error: nil)
+    log_event(level, alert_text,
+              action_context: action_context(invitation),
+              action_type: action_type,
+              csp: csp_param,
+              **{ error: }.compact)
+  end
```
lukey-luke: Is the intention that this function will be extended in other tix?
- if the intention is to DRY up the module, then I don't think this does that. There's a single wrapper `log_event()` and now `handle_csp_error()` does the same thing

ashley-weaver: I was originally planning to have render_or_redirect in this method too, but this may be going a little overboard on the modularization

dpc-portal/app/controllers/concerns/csp_error_handling.rb
```diff
@@ -20,48 +20,56 @@ def csp_param
...
+
+  def action_context(invitation)
+    invitation.nil? ? LoggingConstants::ActionContext::Authentication : LoggingConstants::ActionContext::Registration
+  end
+
+  def redirect_url(invitation)
```
andresen-nava: Would you be open to renaming this to `redirect_link`? The suffix `url` typically indicates that it is a [Rails url helper](https://guides.rubyonrails.org/routing.html#path-and-url-helpers). This is not a technical necessity but may improve readability.

dpc-portal/spec/support/csp_auth_shared_examples.rb
```diff
@@ -374,9 +374,9 @@
         OmniAuth.config.mock_auth[provider] = error
       end
 
-      it 'returns 503 service unavailable' do
+      it 'returns 200 OK' do
         attempt_sign_in
-        expect(response).to have_http_status(:service_unavailable)
+        expect(response).to be_ok
```
lukey-luke: If I'm understanding correctly, the controller would result in a 5xx error on the user's end in this scenario previously, but now we redirect to sign in and will show an appropriate error message (to be covered in [DPC-5714](https://jira.cms.gov/browse/DPC-5714) and [DPC-5715](https://jira.cms.gov/browse/DPC-5715).
- could you please add a check to ensure we landed on the right page in addition to status code?
e.g.
```
        expect(response.location).to eq same_sign_in_page_url()
```

dpc-portal/app/controllers/concerns/csp_error_handling.rb
```diff
@@ -21,47 +21,46 @@ def csp_param
...
               csp: csp_param)
-    handle_error_redirects(invitation, 'csp_signin_cancel')
+    render_or_redirect(invitation, 'csp_signin_cancel')
   end
 
-  # rubocop:disable-next Metrics/AbcSize
-  def handle_error_redirects(invitation, error_display_text)
-    alert_text = "We weren't able to complete identity verification."
+  def render_or_redirect(invitation, error_reason)
+    return render(Page::Utility::ErrorComponent.new(nil, error_reason, csp: csp_param)) if invitation.nil?
```
andresen-nava: For this PR to unblock [DPC-5714](https://jira.cms.gov/browse/DPC-5714) and [DPC-5715](https://jira.cms.gov/browse/DPC-5715), I'm wondering if we may want to redirect to the `sign_in_url` instead of rendering the error component. What do you think?

lukey-luke (approved):

andresen-nava (approved):

PR #3153, opened 2026-09-23
[DPC-5756] set last used csp cookie at end of ao/cd flows

lukey-luke:
## 🎫 Ticket

[DPC-5756](https://jira.cms.gov/browse/DPC-5756)

## 🛠 Changes

- sets last used CSP cookie in the invitation flow

## ℹ️ Context

  - Previously, invitation flow required users to sign in again.
  - Separately, a feature was added to mark the "last used" csp.
  - When invitations were updated to redirect to organizations page and stay signed in, this "last used" feature was not considered.
  - This means that for users that just signed up, when they sign out, they don't get the "last used" indicator.

## 🧪 Validation

 - added additional rspec test


lukey-luke:
> invitations_spec

This is valid. We checked the template output _when_ the cookie was set, but tests prior did not verify that the cookie was set during sign-in. Updated.

ashley-weaver (approved):
good find

andresen-nava (approved):

